Issue #73, opened 2021-05-11
Generalize metadata to work across media types


dhruvkb:
The goal is to extract common metadata out of the image model into a generic one that can then be used across upcoming media types, specifically audio and video.

Branched from WordPress/openverse#19.

PR #79, opened 2021-05-28
Extract generalised media models from image models

dhruvkb:
Partially addresses #73.

This PR extracts common fields from image models into abstract generalised models. Image models now inherit and extend these base models.

The goal was to prevent database migrations (for ease-of-review and mergeability) so help text has not been updated in any of the base model fields. That is a separate PR (#80, that can be merged into this branch or, if this PR is merged first, into `master`).

The `models.py` file (which had grown too big, and will continue to do so) has been changed into a module with a number of concern-based files. This should help manage the size-complexity.


zackkrida (approved):
So nice to split these files up. Excellent

krysal (commented):
It is definitely an improvement to have the models in separate files, as well as the additional comments. Thanks!

krysal (approved):

obulat (approved):
The code looks much cleaner now, thank you, Dhruv!

PR #80, opened 2021-05-28
Update help text in generic media models

dhruvkb:
This PR updates `help_text` in model fields of all abstract base classes. Changes include replacing 'image' with 'media', 'CC' with 'Openverse' among other smaller copy updates.

Can be merged into #79 first, or into `master` later.


zackkrida (approved):
Great!

Issue #75, opened 2021-05-11
Create models for Audio and AudioReport


dhruvkb:
The Audio model will act as the DB representation of audio files and the AudioReport will be used to record reports against audio files. Any OOP generalisations possible between the parallel model classes is encouraged in #73.

Branched from WordPress/openverse#19.

obulat:
There has been some work done on the Audio Model previously, with a list of possible fields: WordPress/openverse#1752

PR #81, opened 2021-06-01
Add Django models for audio

dhruvkb:
Addresses #75.

This PR adds models for audio, building upon the framework established by image models. Functionality that is generic has been migrated up into the base classes.

**References:** 
- https://github.com/WordPress/openverse-catalog/issues/5


dhruvkb:
@krysal needs your inputs on the fields captured in the `Audio` and `AudioSet` models.

openverse-api/catalog/api/models/audio.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    type = models.CharField(
+        max_length=80,
+        blank=True,
+        null=True,
+        db_index=True,
+        help_text='',
+    )
+
+    duration = models.IntegerField(
```
obulat, line 62: The sound effects in freesound have float-type duration, because some of them are less than one second in length. I'm not sure what's the best way to represent that, though. Milliseconds, maybe?

dhruvkb: Less than one second? That's too short. Storing them in milliseconds will just add three 0s to all other sounds which are more than 1s long. But I can't think of any alternative approaches so will have to go with that.

obulat: Tens of millisecond is also an option :)
[Some sounds](https://freesound.org/browse/tags/oh/) seem to be 0 seconds in duration on the search page, but turn out to be 0.xxx seconds when you look at the detail page.

obulat:
Do you think it would make sense to have a `media_url` (or just `url`) column that is common for images and audios, and would hold the link to the canonical file in case of audio, and the largest size available for image? Or is it better to keep them separate, and have `url` field both on audio and image?

openverse-api/catalog/api/models/audio.py
dhruvkb: Changed the default unit to milliseconds in 9da5519.

openverse-api/catalog/api/models/media.py
```diff
@@ -169,6 +146,70 @@ class AbstractMediaReport(models.Model):
...
+        update records and refresh indices.
+
+        Media marked as mature or deleted also leads to instantiation of their
+        corresponding mature or deleted classes.
+
+        Additional kwargs:
+        index_name    : Name of ES index, eg. 'image'
+        media_class   : Class of the media, eg. ``Image``
+        mature_class  : Class that stores mature media, eg. ``MatureImage``
+        deleted_class : Class that stores deleted media, eg. ``DeletedImage``
```
zackkrida, line 167: 😻 📝

zackkrida (approved):
Great work, this actually taught me a few things about the repo. ✅

openverse-api/catalog/api/models/audio.py
```diff
@@ -0,0 +1,180 @@
...
+        return str(self)
+
+class AudioSet(IdentifierMixin, MediaMixin, FileMixin, OpenLedgerModel):
+    """
+    This is an ordered collection of audio files, such as a podcast series or
+    an album. Not to be confused with AudioList which is a many-to-many
+    collection of audio files, like a playlist or favourites library.
+
+    The FileMixin inherited by this model refers not to audio but album art.
+    """
```
krysal, line 51: Do we need the `filesize` of the album art for something? I don't see if it’s used for thumbnails in the case of images for example. Asking because then we could name the url of the album art as `cover_url` or `artwork_url` in that case.

krysal (approved):
Excellent work with documentation, much appreciated! I only have a couple of question but they are no blockers I think.
Migrations go in a separate PR?

openverse-api/catalog/api/models/audio.py
```diff
@@ -0,0 +1,180 @@
...
+    alt_files = JSONField(
+        blank=True,
+        null=True,
+        help_text='JSON describing alternative files for this audio.',
+    )
+
+    @property
+    def alternative_files(self):
+        if hasattr(self.alt_files, '__iter__'):
+            return list(map(lambda val: AltAudioFile(val), self.alt_files))
```
obulat, line 112: Could we rewrite it using list comprehension:
`[AltAudioFile(alt_file) for alt_file in self.alt_file]` ?
It is [more Pythonic and easier to read](https://docs.quantifiedcode.com/python-anti-patterns/readability/using_map_or_filter_where_list_comprehension_is_possible.html) than map/lambda expressions.

openverse-api/catalog/api/models/media.py
```diff
@@ -169,6 +146,70 @@ class AbstractMediaReport(models.Model):
     class Meta:
         abstract = True
 
+    def url(self, media_type):
+        url = ('https://search.creativecommons.org/'
```
obulat, line 150: Adding a base_url constant will probably make changing this URL easier when we need to.

openverse-api/catalog/api/models/audio.py
zackkrida: TIL about list comprehensions 👍

openverse-api/catalog/api/models/audio.py
dhruvkb: The filesize field is not currently being used anywhere. I was just trying to be consistent with the fields used in images, so that we can reuse code as much as possible, especially when using the same image proxy for artwork as image media.

openverse-api/catalog/api/models/media.py
dhruvkb: Addressed in 7e69204.

openverse-api/catalog/api/models/audio.py
dhruvkb: Thanks, I love comprehensions. Resolved in 47eecf5.

dhruvkb:
> Migrations go in a separate PR?

I was waiting for approval on the models fields before proceeding with the migrations, as frequent changes to the file would add unnecessary noise in the commits. Added migrations in c19234f.

openverse-api/catalog/api/models/media.py
```diff
@@ -131,6 +108,8 @@ class AbstractMediaReport(models.Model):
     deleted content.
     """
 
+    BASE_URL = 'https://search.creativecommons.org/'
```
obulat, line 111: Should we update the BASE_URL now?

zackkrida (approved):

obulat (approved):

krysal (approved):

openverse-api/catalog/api/models/media.py
zackkrida: LGTM and do it in a separate PR to rename the base url.

PR #101, opened 2021-06-10
Update release-drafter.yml

zackkrida:
Fix release drafter to run on the primary branch.

Note: the `$default-branch` variable doesn't work in workflows, but it works in workflow _templates_.

https://stackoverflow.com/a/65723433/3433762


dhruvkb (dismissed):
Would need to be made in all repos, I suppose.

dhruvkb (approved):

obulat (approved):

PR #102, opened 2021-06-11
Dependencies update

krysal:
This PR compress all dependencies updates signaled by dependabot in order to address potential security vulnerabilities. Packages updated are:
- django 
- django-redis
- pytest-django
- boto3
- hvac
- among others...

I also added the `cache` service as dependency of the `web` service in the `docker-compose.yml` file, because it's required by the application. This way we can run `docker-compose up web` without errors, omitting the rest of services (indexers, analytics, ...) it's useful for quick checks.


obulat (approved):
Builds fine, tests pass :)

docker-compose.yml
```diff
@@ -63,6 +63,7 @@ services:
     depends_on:
       - db
       - es
+      - cache
```
dhruvkb, line 66: Nice catch!

dhruvkb (approved):
LGTM!

PR #105, opened 2021-06-16
Add DRF serializers for audio

dhruvkb:
Addresses #74. Builds upon PR #81.

This PR adds serializers for audio, building upon the framework established by image serializers. Functionality that is generic has been migrated up into the base classes.

Issue #74, opened 2021-05-11
Create API endpoints for audio


dhruvkb:
The goal is to create endpoints for audio, similar to the image endpoints at `v1/images/`. 

Branched from WordPress/openverse-frontend#61.

PR #108, opened 2021-06-18
Add DRF endpoints for audio

dhruvkb:
Addresses #74. Builds upon PR #105.

This PR adds endpoints for audio, building upon the framework established by image endpoints. Functionality that is generic has been migrated up into the base classes.


openverse-api/catalog/api/views/media_views.py
```diff
@@ -0,0 +1,190 @@
...
+            return input_error_response(value_error)
+
+        context = {'request': request}
+        serialized_results = media_serializer(
+            results,
+            many=True,
+            context=context
+        ).data
+
+        if len(results) < page_size and num_pages == 0:
```
obulat, line 117: I'm not sure it belongs to this PR or a separate one:
The `num_pages` returns a number that is 1 less than the actual number of pages. For example, if the `page_size` is 20, and we only have 40 results, `num_pages` will be 1, although in reality we get 2 pages. If there are fewer than 20 results, there will be 0 pages?

obulat (approved):
One step closer to adding Audio :) Thank you for all the work!

openverse-api/catalog/api/views/media_views.py
dhruvkb: The search controller calculates page size as follows:
https://github.com/WordPress/openverse-api/blob/eefd9a46d09f97ba8c7077d5734772a5611cf50e/openverse-api/catalog/api/controllers/search_controller.py#L487

This calculation always leads to page count less by 1. We can fix this but there will break backwards compatibility with API clients that add 1 to the page count from their end.

zackkrida: We're not going to be transferring API users from CC to Openverse, just notifying them that there's a new API available, so I think _now_ is a reasonable time to make this fix. We can note the behavior in our email to current API users.

zackkrida (approved):

openverse-api/catalog/api/views/media_views.py
dhruvkb: Cool, I'll make this a separate issue as many PRs are chaining up behind this one.

PR #115, opened 2021-06-28
Replace deprecated `falcon.API` with `falcon.App`

### LINKED ISSUE #114: Replace deprecated falcon.api with falcon.app
> The API class was renamed to App in Falcon 3.0. The old class name remains available as an alias for backwards-compatibility, but will be removed in a future release.

[Falcon docs](https://falcon.readthedocs.io/en/latest/api/app.html)

We use `falcon` in `analytics` and in the `ingestion_server`. We should replace deprecated `app` class name with `api` class name.

obulat:
Fixes #114 

Falcon version 3.0 deprecated `falcon.API` class, and replaced it with `falcon.App`. This PR uses the new way of creating the falcon App.

### DIFF
--- a/analytics/server.py
+++ b/analytics/server.py
@@ -83,7 +83,7 @@ cors = CORS(
     allow_all_methods=True,
     allow_all_headers=True
 )
-api = falcon.API(middleware=[cors.middleware])
+api = falcon.App(middleware=[cors.middleware])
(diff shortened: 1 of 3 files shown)


zackkrida (approved):

dhruvkb (approved):
👍 All good!

PR #116, opened 2021-06-28
Remove hardcoded `image` tablename in ingestion_server

obulat:
The ingestion server needs to be adjusted for audio addition. This small PR removes hard-coded `image` value in table names and Elasticsearch index names.


zackkrida (dismissed):
Looks good. It might be a little annoying to have to update this every time we add a new media type, but that won't happen very often so it should be okay.

obulat:
> Looks good. It might be a little annoying to have to update this every time we add a new media type, but that won't happen very often so it should be okay.

Thank you for your comment! I refactored it to make updating less annoying :)
To add a new media type, you only need to add it to the `constants.py` file. If a media type is not found or is not in the list, the scripts default to 'image' for backward compatibility.

I'm not sure `constants.py` is the best name or the root folder is the best place to put the constants, would be happy to edit with better options!

zackkrida (dismissed):
`constants.py` feels like a fine name to me 😄

dhruvkb:
Although `constants.py` is a fine name to boot, it can become a mega-file if we're not careful. If you have bandwidth to refactor it, I'd recommend making it a module with purpose specific files inside it, for example `constants/media_types.py`.

Refs:
- [Grumpy chronicles: The question of utils](https://tonsky.me/blog/utils/)
- [https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/](https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/)

dhruvkb (approved):
Apart from the above suggestion, looks all good to me!

obulat:
> I'm saying this because there are a lot of constants in `catalog-api` too that I want to extract to a `constants` module (but since it's a huge refactor, it's preferable to have tests before undertaking it).

Before reading this paragraph, I was going to suggest to add the `constants` folder once we have, say, more than 3 types of constants. But since we have more constants to refactor anyways, we might as well start now :) I will add the changes today.

dhruvkb:
I wasn't recommending all projects to share the same constants folder (although they can, in theory, with some ill-advised Docker mount-fu), I was just suggesting to make a folder based on my experience in the catalog.

obulat:
> I wasn't recommending all projects to share the same constants folder (although they can, in theory, with some ill-advised Docker mount-fu), I was just suggesting to make a folder based on my experience in the catalog.

I guess it was me trying to over-engineer things. I'm strongly against Docker mount-fu!
So, the final changes have a `constants` folder inside the `ingestion_server` app, which currently has one file, [`media_types.py`](https://github.com/WordPress/openverse-api/blob/media_ingestion/ingestion_server/ingestion_server/constants/media_types.py).
I think this is optimal for now, we can always change later if it does not fit our needs anymore.

dhruvkb (approved):

PR #117, opened 2021-06-28
Replace `string.format()` with `fstrings`

obulat:
Python 3.6 introduced a new way of formatting strings using so-called f-strings. This made formatted strings more readable and concize.

Instead of writing 
```python
who = "world"
print("Hello {who}".format(who=who))
```

we can write 

```python
who = "world"
print(f"Hello {who}")
```

This PR consistently replaces `string.format` with `f-strings`.

### DIFF
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -57,7 +57,7 @@ class TaskResource:
         request_error = self._validate_create_task(raw_body)
         if request_error:
             logging.warning(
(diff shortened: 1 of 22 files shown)


zackkrida (approved):
I like that you also switched simple cases like `VAR_NAME + 'test-123'` to `f'{VAR_NAME}test-123'`; the consistency is appreciated.

dhruvkb (approved):
This is so much better, thank you!

PR #132, opened 2021-07-07
ElasticSearch models audio

### LINKED ISSUE #129: [Feature] Add ElasticSearch Audio model
## Problem

As part of the work on adding Audio to Openverse, we need to make the audio data searchable by indexing the audio data to the ElasticSearch.  The models for image are in [`elasticsearch_models.py`](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/elasticsearch_models.py)

## Description

We should look into extracting details that are common for all media, just as we did in the API models.

## Implementation

obulat:
Fixes #129

ElasticSearch models are declared in `ingestion_server/ingestion_server/elasticsearch_models.py`. They are used to create the mappings from database data to ElasticSearch in `ingestion_server/ingestion_server/es_mapping.py`.

This PR
- extracts common fields from the `Image` model to a parent `Media` class
- adds an `Audio` model as a subclass of `Media`
- replaces template string SQL queries with safe dynamically generated one
- cleans up and refactors code in lots of places
- adds a bunch of documentation


ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -281,6 +281,9 @@ def database_row_to_elasticsearch_doc(row, schema):
             extension=Audio.get_extension(row[schema['url']]),
             categories=row[schema['category']],
             license_url=Audio.get_license_url(meta),
+            bit_rate=row[schema['bit_rate']],
+            sample_rate=row[schema['sample_rate']],
+            genre=row[schema['genre']],  # Not sure how to map array
```
dhruvkb, line 286: Based on [this discussion](https://github.com/elastic/elasticsearch-dsl-py/issues/95#issuecomment-487974464), I think just setting an array as the value works.

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -247,50 +283,50 @@ class Audio(Media):
...
     """
 
+    class Durations(Enum):
+        """
+        Maximum threshold for each audio duration band
+
+        These durations are also hardcoded in the `duration` field in
+        openverse-api/catalog/api/serializers/audio_serializers.py.
+        """
+        SHORT = 10 * 60 * 1e3  # 10 minutes
```
obulat, line 293: Youtube has the following classification:
- under 4 minutes
- 4-20 minutes
- over 20 minutes.
We might also consider making "Short" mean under 3 minutes, so that it would mainly match sound effects like many of the freesound.org sounds, and "Medium" would mean mostly songs or short talks/podcasts etc.

dhruvkb: Thanks! I was just shooting in the dark with the time ranges.

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -212,20 +204,152 @@ def get_popularity(raw):
...
+    class Index:
+        name = 'audio'
+
+    @staticmethod
+    def database_row_to_elasticsearch_doc(row, schema):
+        meta = row[schema['meta_data']]
+        provider = row[schema['provider']]
+        authority_boost = Audio.get_authority_boost(meta, provider)
+
+        attrs = Image.get_instance_attrs(row, schema)
```
krysal, line 327: `Image` or **`Audio`**? it could be `Media` also...

krysal (approved):
This looks excellent @dhruvkb! So organized and well documented 🌟 Thanks!

ingestion_server/ingestion_server/elasticsearch_models.py
zackkrida: Yes, pretty sure this should me `Media.get_instance_attrs` right?

zackkrida (approved):
excellent. so excited for this.

ingestion_server/ingestion_server/elasticsearch_models.py
dhruvkb: Referencing the method from any of the three class names (`Audio`, `Image` or `Media`) works just the same:
- It's a static method so that class name we use to invoke it does not matter.
- It's defined in `Media` but inherited by both `Audio` and `Image` so calling it on Media should not be necessary.

It does make the most sense to call it from the same class as it's used in so that we can override the parent definition if we need some specific fields. Updated that in 9ef51ad. The commit dismissed your reviews unfortunately 😢.

zackkrida (approved):

krysal (approved):

PR #133, opened 2021-07-07
Update field names for set to avoid shadowing the built-in name

dhruvkb:
This PR updates the field names for the `AudioSet` foreign key in `Audio` to avoid shadowing the built-in name `set` from Python collections.


obulat (approved):
LGTM!

krysal (approved):

PR #136, opened 2021-07-08
Remove unnecessary call to clean audio data ingestion server

obulat:
Currently, we have two cleaning steps that are doing exactly the same thing: cleaning tags and metadata, in both Catalog and API's ingestion server. The cleaning step was first added to the API ingestion server, and then to the catalog.

The reason it was not removed from the API repo is the fact that we still might have old image data that hasn't been cleaned up in the catalog.

Since audio data will be newly-ingested and cleaned in the catalog, we do not need the API cleaning step for audio. This PR checks for table name, and skips the call to clean the data if it is 'audio'. This also fixes one error on the path to audio ingestion in the API :)

### DIFF
--- a/ingestion_server/ingestion_server/elasticsearch_models.py
+++ b/ingestion_server/ingestion_server/elasticsearch_models.py
@@ -52,11 +52,113 @@ class SyncableDocType(DocType):
         )
 
 
+class Media(SyncableDocType):
+    """
+    Represents a media object in Elasticsearch. Note that actual mappings are defined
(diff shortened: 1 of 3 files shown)


zackkrida (approved):

zackkrida (approved):
Could you perhaps add a comment in the `cleanup.py` file explaining that it's not in use and why? I don't necessarily want to delete the file so it'd be nice to document why it's here and not in use.

dhruvkb (approved):
This is good!

PR #137, opened 2021-07-09
Update Docker image for ingestion-server to use bind mounts properly

dhruvkb:
This PR updates the ingestion server `Dockerfile` and the project `docker-compose.yml` file to bind-mount the ingestion-server codebase from the repository instead of copying the code during image build. This allows code changes to reflect immediately in the container instead of needing a rebuild everytime.

As a side-effect, two binary files named `db` and `lock` get created in the repo which should not be Git tracked so they have been `.gitignore`-d.

Also this PR makes the `Dockerfile` symmetrical with the API `Dockerfile`, which is good.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -99,7 +99,7 @@ services:
       - es
       - indexer-worker
     volumes:
-      - ./ingestion_server:/ingestion-server
+      - ./ingestion_server:/ingestion_server
     environment:
       PYTHONUNBUFFERED: "0"
       ELASTICSEARCH_URL: 'es'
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -99,7 +99,7 @@ services:
       - es
       - indexer-worker
     volumes:
-      - ./ingestion_server:/ingestion-server
+      - ./ingestion_server:/ingestion_server
```
obulat, line 102: I think the same change needs to be done on line 132 (`ingestion-server` -> `ingestion_server`)

dhruvkb: This PR entirely left out any changes to the indexer workers. I was considering to do that in a separate PR.

obulat (dismissed):
Thank you for this change! It will make development much faster and more convenient!

docker-compose.yml
dhruvkb: Addressed in #138.

zackkrida (approved):
😍

PR #141, opened 2021-07-14
Update sample data for audio

dhruvkb:
**Ref:** [Spreadsheet of data points](https://docs.google.com/spreadsheets/d/1pTPudKh6S19u8Y3dG1QosKBDg_WQbGLnFLmhOVynqtM/edit)

This PR
- renames the `genre` column to `genres`
- changes the `genres` field from JSON field to array field
- renames the `standardized_audio_popularity` column to `standardized_popularity`
- updates the data CSV to match

These changes are so that

- the `genres` column can directly be ingested into the downstream database where the column is governed by the `genres` field in the `Audio` model.
https://github.com/WordPress/openverse-api/blob/f874c3622130136dac2d96815470b0fbd0ccaa9f/openverse-api/catalog/api/models/audio.py#L71
- the data matches sample data for images


dhruvkb:
Changed base branch to `es_models_audio` to use the ES mappings for audio ingestion.

dhruvkb:
🎉 Verified that both audio and image indices are created successfully!

Screenshot 2021-07-20 at 11 22 57 AM

dhruvkb:
🎉 Verified QA indices for audio and images.
Screenshot 2021-07-20 at 11 58 08 AM

load_sample_data.sh
```diff
@@ -17,31 +17,40 @@ docker-compose exec -T "$WEB_SERVICE_NAME" /bin/bash -c "python3 manage.py shell
...
-	ALTER TABLE image_view ADD COLUMN standardized_popularity double precision;
 	CREATE TABLE content_provider(provider_identifier varchar(50), provider_name varchar(250), created_on timestamp, domain_name varchar(500), filter_content boolean, notes text);
 	INSERT INTO content_provider (created_on, provider_identifier, provider_name, domain_name, filter_content) VALUES (now(), 'flickr', 'Flickr', 'https://www.flickr.com', false), (now(), 'behance', 'Behance', 'https://www.behance.net', false);
+	EOF"
+
+# Load sample data for images
+docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" /bin/bash -c "PGPASSWORD=deploy pg_dump -s -t image -U deploy -d openledger -h db | psql -U deploy -d openledger"
+docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" /bin/bash -c "psql -U deploy -d openledger <<-EOF
+	ALTER TABLE image RENAME TO image_view;
+	ALTER TABLE image_view ADD COLUMN standardized_popularity double precision;
```
krysal, line 35: This file looks clearer ✨ The `sample_data.tsv` of images now also has an `ingestion_type` column so it should change too but that is probably for another PR, if possible.

sample_data/sample_audio_data.csv
```diff
@@ -1,101 +1,101 @@
...
+identifier,created_on,updated_on,ingestion_type,provider,source,foreign_identifier,foreign_landing_url,url,thumbnail,duration,bit_rate,sample_rate,category,genres,audio_set,alt_files,filesize,license,license_version,creator,creator_url,title,meta_data,tags,last_synced_with_source,removed_from_source,view_count,standardized_popularity
```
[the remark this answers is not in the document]
dhruvkb: Removed in 42ee22a.

load_sample_data.sh
dhruvkb: The [sample data for images](https://github.com/WordPress/openverse-api/blob/c27aeac9ab6ba51e65a44eb5a3d54fbb980f91a1/sample_data/sample_data.csv) has not been updated yet so we should make a new issue and PR for that.

load_sample_data.sh
```diff
@@ -17,31 +17,40 @@ docker-compose exec -T "$WEB_SERVICE_NAME" /bin/bash -c "python3 manage.py shell
...
-#	\copy audio_view (identifier,created_on,updated_on,ingestion_type,provider,source,foreign_identifier,foreign_landing_url,url,thumbnail,duration,bit_rate,sample_rate,category,genre,audio_set,alt_audio_files,filesize,license,license_version,creator,creator_url,title,meta_data,tags,last_synced_with_source,removed_from_source,standardized_popularity) from './sample_data/sample_audio_data.csv' with (FORMAT csv, HEADER true)
-#	EOF"
+# Load sample data for audio
+docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" /bin/bash -c "PGPASSWORD=deploy pg_dump -s -t audio -U deploy -d openledger -h db | head -n -14 | psql -U deploy -d openledger"
```
obulat, line 40: What does this line do? Could we remove the extra step of altering tables (image -> image_view, audio -> audio_view) and create the necessary tables from the start?

dhruvkb: This creates the `audio_view` table in the upstream DB as a replica of the `audio` table from the API DB. The pipe to `head` (`| head -n -14 `) is for removing the `audio`&rarr;`audio_set` FK.

The benefit is that it prevents us from having to declare all the fields manually and just use the autogenerated DB schema from Django's models.

obulat: Thank you for clear explanations!

ingestion_server/ingestion_server/qa.py
```diff
@@ -10,65 +10,122 @@ class QAScores(Enum):
...
-        provider='test',
-        source=sample_url,
-        license='by',
-        license_version='3.0',
-        foreign_landing_url=sample_url,
-        metadata=None,
-        view_count=0
-    )
+    return {
+        '_id': _id,
```
obulat, line 26: Are both of these `id` fields necessary?

dhruvkb: This [doc page](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-id-field.html) made it seem like a useful field. Plus I'm generally hesitant to touch things that work 😄.

load_sample_data.sh
```diff
@@ -17,31 +17,40 @@ docker-compose exec -T "$WEB_SERVICE_NAME" /bin/bash -c "python3 manage.py shell
...
 docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" /bin/bash -c "psql -U deploy -d openledger <<-EOF
 	ALTER TABLE image RENAME TO image_view;
 	ALTER TABLE image_view ADD COLUMN standardized_popularity double precision;
-	CREATE TABLE content_provider(provider_identifier varchar(50), provider_name varchar(250), created_on timestamp, domain_name varchar(500), filter_content boolean, notes text, media_type varchar(80));
-	INSERT INTO content_provider (created_on, provider_identifier, provider_name, domain_name, filter_content) VALUES (now(), 'flickr', 'Flickr', 'https://www.flickr.com', false, 'image'), (now(), 'behance', 'Behance', 'https://www.behance.net', false, 'image');
 	\copy image_view (id,created_on,updated_on,identifier,provider,source,foreign_identifier,foreign_landing_url,url,thumbnail,width,height,filesize,license,license_version,creator,creator_url,title,tags_list,last_synced_with_source,removed_from_source,meta_data,tags,watermarked,view_count,standardized_popularity) from './sample_data/sample_data.csv' with (FORMAT csv, HEADER true)
```
obulat, line 36:
```suggestion
	\copy image_view (identifier,created_on,updated_on,ingestion_type,provider,source,foreign_identifier,foreign_landing_url,url,thumbnail,width,height,filesize,license,license_version,creator,creator_url,title,meta_data,tags,watermarked,last_synced_with_source,removed_from_source,standardized_popularity) from './sample_data/sample_data.csv' with (FORMAT csv, HEADER true)
```

openverse-api/catalog/api/models/media.py
```diff
@@ -54,7 +54,11 @@ class AbstractMedia(IdentifierMixin, MediaMixin, FileMixin, OpenLedgerModel):
     )
     removed_from_source = models.BooleanField(default=False)
 
-    view_count = models.IntegerField(default=0)
+    view_count = models.IntegerField(
```
obulat, line 57: Should we remove this, too?

obulat (approved):
This is a great improvement that will make API more in line with the Catalog.

I think we can merge after removing `tags_list` and `view_count` fields.

ingestion_server/ingestion_server/qa.py
obulat: Sorry, I added a comment to the wrong line. On line 27, there is another `id` field. I suspect that it corresponds to the now deprecated `id` field from the image_view database. If it does, then it should probably be removed because we no longer have `id` in sample data (we have `identifier` instead)

load_sample_data.sh
dhruvkb: Will just removing the field names from the `\copy` command work or do we have to delete the columns from the CSV as well?

openverse-api/catalog/api/models/media.py
dhruvkb: I'm not sure. I made it nullable so that it gets out of the way to begin with.

load_sample_data.sh
dhruvkb: Updated in 75ca546.

dhruvkb:
With 4c8ff4b, the ingestion for audio and images both works but the tests break because they used responses from the old sample data.

Eg: the old sample data had results for "Claude Monet" and "Jim Monet". They're gone now and so following test breaks:
https://github.com/WordPress/openverse-api/blob/c27aeac9ab6ba51e65a44eb5a3d54fbb980f91a1/openverse-api/test/v1_integration_test.py#L202

dhruvkb:
Now the API integration tests are fixed but the analytics tests are failing 😞.

dhruvkb:
🎉

obulat (approved):

openverse-api/test/v1_integration_test.py
```diff
@@ -221,7 +221,7 @@ def test_creator_quotation_grouping():
     # Did we find only Claude Monet works, or did his lesser known brother Jim
     # Monet sneak into the results?
```
krysal, line 222: The comment is outdated.
```suggestion
    # Did we find only William Ford Stanley works, or also by others?
```

krysal (approved):
⭐️

Issue #143, opened 2021-07-15
[Bug] Integration tests of Ingestion server are failing


krysal:
## Description
The integration tests of the `ingestion_server` are throwing errors and therefore, failing, details of the output later.

## Reproduction

Steps to run tests are in the README but I write here for convenience as well:
1. Change directory to ingestion server: `cd ingestion_server`
2. Install dependencies: `pipenv install`
3. Run the integration tests: `pipenv run python3 test/integration_tests.py`
4. See errors.

## Expectation

Tests should execute correctly and pass.

## Environment

 - Device: MBP M1
 - OS: macOS Big Sur (11.4)
 - Docker version: 20.10.7, build f0df350 with docker-compose version 1.29.2, build 5becea4c

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

Issue #104, opened 2021-06-14
Update Postgres in ingestion server integration test


zackkrida:
Should be `13.2`

https://github.com/WordPress/openverse-api/blob/9d0d724651f18cc9f96931e01bea92b8032bd6a0/ingestion_server/test/integration-test-docker-compose.yml#L15

krysal:
Note: This file is updated when running Ingestion server tests (instructions in README file, may require installing additional dependencies).

zackkrida:
Ah yes, I learned this recently too, that this file is generated! I'm going to close this issue, thanks.

PR #144, opened 2021-07-15
Modify docker-compose file for Ingestion-server's integration tests

krysal:
Related to #143 and #104.

When executing integration tests (steps in [README][readme_ref] file) of the ingestion-server, the **integration-ingestion** service cannot start correctly due to source files not being mounted in the docker container. The error in the logs appeared as:
```
INFO:root:Waiting for ingestion server to come up...
integration-ingestion_1  | Error: could not find config file config/supervisord.conf
integration-ingestion_1  | For help, use /usr/bin/supervisord -h
test_integration-ingestion_1 exited with code 2
INFO:root:Waiting for ingestion server to come up...
INFO:root:Waiting for ingestion server to come up...
INFO:root:Waiting for ingestion server to come up...
INFO:root:Waiting for ingestion server to come up...
INFO:root:Waiting for ingestion server to come up...
INFO:root:Waiting for ingestion server to come up...
ERROR:root:Ingestion server timed out. Giving up.
```

This PR fixes the generation of the docker-compose file for ingestion-server's tests. Also moves the increment of the `attempts` variable to the `finally` block of the exception, in order to prevent an infinite loop, previously it wasn't incrementing if the `except` block was executed.

This does not fix the integration tests completely but enables them to run, a first step.

[readme_ref]: https://github.com/WordPress/openverse-api/blob/dc_integration_tests/README.md#how-to-run-ingestion-server-tests

### DIFF
--- a/ingestion_server/test/generate_integration_test_docker_compose.py
+++ b/ingestion_server/test/generate_integration_test_docker_compose.py
@@ -31,6 +31,7 @@ with open(parent_docker_compose, 'r') as docker_compose_file:
         es = docker_compose['services']['es']
         ingestion_server = docker_compose['services']['ingestion-server']
         upstream_db = docker_compose['services']['upstream_db']
+
         # Delete services we're not testing.
         desired_services = {'es', 'db', 'ingestion-server', 'upstream_db'}
         for service in dict(docker_compose['services']):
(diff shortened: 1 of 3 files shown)


ingestion_server/test/generate_integration_test_docker_compose.py
```diff
@@ -55,6 +56,9 @@
         ingestion_server['depends_on'] = ['integration-es', 'integration-db']
         ingestion_server['build'] = '../'
 
+        # Set the directory correctly in the volume
+        ingestion_server['volumes'] = ['../:/ingestion_server']
+
```
dhruvkb, line 61: Seems like this change was needed because of #137.

dhruvkb (approved):
The integration tests are very necessary, thanks for this!

ingestion_server/test/generate_integration_test_docker_compose.py
krysal: Yes, but it's much better how it is now :)

PR #149, opened 2021-07-21
Add media type to content provider

dhruvkb:
Since we are now adding new media types, the `ContentProvider` model needs to specify which media type it provides. This PR adds the new `media_type` field to the `ContentProvider` model.


dhruvkb:
Checking test failure @ `test_stats`

obulat (commented):
What about providers with several media types, like Wikimedia or Europeana? Is an array a good option for media_type field?

dhruvkb:
It'll make filtering complicated. Can we not keep two different providers with the same display name but different IDs (like Wikimedia internally stored as wikimedia_images and wikimedia_audio)?

openverse-api/catalog/api/constants/media_types.py
```diff
@@ -0,0 +1,11 @@
+"""
```
obulat, line 1: Nice to see constants in the API, too!

openverse-api/catalog/api/models/models.py
```diff
@@ -14,6 +15,10 @@ class ContentProvider(models.Model):
         verbose_name='Hide content'
     )
     notes = models.TextField(null=True)
+    media_type = models.CharField(
```
obulat, line 18: I think we should add a node somewhere where we create the content providers about the different internal representation for providers that have several content types (wikimedia_image, wikimedia_audio)

obulat (approved):
LGTM!

openverse-api/catalog/api/models/models.py
zackkrida: Good suggestion @obulat

zackkrida (approved):

zackkrida:
> two different providers with the same display name but different IDs

Yeah i think this is fine for now.

openverse-api/catalog/api/models/models.py
dhruvkb: Added usage comment in 99c18e0.

dhruvkb: Added usage comment in 99c18e0.

PR #154, opened 2021-08-02
Add the `sound_effect` category

dhruvkb:
This PR adds the 'sound_effect' category in the model, migration and serializer. The migration `0032_audio_models.py` has been modified instead of creating a new one because the change hasn't gone into production.


obulat (approved):
LGTM!

zackkrida (approved):

PR #201, opened 2021-09-07
Add just and pre-commit

sarayourfriend:
## Fixes

Fixes #193 by @sarayourfriend 
Fixes #187 by @sarayourfriend 

## Description
Adds `pre-commit` using the same configuration as the `openverse-catalog` with some modifications to ignore rules in certain files that were problematic.

Also adds a `justfile` to make docker stuff easier.

## Technical details
I had to add some `noqa` comments for long strings (I wish flake8 would ignore strings...)

I also added `.flake8` that mirrors the pre-commit configuration for flake8 so that editor integration will work

## Tests
Checkout the branch and run `just install` then `pipenv run pre-commit run --all-files` and ensure that there are no errors reported.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main` or `master`).
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #202, opened 2021-09-08
Refactor the use of environment variables

### LINKED ISSUE #192: [Feature] Use environment files instead of hardcoding in Compose
## Problem

The compose file has lots of environment variables hardcoded inside the file itself.

## Description

Extracting them into .env files would make it easier to 
- change variables in the future as needed
- use the same config locally, outside of Docker

## Implementation

dhruvkb:
## Fixes

Fixes #192 by @dhruvkb

## Description

This PR
- cleans up environment variables, removing unused ones
- adds maximum defaults for environment variables to enable running outside Docker
- defines environment variables for running in Docker
- provides documentation for running the services locally and running tests locally

## Technical details

There are two types of `.env` files defined in the PR
- `.env`: automatically detected and used by Pipenv when running locally
- `.env.docker`: defined in `docker-compose.yml` to be used by Docker-Compose

## Checklist

- [x] My pull request targets the *default* branch of the repository (`main` or `master`).

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
Reminder to self: resolve upcoming conflicts with #201.

zackkrida (approved):
I like this a lot! I left some small notes. The documentation improvements are very welcome.

I do wonder if we're violating expectations by tracking and including `.env` files in the repo, or if keeping those untracked and user-created is only a convention in front-end / JavaScript communities.

dhruvkb:
You're right about the convention of not including `.env` files but since they are very consistent for all developers using Docker and Pipenv, I felt that including them would make development￼ easier.￼

README.md
```diff
@@ -2,89 +2,101 @@
...
+
+2. Clone the repository and `cd` into it. This is the monorepo root.
+    ```bash
+    git clone https://github.com/WordPress/openverse-api.git
+    cd openverse-api/
+    ```
+
+3. From the monorepo root, bring up the Docker Compose system. Docker Compose will automatically read the necessary environment variables from `.env.docker` files from project directories.
+    ```bash
+    docker-compose up
```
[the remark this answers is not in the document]
dhruvkb: Yes, I left a reminder to self above (https://github.com/WordPress/openverse-api/pull/202#issuecomment-915836835).

README.md
```diff
@@ -2,89 +2,101 @@
...
-- Elasticsearch
-- Redis
-- A thumbnail-generating image proxy
-- ingestion-server, a service for bulk ingesting and indexing search data.
-- analytics, a REST API server for collecting search usage data
+- [Git](https://git-scm.com/downloads)
+- [Docker](https://docs.docker.com/install/)
+- [Docker Compose](https://docs.docker.com/compose/install/)
+- [Pipenv](https://pipenv.pypa.io/en/latest/#install-pipenv-today)
+
```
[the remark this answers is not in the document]
dhruvkb: Yes, I left a reminder to self above (https://github.com/WordPress/openverse-api/pull/202#issuecomment-915836835).

README.md
```diff
@@ -2,89 +2,101 @@
...
+    ```bash
+    ./load_sample_data.sh
+    ```
+
+7. Make an API request using cURL. You should receive a JSON response.
+    ```bash
+    curl localhost:8000/v1/images/?q=honey
+    ```
+
+    Piping the response through `json_pp` should yield an output like this.
```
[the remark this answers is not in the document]
dhruvkb: Not sure which is the more popular tool but iirc `json_pp` comes bundled with the OS. Or I might have installed it long ago and forgotten 🤔.

krysal (changes requested):
The `.env` file must be excluded of tracking. Typically, it's a hidden file intended to store the secrets of _each user_, so when one changes it will track the change and will be troublesome for the rest. These new files should be named `.env.template` or `.env.example` and indicate they must be copied to a `.env` properly when doing the project setup.

zackkrida:
@dhruvkb I think I support @krysal's thinking, mostly to prevent people accidentally attempting to publish local config. If we had a command to do all the `.env.example` copying to `.env` that would be pretty neat.

dhruvkb:
There are no secrets in the env files for the API. But that's okay, I understand the reasoning behind it. I'll wait for @sarayourfriend's PR that adds `just` and then use it for the environment files.

README.md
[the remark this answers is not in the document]
dhruvkb: Updated reference to `jq`.

README.md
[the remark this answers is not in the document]
dhruvkb: Added dependency on Just.

README.md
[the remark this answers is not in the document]
dhruvkb: Updated commands to Just.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -39,26 +40,28 @@
...
+ELASTICSEARCH_URL = os.getenv("ELASTICSEARCH_URL", "localhost")
+ELASTICSEARCH_PORT = int(os.getenv("ELASTICSEARCH_PORT", 9200))
+
+AWS_REGION = os.getenv("AWS_REGION", "us-east-1")
+
+DATABASE_HOST = os.getenv("DATABASE_HOST", "localhost")
+DATABASE_PORT = int(os.getenv("DATABASE_PORT", 5432))
+DATABASE_USER = os.getenv("DATABASE_USER", "deploy")
+DATABASE_PASSWORD = os.getenv("DATABASE_PASSWORD", "deploy")
+DATABASE_NAME = os.getenv("DATABASE_NAME", "openledger")
```
sarayourfriend, line 55: TIL there's no difference between `os.getenv` and `os.environ.get` 😮 https://stackoverflow.com/questions/16924471/difference-between-os-getenv-and-os-environ-get

openverse_api/README.md
```diff
@@ -0,0 +1,71 @@
...
+   ```
+
+3. Setup environment if you haven't already.
+   ```bash
+   pipenv install
+   ```
+
+4. Start the Django dev server. Pipenv will automatically read the necessary environment variables from `.env`.
+   ```bash
+   pipenv run python manage.py runserver
```
sarayourfriend, line 26: Could we add a `just` recipe for `manage.py`?

openverse_api/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+2. Change into the API directory.
+   ```bash
+   cd openverse_api/
+   ```
+
+3. Run the tests in a separate terminal instance. Pipenv will automatically read the necessary environment variables from `.env`.
+   ```bash
+   pipenv run ./test/run_test.sh
+   ```
```
sarayourfriend, line 55: Couldn't this just be `just testlocal`?

openverse_api/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. Ensure that Docker containers are up. See the section above for instructions.
+   ```bash
+   docker-compose ps
+   ```
+
+2. Run the tests in an interactive TTY connected to a `web` container.
+   ```bash
+   docker-compose exec web bash ./test/run_test.sh
+   ```
```
sarayourfriend, line 41: This could be replaced by `just test` I think, right?

openverse_api/catalog/settings.py
```diff
@@ -30,21 +30,28 @@
 # See https://docs.djangoproject.com/en/2.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = os.environ.get("DJANGO_SECRET_KEY")
+SECRET_KEY = os.environ["DJANGO_SECRET_KEY"]  # required
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.environ.get("DJANGO_DEBUG_ENABLED", default=False) in true_strings
+DEBUG = os.getenv("DJANGO_DEBUG_ENABLED", "True").lower() in TRUE_STRINGS
```
sarayourfriend, line 36: Unrelated to this PR, but should this security warning be codified somehow? It seems like we could write a condition that will crash the server or at least show a massive warning if this setting is on in production.

openverse_api/catalog/settings.py
```diff
@@ -53,7 +60,7 @@
 }
 SHORT_URL_PATH_WHITELIST = ["/v1/list", "/v1/images/"]
 
-USE_S3 = os.getenv("USE_S3", False)
+USE_S3 = os.getenv("USE_S3", "False") in TRUE_STRINGS
```
sarayourfriend, line 63:
```suggestion
USE_S3 = os.getenv("USE_S3", "False").lower() in TRUE_STRINGS
```

Should there be a small utility function to interpret boolean env values? In some sense the `https://django-environ.readthedocs.io/en/latest/` might be helpful here anyway as it already does this, but something like `bool_env("USE_S3" False)` might make sense to have as this code is repeated several times.

krysal:
I have found that the best way to read environment variables is using [decouple](https://pypi.org/project/python-decouple/), it takes care of the casting and default values and is really easy to use. For example:
```python
from decouple import config

DEBUG = config("DEBUG", default=False, cast=bool)
ELASTICSEARCH_PORT = config("ELASTICSEARCH_PORT", default=9200, cast=int)
```
It gets rid of all those tricky workarounds, so it also makes the code looks cleaner, seems like a good time to start using it.

openverse_api/catalog/settings.py
zackkrida: Reminds me of this nightmare code on the frontend: 
https://github.com/WordPress/openverse-frontend/blob/main/src/utils/string-to-boolean.js

openverse_api/catalog/settings.py
krysal: With `decouple` all empty env vars without a default value make the application crash because it means it is required :) and also good catch here, I think the default value for `DEBUG` should remain `False` as if it were production.

ingestion_server/ingestion_server/indexer.py
AetherUnbound: Woah, TIL as well!

zackkrida: TILx3

zackkrida:
`decouple` looks great for casting environment variables and setting defaults.

openverse_api/catalog/settings.py
dhruvkb: Used `decouple` in e58d890.

openverse_api/catalog/settings.py
dhruvkb: Changed `DEBUG` to be false by default in 62fb393.

openverse_api/README.md
dhruvkb: Yes, you're right @sarayourfriend. I forgot about the `just` recipes, my bad. Resolved in 739c79a.

openverse_api/README.md
dhruvkb: Sorry, I forgot about the `just` recipes. Resolved in 739c79a.

openverse_api/README.md
dhruvkb: Added a recipe `dj` in e240a48.

ingestion_server/Pipfile.lock
```diff
@@ -73,11 +73,11 @@
...
         "elasticsearch": {
             "hashes": [
-                "sha256:1a9f146b7126a7e0621085f1825b6b2d091693a714d3b16c208749762e79c2bc",
-                "sha256:f928898fe06869516f2603f9a96a6f166c06888233806b31ac6568bac0266501"
+                "sha256:6c47eb68f643f3d7d6a6d971e571448b900b9470dc481421c9af9d28a1303911",
+                "sha256:bf1d210d82c78fbd3ae5cbedd756716d732433cc97184966688dcd1c75df4619"
             ],
             "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3' and python_version < '4'",
-            "version": "==7.14.1"
+            "version": "==7.15.0"
```
krysal, line 80: This is updating elasticsearch and other packages too. Is it the intention?

justfile
```diff
@@ -1,46 +1,77 @@
...
+# Show logs of all, or named, Docker services
+logs services="":
+    docker-compose {{ DOCKER_FILE }} logs -f {{ services }}
+
+
+########
+# Init #
+########
 
+# Create .env files from stencils
```
krysal, line 31:
```suggestion
# Create .env files from templates
```

ingestion_server/README.md
```diff
@@ -1,64 +1,65 @@
...
-
-This runs a simulated environment in Docker containers and ensures that ingestion is working properly.
+## Running on the host
 
-```
-mkvirtualenv venv
-source venv/bin/activate
-python test/integration_tests.py
-```
+1. Create environment variables from the stencil file.
```
krysal, line 24:
```suggestion
1. Create environment variables from the template file.
```

README.md
```diff
@@ -2,101 +2,94 @@
...
+  - upstream data source simulator
+  - API application database
+- [Elasticsearch](https://www.elastic.co/elasticsearch/)
+- [Redis](https://redis.io/)
+- [imageproxy](https://github.com/willnorris/imageproxy)
+- **web** (`openverse_api/`)
+- **ingestion-server** (`ingestion_server/`)
+- **analytics** (`analytics/`)
+
+The last three are subproject of this monorepo, described below.
```
krysal, line 88:
```suggestion
The last three are subprojects of this monorepo, described below.
```

krysal (commented):
A lot of improvements here, I just wonder why keep the `env.docker` files if they are almost identical to `env.template` files?

README.md
dhruvkb: Applied suggestion in 7c89cb0.

ingestion_server/README.md
dhruvkb: Applied suggestion in 7c89cb0.

justfile
dhruvkb: Applied suggestion in 7c89cb0.

ingestion_server/Pipfile.lock
dhruvkb: That should not have been committed. Reverted the change in fbe3af6.

dhruvkb:
@krysal the Docker environment files are separate from the template files because inside the Docker Compose network the containers can refer to each other based on the service name whereas on the host, all the containers can be reached at different ports on the `localhost` hostname.

krysal (approved):
I see, thank you for clarifying! And good idea moving the images to an assets folder ✅

zackkrida (approved):

PR #346, opened 2021-11-09
Production deployment files

AetherUnbound:
## Description

This PR adds a few ingestion-server specific deployment files. These are based off the top level `docker-compose` and `justfile`, but are specifically intended for production deployments.

These have currently been used on production to launch & kick off the data refresh.

## Testing Instructions

Run `just pull` locally within the `ingestion_server` directory, then `just up`. Note that the image is pulled from Dockerhub and is launched successfully.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb (commented):
Maybe I'm not understanding but the commands in the new `justfile` seem to exactly match the outer one. Is there some reason for redefining them here? As per my understanding `just` recursively looks in the parent directory till it finds one `justfile`.

AetherUnbound:
The `justfile` at the top level directory corresponds to the dev environment, which includes all of the available services (API, analytics, elasticsearch, etc.). For production we only need the ingestion server running. I can try to test this, but I suspect that even if `just` is run from this directory, the `docker-compose` file it will pick up is the top level one and so it may attempt to spin up all of the services. If it ends up picking up the `docker-compose` file in the `ingestion_server` subfolder then we definitely don't need to duplicate the `justfile`!

AetherUnbound:
I tested this locally and no matter where you initiate `just` within the repo, the files it refers to are always relative to the `justfile`. I've added an `IS_PROD` variable similar to the [catalog](https://github.com/WordPress/openverse-catalog/blob/14c2d6d7d164db309a98f954a8a036612e2bdf95/justfile#L7) which changes the docker-compose file that `just` references. I ran the steps locally and we can reuse all the ingestion server bits without having to duplicate the `justfile`!

zackkrida (approved):
I tested this yesterday and was pretty excited, the additions today are even better! Awesome. Thanks for the `justfile` path research too.

ingestion_server/docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: "2.4"
+services:
+    ingestion_server:
+        image: openverse/ingestion_server:1.4.0
+        ports:
+            - "8001:8001"
+        env_file:
+            - .env
+        dns:
+            - 172.31.0.2
```
dhruvkb, line 10: What does this do and why this specific IP? Unless I'm extra-ignorant (which is a very real possibility) a comment explaining this key and value might help.

dhruvkb (approved):
LGTM! I do have one question and one recommendation though.

ingestion_server/docker-compose.yml
AetherUnbound: Great call! I'm uncertain of the IP address specifically, but it has to do with assigning a specific DNS server. It was present in the run command prior to my interacting with the ingestion server, and I don't want to remove it for fear of breaking things :sweat_smile:

zackkrida: Oooh, I know this! 172.16.0.0–172.31.255.255 is a reserved IP range by the (IANA) for private / internal networking. This specific IP is the default for virtual private networks in AWS. This specific IP is the dns resolution address for a default AWS Virtual Private Cloud.

zackkrida: And also the default for DNS resolution in Docker! Which is probably the use case here. I can't find good documentation on this anywhere otherwise I'd share.

PR #356, opened 2021-11-16
Protect foreign key constraint between `imagelist_images` and `imagelist`

dhruvkb:
## Description

Since migration 0002 changes the primary key on the `ImageList` model, it breaks the foreign key constraint that refers to that column. This PR merges 0002 into 0001 so that the constraint is set **after** the changes to the primary key.

Credits to @AetherUnbound for pointing this bug out in prod.

## Testing Instructions

1. Migrate the database
    ```sh
    docker-compose exec -T web python manage.py migrate
    ```
1. Describe the `imagelist_images` table.
    ```sh
    docker-compose exec -T db psql -U deploy -d openledger <<-EOF
    \d+ imagelist_images;
    EOF
    ```
1. Ensure 2 foreign key constraints:
    - `FOREIGN KEY (image_id) REFERENCES image(id)`
    - `FOREIGN KEY (imagelist_id) REFERENCES imagelist(id)`

## ⚠️ Production notice

Since 0002 is already a performed migration, removing the corresponding file might interfere with rollbacks etc. To prevent potential future problems, remove the entry from the `django_migrations` table.

```sql
DELETE FROM 
  django_migrations
WHERE
  app = 'api' 
  AND name = '0002_auto_20180723_1737';
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/test/unit_test.py
```diff
@@ -140,9 +140,9 @@ def test_url_protocol_fix():
         result = CleanupFunctions.cleanup_url(bad_url, tls_support_cache)
         expected = "'https://flickr.com'"
 
-        bad_http = "neverssl.com"
+        bad_http = "baidu.com"
         result_http = CleanupFunctions.cleanup_url(bad_http, tls_support_cache)
-        expected_http = "'http://neverssl.com'"
+        expected_http = "'http://baidu.com'"
```
sarayourfriend, line 145: Was this change intentional?

zackkrida, line 145: I can speculate as to why, but what's the reason for the domain switch here?

sarayourfriend (commented):
Is there any way to fix this that doesn't involve changing the migration history? If not this LGTM but I think leaving the migration history as immutable is preferrable to modifying it.

openverse_api/catalog/api/migrations/0003_image_view_count.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('api', '0002_auto_20180723_1737'),
+        ('api', '0001_initial'),  # 0002 was merged into 0001 to protect FK constraint
```
zackkrida, line 9: 👍

zackkrida (approved):
Thank you @dhruvkb and @AetherUnbound, this is great.

ingestion_server/test/unit_test.py
dhruvkb: NeverSSL now supports SSL ☹️

dhruvkb: NeverSSL now supports SSL ☹️ I read that Baidu doesn't. The tests are passing locally, not sure why they're failing on CI.

zackkrida: Another one I know is http://mitm.cool (`man in the middle` lol)

sarayourfriend: Do these make an actual HTTP request during the tests?

If so, depending on the behavior of a 3rd party server for the validity of our test suite is not good. If we need to test a non-SSL http request, then we should mock the request using something like VCR or nock (whatever the Python equivalent of those is).

dhruvkb: Updated the test to use mocks in 2181e32.

dhruvkb: That's cool but mocking is the stablest solution.

ingestion_server/test/unit_test.py
```diff
@@ -141,6 +142,7 @@ def test_url_protocol_fix():
         expected = "'https://flickr.com'"
 
         bad_http = "neverssl.com"
+        TlsTest.test_tls_supported = MagicMock(return_value=False)
```
krysal, line 145: Nice! It is a more robust test now.

krysal (commented):
Django has a command for [squashing migrations](https://docs.djangoproject.com/en/3.2/topics/migrations/#squashing-migrations), didn't it work here? Altering the old ones manually is not a good practice. Otherwise, it seems to work well.

dhruvkb:
@krysal the reason I did not squash the migrations and rather edit them by hand was that this approach removes the need to actually run any migrations in production. 

Editing migrations is bad but, in our case, the prod environment and dev are quite out of sync so the usual guidelines need to be bent a little to bring all environments to the same level again.

krysal (approved):

AetherUnbound:
This migration ran successfully on a snapshot of prod :heavy_check_mark:

PR #394, opened 2021-12-11
Fix page_size setting for related media

### LINKED ISSUE #393: page_size for /related/ endpoint is set incorrectly and causes 400 error
Currently on the main branch, `/related/` endpoint returns a 400 error. This is caused by incorrectly setting the `page_size` to the number of search results for related. When setting the `page_size`, we also check for the number to be between 0 and 500, and when the number of search results is more than that, the `/related` endpoint returns a 400 error:

https://github.com/WordPress/openverse-api/blob/13673ee8c196c9a6b786a1322d6d65b97c6ca708/openverse_api/catalog/api/views/media_views.py#L97-L108

This is the code that throws the error, in `openverse_api/catalog/api/utils/pagination.py`:
```
 @page_size.setter
    def page_size(self, value):
        logging.info(f'Setting page size to {value}')
        if value is None:
            return
        value = int(value)  # convert str params to int
        if value <= 0 or value > 500:
            raise get_api_exception("Page size must be between 0 & 500.", 400)
        self._page_size = value
```

obulat:

## Fixes

Fixes #393 by @obulat

## Description

This PR sets the page_size for `/related/` endpoint to a constant of 10. This is not elegant, the constant should probably be set elsewhere, but I'm not familiar enough with the API codebase to find the correct place.

## Testing Instructions

Run the api (just up/just init), and then go to `http://localhost:8000/v1/images/41d51a9c-cdf5-4ca6-acfc-ebd63f807ecf/related/` 
You should see that although the result count is over 1200, the page_size is set to 10, and there is no error.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/openverse_api/catalog/api/views/media_views.py
+++ b/openverse_api/catalog/api/views/media_views.py
@@ -105,7 +105,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
             )
             self.paginator.result_count = num_results
             self.paginator.page_count = 1
-            self.paginator.page_size = num_results
+            self.paginator.page_size = 10
         except ValueError as e:
             raise get_api_exception(getattr(e, "message", str(e)))
 
--- a/openverse_api/catalog/api/views/media_views.py
+++ b/openverse_api/catalog/api/views/media_views.py
@@ -105,6 +105,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
             )
             self.paginator.result_count = num_results
             self.paginator.page_count = 1
+            # `page_size` refers to the maximum number of related images to return.
             self.paginator.page_size = 10
         except ValueError as e:
             raise get_api_exception(getattr(e, "message", str(e)))


openverse_api/catalog/api/views/media_views.py
```diff
@@ -105,7 +105,7 @@ def related(self, request, identifier=None, *_, **__):
             )
             self.paginator.result_count = num_results
             self.paginator.page_count = 1
-            self.paginator.page_size = num_results
+            self.paginator.page_size = 10
```
AetherUnbound, line 108: We should probably add a note here saying that this used to be `num_results` and may need to be set back!

zackkrida, line 108:
```suggestion
            # `page_size` refers to the maximum number of related images to return.
            self.paginator.page_size = 10
```

zackkrida (approved):
Added a comment suggestion, LGTM!

AetherUnbound (approved):

PR #408, opened 2021-12-14
Add docker volumes for local development

### LINKED ISSUE #407: Docker volumes for local development
## Problem

Once the docker-compose stack is taken down, the API, catalog, and Elasticsearch all need to be re-initialized once it is started up again.

## Description

We should add named docker volumes to these 3 services so the stack can be safely restarted without losing data.

## Implementation

AetherUnbound:
## Fixes

Fixes #407 by @AetherUnbound

## Description

This PR adds docker volumes for both DBs and ES.

## Testing Instructions

1. `just init`
1. `just down`
1. `just up` - databases should be migrated and indices should still exist in ES

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,8 @@ services:
       - postgres/env.docker
     healthcheck:
       test: "pg_isready -U deploy -d openledger"
+    volumes:
+      - api-postgres:/var/lib/postgresql/data
 
   thumbs:
     image: willnorris/imageproxy
@@ -20,6 +22,7 @@ services:
     ports:
       - "5433:5432"
     volumes:
+      - catalog-postgres:/var/lib/postgresql/data
       - ./sample_data:/sample_data
     env_file:
       - postgres/env.docker
@@ -113,3 +116,7 @@ services:
       - ./analytics:/analytics
     env_file:
     - ./analytics/env.docker
+
+volumes:
+  api-postgres:
+  catalog-postgres:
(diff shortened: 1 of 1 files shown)


sarayourfriend (commented):
Unfortunately I can't fully test this because I'm running into this issue, causing the web container to never be healthy and so `just init` runs forever:

```
web_1               |   File "/openverse_api/catalog/api/serializers/audio_serializers.py", line 214, in AudioWaveformSerializer
web_1               |     points = serializers.ListField(serializers.FloatField(min_value=0, max_value=1))
web_1               | TypeError: ListField.__init__() takes 1 positional argument but 2 were given
```

However, this approach does match what I've seen done for volumes for these services, so I think they make sense. The volumes are also correctly being created and torn down, even if I can't get the web container to run.

krysal (approved):
It does work for me. Perhaps we should use `stop` instead of `down` to not delete the container also, going along the same vein of this change.

zackkrida (approved):
Nice! Can confirm that the data is preserved and the volumes are created successfully:
```shell
❯ docker volume list | rg 'api_'
local     api_api-postgres
local     api_catalog-postgres
local     api_es-data
```

zackkrida:
@sarayourfriend does `just down && just recreate && just init` remove that error?

AetherUnbound:
> It does work for me. Perhaps we should use `stop` instead of `down` to not delete the container also, going along the same vein of this change.

I have to admit, I don't use stop that frequently :sweat_smile: Maybe I should, but it seems to be part of the "docker-compose paradigm" to use `up` and `down`!

PR #411, opened 2021-12-16
Add sentry

### LINKED ISSUE #399: Add Sentry to API
## Problem

We don't have any visibility into the API service. Sentry would be a good and easy first step.

## Description

Let's add Sentry. Long term we have goals of adding other monitoring but Sentry is a good and easy first step.

## Implementation

- [x] 🙋 I would be interested in implementing this feature.

sarayourfriend:
## Fixes

Fixes #399  by @sarayourfriend 

## Description

Adds basic Sentry monitoring to the API to capture unhandled exceptions. This is the bare minimum monitoring we should have.

Something is causing the dead-link filters integration test to fail. I'm not sure how to fix this, hopefully someone else can help debug the issue. It might have to do with DRF being upgraded from 3.12.x to 3.13.0 due to installing the `sentry-sdk` library.

## Testing Instructions

The only way to properly test this is to checkout this branch and run `just recreate && just init` then cause an error to happen (easiest way is through a bad code change that will only cause a runtime error) and verify it shows in the Sentry dashboard. The dashboard is private so if you need access to it ping me and I'll share how to access it.

You'll also need to run the app ~in~ without DEBUG mode or change the conditional in settings.py to allow Sentry to run locally. Of course we don't want Sentry errors to populate outside of production normally.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal:
I thought it was due to the change of the sample data but the same test that is falling here is the culprit in my PR too 🤔 Must be something else.

About the last part, my idea for Sentry is that it will mainly help spot and trace errors in production, so shouldn't it rather be used and tested with DEBUG mode turned off?

sarayourfriend:
> About the last part, my idea for Sentry is that it will mainly help spot and trace errors in production, so shouldn't it rather be used and tested with DEBUG mode turned off?

Sorry I miswrote the PR description! 

https://github.com/WordPress/openverse-api/pull/411/files#diff-6a4fc5f5b4ffed84bddc080d0a6c5aeb5db669ee9ff3ea5fa06b171e8a5980afR334

DEBUG is off in production so Sentry will only be enabled in production. To test it locally you need to turn DEBUG _off_ not on, whoops!

openverse_api/env.docker
```diff
@@ -16,3 +16,5 @@ UPSTREAM_DATABASE_PORT="5432"
 SEMANTIC_VERSION="1.0.0"
 
 ELASTICSEARCH_URL="es"
+
+SENTRY_DSN="https://08f4706d16004f57bcd37eb907bfc2e7@o787041.ingest.sentry.io/6107216"
```
krysal, line 20: Looks redundant to have this env var here and in the `env.example` too, in fact, I had to remove the Django variables from here in order to turn off the DEBUG mode. Can we leave this file only for docker-specific variables?

krysal (approved):
Works great! Even requesting an inexisting page that returns a 404 is shown in the Sentry dashboard, and tests are passing. This should help a lot :)

openverse_api/env.docker
sarayourfriend: Whoops! I misunderstood the reason for the `env.docker`. I'll remove it.

krysal (approved):
💯

openverse_api/catalog/settings.py
```diff
@@ -322,3 +324,17 @@
 
 # Whether to boost results by authority and popularity
 USE_RANK_FEATURES = config("USE_RANK_FEATURES", default=True, cast=bool)
+
+SENTRY_DSN = config(
+    "SENTRY_DSN",
+    default="https://08f4706d16004f57bcd37eb907bfc2e7@o787041.ingest.sentry.io/6107216",
```
AetherUnbound, line 330: This is our sentry URL, right? Do we want to have this as the default in-code, or rather just have it as a config variable? It feels weird to have this be codified in our codebase when other people might spin this up or work on it, but perhaps it's an unnecessary concern.

zackkrida (approved):
I agree with @AetherUnbound about the hardcoded variable, but we also have our Sentry DSN public in the frontend so I'm not blocking over it.

Also, and you'll remember this from the front-end, but I do really like having sentry configured for our staging environment too. Say, for example, we eventually have a test suite that runs against api-dev.openverse.engineering, the tests fail, and then we can go to Sentry and easily diagnose the issues before they make it to production.

sarayourfriend:
> Say, for example, we eventually have a test suite that runs against api-dev.openverse.engineering, the tests fail, and then we can go to Sentry and easily diagnose the issues before they make it to production.

Sounds good to me, but in that case I think we should have a separate project to send it to so that our prod sentry is separate from everything else, which the environment variable will allow us to do.

zackkrida:
That works for me! I do prefer the _environments feature_[^1] over distinct projects, but either would be a welcome addition. LGTM!


[^1]: https://docs.sentry.io/product/sentry-basics/environments/

sarayourfriend:
Ah yeah, environments does make more sense! I'm not sure what to pass for that tag in this case though because we don't actually seem to be passing an explicit environment, at least not in the env.template file. Do we pass something in production and staging?

zackkrida:
@sarayourfriend yep we'd want to explicity pass configuration to the sentry sdk like this:

https://docs.sentry.io/platforms/python/configuration/environments/

```py
import sentry_sdk
sentry_sdk.init(environment="production")
```

I would think 'production' in production and 'staging' in staging, for parity with the frontend.

justfile
```diff
@@ -174,6 +174,10 @@ dj args="":
 stats media="images":
     curl "http://localhost:8000/v1/{{ media }}/stats/"
 
+# Attach to ipython
+ipython:
+    docker-compose exec web ipython
```
dhruvkb, line 179: Instead of pure `ipython`, I believe `python manage.py shell` would be more useful as it sets up some Django magic automatically.
```suggestion
    docker-compose exec web python manage.py shell
```

openverse_api/catalog/api/serializers/audio_serializers.py
```diff
@@ -211,7 +211,9 @@ def create(self, validated_data):
 
 class AudioWaveformSerializer(serializers.Serializer):
     len = serializers.SerializerMethodField()
-    points = serializers.ListField(serializers.FloatField(min_value=0, max_value=1))
+    points = serializers.ListField(
+        child=serializers.FloatField(min_value=0, max_value=1)
+    )
```
dhruvkb, line 216: I thought Black was configured to break lines at 88 characters. I might be wrong.

openverse_api/catalog/settings.py
dhruvkb: I feel that instead of a default, the fallback behaviour should be to turn off Sentry when the variable is not supplied.

dhruvkb (changes requested):

dhruvkb (approved):
Approving because my feedback was only made up of non-blocking suggestions.

sarayourfriend:
@zackkrida yes, I understand how to pass the variable to sentry, but I don't know how to tell what environment we're in because it doesn't look like we're passing any explicit environment variable. There's nothing like `NODE_ENV` as far as I can tell (just `DEBUG` but that's a boolean, so not appropriate for determining exactly what environment we're currently running in).

AetherUnbound:
While we don't have this set up now, I think it would be pretty easy to pass the environment to the API! Currently when setting up an API worker, we [set a bunch of environment variables](https://github.com/Automattic/openverse-infrastructure/blob/main/modules/services/openverse-api/init.tpl#L19-L53). While we aren't setting the environment there, we are using the `staging_environment` variable elsewhere in the file ([example](https://github.com/Automattic/openverse-infrastructure/blob/main/modules/services/openverse-api/init.tpl#L159)). It should be straightforward to export that directly (possibly as `DJANGO_ENVIRONMENT` or `DJANGO_ENV`?) and use that as the sentry environment. What do you think?

sarayourfriend:
`DJANGO_ENV` or `PYTHON_ENV` (might be nice to be more portable so we can use a similar set-up in other Python projects?) sounds good to me!

openverse_api/catalog/settings.py
```diff
@@ -37,6 +37,8 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = config("DJANGO_DEBUG_ENABLED", default=False, cast=bool)
 
+PYTHON_ENV = config("PYTHON_ENV", default="development")
```
AetherUnbound, line 40: I think we use [`dev`](https://github.com/Automattic/openverse-infrastructure/blob/main/environments/dev/main.tf#L62) and [`prod`](https://github.com/Automattic/openverse-infrastructure/blob/main/environments/prod/main.tf#L65) elsewhere, should we keep this consistent?

sarayourfriend: Ah, sure. We use the full names in the frontend when using it to change runtime behavior... I wonder if that's even working if we're using abbreviations in the terraform variables.

I actually don't think they're even related, cause `NODE_ENV` isn't being set anywhere in terraform as far as I can tell.

krysal: Also, the default here should be `production`/`prod` whatever string we end up using, to be consistent with `DEBUG`. I think the best is explicitly to indicate the environment in the `.env` file.

sarayourfriend: @krysal Agreed to align the default with `DEBUG`. If it's best to define it explicitly though, should we exclude the default altogether?

krysal: It could be without a default, I don't have a strong preference here. Django defaults point to production. Our default in the `env.example` can be `development` for quicks setups.

sarayourfriend: Okay, yeah I think it makes sense to have the `env.example` be dev focused with the defaults be production focused.

PR #421, opened 2021-12-21
Data refresh slack notifications

### LINKED ISSUE #412: Ingestion server slack alerts
## Problem

It's difficult to track where the ingestion server is in its processing while running the data refresh.

## Description

It would be ideal if we could receive slack messages any time the data refresh moves onto a new step. We should also receive messages if the data refresh fails.

## Implementation

AetherUnbound:
## Fixes

Fixes #412 by @AetherUnbound

## Description

This PR adds a very simple slack notification mechanism for the ingestion server. During data refresh, the ingestion server will give periodic updates in slack regarding the progress of a given data refresh. If the refresh fails at any point, the ingestion server will also send the exception to Slack.

### Examples:
image
image
image

## Testing Instructions

1. Set the `SLACK_WEBHOOK` variable in `env.docker`
1. Spin up the stack
1. Run `just ingest-upstream`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/ingest.py
+++ b/ingestion_server/ingestion_server/ingest.py
@@ -30,7 +30,7 @@ from ingestion_server.queries import (
     get_fdw_query,
     get_go_live_query,
 )
-
+from ingestion_server import slack
 
 UPSTREAM_DB_HOST = config("UPSTREAM_DB_HOST", default="localhost")
 UPSTREAM_DB_PORT = config("UPSTREAM_DB_PORT", default=5433, cast=int)
@@ -266,6 +266,9 @@ def reload_upstream(table, progress=None, finish_time=None, approach="advanced")
     """
 
     # Step 1: Get the list of overlapping columns
+    slack.message(
+        f"Starting data refresh on `{table}` | Next: copying data from upstream"
+    )
     downstream_db = database_connect()
     upstream_db = psycopg2.connect(
         dbname=UPSTREAM_DB_NAME,
@@ -296,6 +299,9 @@ def reload_upstream(table, progress=None, finish_time=None, approach="advanced")
         copy_data = get_copy_data_query(table, shared_cols, approach=approach)
         log.info(f"Running copy-data query: \n{copy_data.as_string(downstream_cur)}")
         downstream_cur.execute(copy_data)
+
+    next_step = "starting data cleaning" if table == "image" else "re-applying indices & constraints"
+    slack.message(f"Data copy complete for `{table}` | Next: {next_step}")
     downstream_db.commit()
     downstream_db.close()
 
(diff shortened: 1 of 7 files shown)


sarayourfriend:
@AetherUnbound you said that in real usage these notifications would likely be sent hours apart right? Just wanting to confirm that, otherwise I might suggest somehow updating the existing message for successful passes or something instead of sending a new one (but still sending a new message if it fails) 😅 

Should failures ping `@here` or something like that? How important are they to take immediate action on?

AetherUnbound:
Yes! In production, it'll be a few hours between each step. For example, here are the rough runtimes based on the `image` table:
- Data copy: ~5 hours
- Data cleaning: ~14 hours
- Re-apply index: ~6 hours
- Elasticsearch reindex: ~6.5 hours
- Elasticsearch promotion: ~2 hours

As for failures, I don't think we need to `@here`. The way the data refresh is built, it can fail anywhere along the way and production won't be affected, so there isn't a ton of urgency to responding to issues.

dhruvkb (approved):
LGTM 👍

sarayourfriend (approved):

ingestion_server/ingestion_server/ingest.py
```diff
@@ -319,6 +319,9 @@ def reload_upstream(table, progress=None, finish_time=None, approach="advanced")
             f"_Next: re-applying indices & constraints_"
         )
 
+    # The server sometimes hangs on or before this next step. Adding this debug
+    # message here because apparently that unblocks it? May have to do with GC.
+    log.debug("Getting downstream cursor again")
```
sarayourfriend, line 324: Could try verifying this using `gc.collect`?

AetherUnbound: I'm guessing that any additional call made here, prior to the database connection call, would unblock this step. Work a shot though :woman_shrugging:

sarayourfriend: Hmmm... any additional call... would that include like `int("0")` or something? I wonder what just calling a function could be causing it to clear 🤔

AetherUnbound: I'm not sure! What I noticed is that all of the following would cause code execution to continue as expected:
- `log.info`
- `log.debug`
- `print`

I'm find leaving in the GC and seeing if it pops up again. It's frustrating because it's so intermittent!

sarayourfriend: Interesting. I wonder if it's somehow IO related?

AetherUnbound: Maybe! I'm really at a loss

PR #414, opened 2021-12-17
Update sample data for images with missing columns filled

krysal:
## Fixes

Fixes #270 by @dhruvkb 

## Description

This PR adds new sample data for images with `category`, `filesize` and `filetype` columns filled. The data comes from the recently updated Flickr and StockSnap scripts (2500 rows each) using this query on the catalog DB:

```sql
SELECT
	identifier, created_on, updated_on, ingestion_type, provider, source, 
	foreign_identifier, foreign_landing_url, url, thumbnail, width, height, filesize
	license, license_version, creator, creator_url, title, meta_data, tags, watermarked,
	last_synced_with_source, removed_from_source, filetype, category, standardized_popularity
FROM (
	SELECT *, ROW_NUMBER() OVER (
		PARTITION BY provider ORDER BY standardized_popularity DESC
	) AS row_num
	FROM image_view
) AS T
WHERE row_num <= 2500;
```

I also updated the tests accordingly. The only test I'm still not sure how was produced/mocked before is `test_dead_links_are_correctly_filtered`, is the only one that still fails. Perhaps @dhruvkb do you have a suggestion for this? Would be really useful.

## Testing Instructions

See the results of the test running:
```
just api-test
``` 
And/or explore the new data.

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
[...]

PR #422, opened 2021-12-21
Make API tests pass and reduce flakiness

### LINKED ISSUE #416: `test_dead_links_are_correctly_filtered` Flakiness
## Description

Based on reported issues in #411  and #414, the `test_dead_links_are_correctly_filtered` integration test has some flakiness that can cause unknown failures. I am not certain how to replicate this problem and cannot provide detailed information.

dhruvkb:
## Fixes

Fixes #416 by @zackkrida 

Fixes broken tests in #414 by @krysal 

## Description

This PR
- fixes the broken dead-link filtering test in PR #414 
- uses mocks and patching instead of actual web requests to more reliably simulate dead links
- reorganises tests to fix tests broken by the above two changes

## Testing Instructions

Let CI pass.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
These changes finally got [the tests to pass](https://github.com/WordPress/openverse-api/actions/runs/1606193050).

analytics/tests.py
```diff
@@ -25,7 +25,7 @@
 
 API_URL = os.getenv("ANALYTICS_SERVER_URL", "http://localhost:8090")
 session_id = "00000000-0000-0000-0000-000000000000"
-result_id = "29cb352c-60c1-41d8-bfa1-7d6f7d955f63"
+result_id = "cdbd3bf6-1745-45bb-b399-61ee149cd58a"
```
sarayourfriend, line 28: Why is this change necessary? Did we break something with the previous result id that we should be aware of for the future?

openverse_api/test/dead_link_filter_test.py
```diff
@@ -0,0 +1,137 @@
...
+            responses.append(mocked_res)
+        return responses
+
+    return patch("grequests.map", side_effect=grequests_map)
+
+
+@pytest.mark.django_db
+def test_dead_link_filtering(client):
+    _patch_redis().start()
+    mocked_map = _path_grequests().start()
```
sarayourfriend, line 40: Why use the patches this way? If we do, we need to call `stop` on them: https://docs.python.org/3/library/unittest.mock.html#patch-methods-start-and-stop

Using the decorator syntax might be preferrable though:

```suggestion
@pytest.mark.django_db
@_patch_redis()
@_path_grequests()
def test_dead_link_filtering(mocked_map, mocked_redis, client):
```

I think something like that would work? If not then we'll need to carefully call `.stop` on the patch result and make sure that it is called no matter where the test fails (as each assertion wil cause the method to break out and `stop` at the end wouldn't get called)

sarayourfriend (changes requested):
Blocking on the `start/stop` patch issue.

analytics/tests.py
dhruvkb: As far as I know, the ID needs to match the identifier used in the API tests so that the stats exist. It's a result of all the integration tests being deeply dependent on each other #423.

openverse_api/test/dead_link_filter_test.py
dhruvkb: The decorator syntax works great, thanks for recommending it! Implemented the change in 49373fd.

sarayourfriend (approved):
LGTM! Thanks for taking the time to figure out why these were flaking 🎉

krysal (approved):
**Amazing**. Thanks, @dhruvkb for such improvements in these tests! 🌟

krysal:
I'll merge it, so #414 can be closed by merge too :)

PR #442, opened 2022-01-03
Refactor `analytics`

dhruvkb:
## Description

This PR
- refactors the analytics codebase to more closely match the structure of ingestion-server
- refactors the ingestion-server API to reflect improvements from analytics
- sets up automatic swagger and ReDoc generation
- fixes an ORIGIN bug (should not contain path) and completes a todo task
- improves code quality

## Checklist

- [x] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [x] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [x] My commit messages follow [best practices][best_practices].
- [x] My code follows the established code style of the repository.
- [x] I added or updated tests for the changes I made (if applicable).
- [x] I added or updated documentation (if applicable).
- [x] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

</details>

PR #455, opened 2022-01-04
Copy operational files into final docker images

AetherUnbound:
## Description

Presently, we are unable to launch the ingestion server in production using only the published image. The current setup assumes that the operational files will be mounted into the containers when they are spun up, which may not be the case for all services running in production.

This PR adds steps in each `Dockerfile` to copy all of the necessary operational files into the image so they are available for the container on its own.  Images can now be spun up on their own and run successfully without any attached volumes. This ensures that the images that are pulled from our container registry can be launched directly without requiring a copy of the code locally. It also means that our final images are immutable (at least in production), so we won't have any unexpected changes at runtime even if the repository code were to change.

I also added a `just build` recipe for assistance with this work.

## Testing Instructions

1. Comment out all of the code volume mounts (e.g. `./api:/api`) in the `docker-compose.yml`
1. Run `just build`
1. Run `just up`
1. Observe that the stack starts successfully even though the local code is not attached as a volume to the containers

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -65,9 +65,8 @@ RUN apt-get update \
       && rm -rf /var/lib/apt/lists/* \
     && mkdir -p /var/log/openverse_api/openverse_api.log
 
-# Copy the operational code into the final image
-COPY run.sh manage.py /api/
-COPY catalog/ /api/catalog/
+# Copy code into the final image
+COPY . /api/
 
 # Exposes
 # - 8000: Dev server for API Django app
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -65,6 +65,10 @@ RUN apt-get update \
       && rm -rf /var/lib/apt/lists/* \
     && mkdir -p /var/log/openverse_api/openverse_api.log
 
+# Copy the operational code into the final image
+COPY run.sh manage.py /api/
+COPY catalog/ /api/catalog/
+
 # Exposes
 # - 8000: Dev server for API Django app
 # - 3000: Sphinx live server
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -65,7 +65,7 @@ services:
 
   web:
     build: ./api/
-    image: api
+    image: openverse-api
     volumes:
       - ./api:/api
     ports:
@@ -82,7 +82,7 @@ services:
 
   ingestion_server:
     build: ./ingestion_server/
-    image: ingestion_server
+    image: openverse-ingestion_server
     command: gunicorn -c ./gunicorn.conf.py
     ports:
       - "8001:8001"
@@ -99,7 +99,7 @@ services:
 
   indexer_worker:
     build: ./ingestion_server/
-    image: ingestion_server
+    image: openverse-ingestion_server
     command: gunicorn -c ./gunicorn_worker.conf.py
     ports:
       - "8002:8002"
@@ -115,7 +115,7 @@ services:
 
   analytics:
     build: ./analytics/
-    image: analytics
+    image: openverse-analytics
     ports:
       - "8090:8090"
     volumes:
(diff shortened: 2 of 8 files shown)


zackkrida (approved):
Very nicely commented, thank you!

dhruvkb:
So when we attach the same directory as a volume during dev, the copied code is overwritten?

AetherUnbound:
> So when we attach the same directory as a volume during dev, the copied code is overwritten?

Yes! The dev version will have a "live" version of the code, while the prod version will have a static set available within the image. More information [can be found here](https://docs.docker.com/storage/bind-mounts/#mount-into-a-non-empty-directory-on-the-container), and the Docker docs (say that five times fast) actually describe exactly the scenario that's happening here:

> If you bind-mount into a non-empty directory on the container, the directory’s existing contents are obscured by the bind mount. This can be beneficial, such as when you want to test a new version of your application without building a new image.

One other thing I forgot to mention - I prefixed the image names with `openverse-`. When the images get built, they were previously under the repository name `api`, `analytics`, etc. I have dozens of images on my machine, and so `api` becomes quite ambiguous :sweat_smile: this shouldn't affect the target tags or the artifacts that are built during CI/CD (as evidenced by the successful tests).

api/Dockerfile
```diff
@@ -65,6 +65,10 @@ RUN apt-get update \
       && rm -rf /var/lib/apt/lists/* \
     && mkdir -p /var/log/openverse_api/openverse_api.log
 
+# Copy the operational code into the final image
+COPY run.sh manage.py /api/
+COPY catalog/ /api/catalog/
```
dhruvkb, line 70: Could we write 
```dockerfile
COPY . /api
```
like we did for analytics?

docker-compose.yml
```diff
@@ -65,7 +65,7 @@ services:
 
   web:
     build: ./api/
-    image: api
+    image: openverse-api
```
dhruvkb, line 68: This is really good, thanks!

dhruvkb (approved):
The Docker docs (...Docker docs Docker docs Docker docs Docker docs) were very clear about it, thanks for the reference. LGTM!

api/Dockerfile
dhruvkb: ^ This would make the `COPY` statement
- same in all Dockerfiles
- identical to the `volumes` rules from the Docker-Compose file

AetherUnbound: My only concern with doing this is that it copies _all_ files in each directory, which includes a number of files which are irrelevant for production use (including all of the tests). I've generally been steered away from that, but it might not be _too_ much extra space. What do you think?

dhruvkb: Would a `.dockerignore` file help with that? We could reduce a lot of cruft from going into the context.

krysal (approved):
Tried it and works well. The prefix for images makes sense too 👍

api/Dockerfile
dhruvkb: I also submitted #442 which would make the `analytics/` code structure similar to `ingestion_server/`. If you could maybe rebase this PR on top of that one, the same exclusion strategy would work for both codebases.

AetherUnbound: Can do! This is also a good reminder to review that PR :smile:

AetherUnbound:
I've updated this to point at #442 so we can unify the approach. Tested with `just up`, `just api-test`, `just ing-testlocal` and `just up` (with commented out local code bind mounts) - all looked great!

dhruvkb (approved):
This is awesome!

PR #466, opened 2022-01-07
Add logger configuration to allow prod exception handling

AetherUnbound:
## Description

This PR adds some additional logging configuration to enable exception logging in production. Currently, when an exception or error occurs, we get unhelpful logs that look like this:

```
2022-01-07T13:06:57.964-08:00	Internal Server Error: /v1/audio/
2022-01-07T13:06:57.965-08:00	Internal Server Error: /v1/audio/
2022-01-07T13:06:57.984-08:00	Internal Server Error: /v1/audio/
2022-01-07T13:06:57.985-08:00	Internal Server Error: /v1/audio/
```

The above errors are HTTP 500 errors, but we aren't able to see any information about what is actually failing. The logging configuration in this PR ensures that:

1. Intentional logging within the service gets bubbled up appropriately
1. Django logging in general is set to `WARNING` in production
1. Exceptions are bubbled up to the logs appropriately in production

This does _not_ mean that users will see the debug view that Django provides in production. When we have `DEBUG=False`, all of that will be correctly hidden. This just makes it easier for us to troubleshoot production problems (particularly ones we can't replicate locally!).

I've used this config previously with great success! We also have the option to set up admin emails down the line where we'll receive an email on server errors. I think for a project of this size, that might be inadvisable and we would want a different way of aggregating/reporting errors.

## Testing Instructions

_Note that these steps are not to be committed and are only intended to replicate a production setup locally_
1. Set `DJANGO_DEBUG_ENABLED=False` in `api/env.docker`
1. Set the condition on line 54 of `settings.py` from `if DEBUG:` to `if True:`
1. Add an exception under the healthcheck view in `api/catalog/api/views/health_views.py` (I just added `raise ValueError("Whoops!")` above line 17)
1. `just up`
1. `just logs web`
1. Visit localhost:8000/healthcheck in your browser

If you follow these steps and are on `main`, you should receive an HTTP 500 error page and there should be nothing printed out in the logs (or at least not an exception). However if these steps are done on this branch, you should still receive the same generic HTTP 500 page in the browser, but the logs should contain the full exception traceback.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/logger.py
+++ b/api/catalog/logger.py
@@ -1,0 +1,82 @@
+from logging import LogRecord
+
+
+def health_check_filter(record: LogRecord) -> bool:
+    # Filter out health checks from the logs, they're verbose and happen frequently
+    return not ("GET /healthcheck" in record.getMessage() and record.status_code == 200)
+
+
+# Logging configuration
+LOGGING = {
+    # NOTE: Most of this is inherited from the default configuration
+    "version": 1,
+    "disable_existing_loggers": False,
+    "filters": {
+        "require_debug_false": {"()": "django.utils.log.RequireDebugFalse"},
+        "require_debug_true": {"()": "django.utils.log.RequireDebugTrue"},
+        "health_check": {
+            "()": "django.utils.log.CallbackFilter",
+            "callback": health_check_filter,
+        },
+    },
+    "formatters": {
+        "django.server": {
+            "()": "django.utils.log.ServerFormatter",
+            "format": "[{server_time}] {message}",
+            "style": "{",
+        },
+        "console": {
+            "format": "[%(asctime)s - %(name)s - %(lineno)3d][%(levelname)s] %(message)s",  # noqa
+        },
+    },
+    "handlers": {
+        # Default console logger
+        "console": {
+            "level": "INFO",
+            "filters": ["require_debug_true"],
+            "class": "logging.StreamHandler",
+            "formatter": "console",
+        },
+        # Add a clause to log error messages to the console in production
+        "console_prod": {
+            "level": "WARNING",
+            "filters": ["require_debug_false"],
+            "class": "logging.StreamHandler",
+            "formatter": "console",
+        },
+        # Handler for all other logging
+        "general_console": {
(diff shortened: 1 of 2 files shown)


obulat (approved):
Should be really helpful! 
Thank you for the detailed description and testing instructions, it's really interesting to read.

AetherUnbound:
Ugh, this is suffering the same hanging issue as I tried to address in #421 :/ I'll need to investigate this a bit more

api/catalog/settings.py
```diff
@@ -17,6 +17,8 @@
 from decouple import config
 from sentry_sdk.integrations.django import DjangoIntegration
 
+from .logger import LOGGING  # noqa: F401
```
zackkrida, line 20: `# noqa: F401`

What does this mean?

api/catalog/logger.py
```diff
@@ -0,0 +1,82 @@
+from logging import LogRecord
+
+
+def health_check_filter(record: LogRecord) -> bool:
+    # Filter out health checks from the logs, they're verbose and happen frequently
+    return not ("GET /healthcheck" in record.getMessage() and record.status_code == 200)
```
zackkrida, line 6: ❤️

zackkrida (approved):
Looks nice! @rbadillap I just want to point out that you were an assigned reviewer here; in the future just throw a comment on someone else's PR if your not able to review in within a day or two and we'll assign someone else. Thanks!

api/catalog/settings.py
AetherUnbound: This `LOGGING` variable is used within `settings.py` and is expected to be present when Django tries to access `settings.LOGGING`. However we don't use it anywhere else in the file, so our linter complains that it's an unnecessary import. This `noqa` disables that particular check (`F401`) for this line.

zackkrida: Thanks!

PR #474, opened 2022-01-20
Record transfer limit for data refresh

### LINKED ISSUE #472: Capture subset of data for refresh on staging
## Problem

If we would like to run the data refresh in development, we currently don't have a way of capturing a subset of the data present in the catalog to ingest into the API.

## Description

It would be great if we had some method of gathering a semi-randomized subset of data when we know that the instance is running on staging. Something that would only transfer 100k rows from the catalog to the dev API database.

## Additional context

One way to accomplish this would be by adding an `ORDER BY identifier` and then `LIMIT 100000` in the initial select from the catalog. Since the GUIDs should be unique and non-sequential, forcing them to be sequential might allow us to easily and quickly simulate a random subset (since the `identifier` column is indexed on the catalog).

## Implementation

- [x] 🙋 I would be interested in implementing this feature.

AetherUnbound:
## Fixes

Fixes #472 by @AetherUnbound

## Description

This PR adds a new variable, `DATA_REFRESH_LIMIT`, and some additional logic for interpreting an appropriate limit based on the environment the ingestion server is deployed in. In `prod` or `production`, the default limit is removed and all records are copied. In any other environment, at most 100k records are copied by default (although more or fewer can be specified using the aforementioned environment variable).

I also split the unit tests up and added tests for the altered query. This was a huge pain, because trying to convert the `psycopg2.SQL` objects into strings requires a database connection which is way too much overhead for such a simple test.

## Testing Instructions

`just ing-testlocal`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/ingestion_server/ingest.py
```diff
@@ -298,7 +306,15 @@ def reload_upstream(table, progress=None, finish_time=None, approach="advanced")
 
         # Step 3: Import data into a temporary table
         log.info("Copying upstream data...")
-        copy_data = get_copy_data_query(table, shared_cols, approach=approach)
+        environment = config("ENVIRONMENT", default="local").lower()
+        limit_default = 100_000
+        if environment in {"prod", "production"}:
```
dhruvkb, line 311: We should settle on one of these in the future.

dhruvkb (approved):
LGTM, except a note for the future.

ingestion_server/ingestion_server/ingest.py
AetherUnbound: Agreed, just trying to cover all my bases :sweat_smile: We've had some discussion about `local` as well in #475, I'll make an issue for unifying this everywhere.

zackkrida (approved):
Again, not the biggest 🐍 python authority but the code makes sense and I don't see any issues. Nice!

AetherUnbound:
So I ran into that issue where it hangs on the cleaning step _again_, and I was able to resolve it by removing the garbage collection steps :upside_down_face: I'll have to do a deep dive on that at some point

PR #476, opened 2022-01-20
Use the existing and configured env indicator

### LINKED ISSUE #475: Replace PYTHON_ENV with ENVIRONMENT
## Description

Our deployment infrastructure already configures an `ENVIRONMENT` variable and it is already used in the ingestion server code. Rather than introduce a new one that duplicates that, we can just use that everywhere.

Making this change will solve the issue we're seeing Sentry where staging is reporting itself as `production`.

## Resolution

sarayourfriend:
## Fixes

Fixes #475 by @sarayourfriend 

## Description

Replace `PYTHON_ENV` with the already used `ENVIRONMENT`.

## Testing Instructions

Checkout the branch and update your local instance (`just recreate && just init` should do it but will wipe local data) and verify the API runs.

Unit tests should also pass of course.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/settings.py
+++ b/api/catalog/settings.py
@@ -39,7 +39,7 @@ SECRET_KEY = config("DJANGO_SECRET_KEY")  # required
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = config("DJANGO_DEBUG_ENABLED", default=False, cast=bool)
 
-PYTHON_ENV = config("PYTHON_ENV", default="production")
+ENVIRONMENT = config("ENVIRONMENT", default="local")
 
 ALLOWED_HOSTS = [
     "api-dev.openverse.engineering",
@@ -318,5 +318,5 @@ if not DEBUG:
         integrations=[DjangoIntegration()],
         traces_sample_rate=SENTRY_SAMPLE_RATE,
         send_default_pii=False,
-        environment=PYTHON_ENV,
+        environment=ENVIRONMENT,
     )
--- a/api/env.template
+++ b/api/env.template
@@ -1,5 +1,4 @@
 PYTHONUNBUFFERED="0"
-PYTHON_ENV="development"
 
 DJANGO_SETTINGS_MODULE="catalog.settings"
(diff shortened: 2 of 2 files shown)


api/catalog/settings.py
```diff
@@ -39,7 +39,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = config("DJANGO_DEBUG_ENABLED", default=False, cast=bool)
 
-PYTHON_ENV = config("PYTHON_ENV", default="production")
+ENVIRONMENT = config("ENVIRONMENT", default="local")
```
krysal, line 42: Why did it change to "local"?

sarayourfriend: It's the default that is used by the ingestion_server when it accesses this same variable:

https://github.com/WordPress/openverse-api/blob/ad2f74a111af4ea35f939ec597f394c83f6dbc23/ingestion_server/ingestion_server/distributed_reindex_scheduler.py#L73

krysal: I think we should keep "production" and change to "staging" or the corresponding string in the staging server. "local" doesn't seem very useful as everything is local to its own machine.

AetherUnbound: Yea I agree that naming convention isn't too useful! :sweat_smile: I think `dev` or `staging` makes sense. A lot of the environment checks in the ingestion server are checking explicitly if in prod (`environment == "prod"`) so it hasn't mattered much what _not prod_ is.

AetherUnbound (approved):
Looks good! We can follow up with a future PR to change "local" to something clearer :slightly_smiling_face:

krysal (approved):

api/catalog/settings.py
sarayourfriend: I agree `local` is ambiguous.

In my mind (and indeed I've seen this in practice) `dev` is local development, `staging` is the pre-production deployment, `prod` is production, and `test` is when tests are running regardless of environment. Also variations on those strings exist. Node (and most JavaScript projects like React and Vue) uses the full `development` and `production` specifically for example, so I like the consistency of using that for the less-opinionated Python environment.

PR #510, opened 2022-02-07
Cache waveform data in database

### LINKED ISSUE #490: Cache waveform data
## Problem

Fetching waveform data from the API currently generates the waveform on every hit.

## Description

This should be cached to only be calculated the first time the waveform for an audio is requested.

We can use any caching strategy we want, but it'll have to be cached on disk rather than memory (I would think) to avoid filling up RAM and forcing the main application processes to use swap. Could we just cache it directly in the DB?

Django supports many caching strategies that allow using the Django provided `cache` object: https://docs.djangoproject.com/en/4.0/topics/cache/

## Implementation

sarayourfriend:
## Fixes

Fixes #490 by @sarayourfriend

## Description

This PR indefinitely caches waveform data in the postgres DB directly on the waveform model after the first time it is generated.

## Testing Instructions

I added a thorough unit test to test the `get_or_create_waveform` function. There were no existing integration tests for the waveform endpoint. Should we add some? I'd appreciate some guidance about how best to do that and prevent the call to the `audiowaveform` binary and an external call to retrieve the audio file (can we mock the return of the audio file from `requests.get`?)

You can test this manually by running the API locally following the instructions in the README and then visiting the http://localhost:8000/v1/audio/010356c6-ff57-44f1-bddc-675a2a1e5a38/waveform/ endpoint. The first time should take noticably longer than subsequent visits.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


CONTRIBUTING.md
```diff
@@ -46,6 +46,12 @@ cd openverse_api
 pipenv install
 ```
 
+Note: The following non-Python libraries or binaries are transient dependencies which will also need to be present on your computer for the project to run and install as expected:
+
+* `librdkafka`
+* `exempi`
+* `audiowaveform`
+
```
zackkrida, line 54: 👍

api/catalog/api/utils/waveform.py
```diff
@@ -128,3 +129,14 @@ def cleanup(file_name):
         os.remove(file_path)
     else:
         log.info("File not found, nothing deleted")
+
+
+def generate_peaks(audio) -> List[float]:
```
dhruvkb, line 134: As of Python 3.10, we can use `list[float]` as the type without needing to import `List` from `typing`.

api/catalog/api/models/audio.py
```diff
@@ -209,3 +223,26 @@ class Meta:
...
+
+class AudioWaveformAddOn(AbstractMediaAddOn):
+    audio = models.OneToOneField(
+        to=Audio,
+        on_delete=models.CASCADE,
+        related_name="waveform",
+        help_text=("The foreign key of the audio."),
+    )
+
+    peaks = ArrayField(
```
dhruvkb, line 236: Instead of the strict typing as an array of floats, we could possibly use a `jsonb` field? That allows us to tweak waveform date if needed in the future & prevents the need for an approximation of 1500.

dhruvkb (dismissed):
I'm not so about making a separate model for media add-ons. Imo, we should
- add the field to `Audio`, while making it nullable 
- make the ingestion server capable of copying only the common fields

dhruvkb (commented):
I meant to click "Comment", not <s>"Approve"</s>.

api/catalog/api/models/audio.py
sarayourfriend: I thought about using `jsonb` but I actually prefer the strict typing. I wonder if there are any performance characteristics we should consider too, if inserting an array is faster than jsonb over time? For a single client it's possible that up to 10 waveform requests will be made for an audio search result page which could cause a pretty large number of transactions under heavy traffic.

sarayourfriend:
I think I prefer that solution too @dhruvkb. I probably over-engineering this current approach.

The first bullet point is straightforward to implement. What are the implications of the second one? I'm not very familiar with the ingestion server so I don't know what the effect of changing the model from the API side would be.

api/catalog/api/utils/waveform.py
sarayourfriend: Ah, I thought so but then I forgot to go back and change it!

dhruvkb:
Iirc the ingestion server already has the ability to identify columns with matching names and only copy them. Adding a Django field with `blank=True, null=True` should be sufficient.

zackkrida:
> Iirc the ingestion server already has the ability to identify columns with matching names and only copy them.

That's correct, it lists columns from both DBs and only copies the shared ones, so this should work. Tagging @AetherUnbound because she's looked at that semi-recently and can confirm that our understanding is correct.

api/catalog/api/models/audio.py
dhruvkb: To boost the performance, I was at some point considering compressing the JSON data of the waveform and storing it as a text field in the DB 😄. I'm not sure which of our choices is most performant, maybe @AetherUnbound has more knowledge in that area, but imo, we should aim for max performance even if the cost is absence of validation.

sarayourfriend:
Okay! Updated to add the waveform directly to the Audio model. Everything else still works the same.

AetherUnbound (approved):
I like the field solution! And yes, the data refresh would only be copying over common columns. I believe the implication of this is that we'd have to rebuild the waveforms after a data refresh, since the entire table is replaced.

RE the typing question, I'm really not sure! My gut says that matching the types would be most ideal, but it depends on how we're using it. If it's *just* being passed to the frontend to be hydrated into a JSON array, then maybe it makes sense to store it as JSON or even a string :woman_shrugging:. On the other hand, if we're trying to do anything more nuanced with the column the stricter typing is probably better (and [may help out the planner in those cases](https://stackoverflow.com/a/59125847/3277713)).

sarayourfriend:
> I believe the implication of this is that we'd have to rebuild the waveforms after a data refresh, since the entire table is replaced.

Eek! This makes me not want to share the table (and also to put a big stinking warning in the code on the model that that will happen to any data stored in the table that isn't directly from the catalog)!

AetherUnbound:
Yea...it's not ideal. Putting it in a separate table could help reduce the duplicated computation, but it may also cause some drift to occur if an audio record gets removed from the catalog for some reason while the waveforms get left behind.

sarayourfriend:
> but it may also cause some drift to occur if an audio record gets removed from the catalog for some reason while the waveforms get left behind.

Hmm, we wouldn't be able to use cascading deletes? Or does the refresh totally wipe out the audio table and rebuild it?

sarayourfriend:
I'm not sure what the right answer is then. It's a waste to generate the waveform more than once for a given file, right? As the audio dataset gets larger it seems senseless for the API to get hit harder every time a data refresh happens because it has to regenerate waveforms for every single audio result.

dhruvkb:
Can the ingestion process build the waveforms and also copy them to the API?

AetherUnbound:
Ha, jinx @dhruvkb :grin: 

If we wanted the waveforms cached immediately, I think this might be an okay strategy to start with. But we'll have to take a look at it again down the road as our volume of audio data expands.

sarayourfriend:
How does the thumbnail cache work? The waveform data seems to be basically the same thing as a thumbnail cache but for audio.

dhruvkb:
Thumbnails are built on demand and then cached by the thumbnail-server without being stored to the database. The thumbnail server has that [functionality built-in](https://github.com/willnorris/imageproxy#cache).

sarayourfriend:
I see. It'd be nice to avoid spinning up a whole new service just to cache waveform data.

sarayourfriend:
Is it absolutely necessary for the data-refresh to tear down and completely recreate the table?

AetherUnbound:
> Is it absolutely necessary for the data-refresh to tear down and completely recreate the table?

I don't think so, there's some further discussion of that here: https://github.com/WordPress/openverse/discussions/120. The TL;DR is that we'll need to do some performance testing to try any other strategies, as they may be prohibitively time-consuming.

AetherUnbound:
Essentially, yes. It builds a table with all the data from the catalog but only the columns in the API, applies all the constraints the current API tables have, then swaps them. This is done because the API has a bunch of indices and those indices can affect write time on the catalog end.

sarayourfriend:
@AetherUnbound @dhruvkb this is ready for another review when y'all have a chance. Thanks!

api/catalog/api/models/audio.py
```diff
@@ -97,6 +97,46 @@ class Meta:
         abstract = True
 
 
+class AudioAddOn(OpenLedgerModel):
+    audio_identifier = models.UUIDField(
+        db_index=True,
+        unique=True,
+        blank=False,
+        null=False,
```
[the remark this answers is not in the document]
sarayourfriend: Hmm, interesting. I guess I usually think of PK as the canonical key for the table, but I suppose we could ditch the default `id` column for this one anyway... it's basically a "join" table of sorts and is only ever accessed through the audio identifier anyway.

I'll update this tomorrow!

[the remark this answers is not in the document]
AetherUnbound: Ah! That's a fair point, I forget that Django does some magic there for you :sweat_smile: But agreed, I think it makes sense to have that be the PK!

AetherUnbound (approved):
Looks great!

api/catalog/api/models/audio.py
```diff
@@ -96,6 +97,40 @@ class Meta:
...
+    This cannot be a "ForeignKey" or "OneToOneRel" because the refresh process
+    wipes out the Audio table completely and recreates it. If we made these a FK
+    or OneToOneRel there'd be foreign key constraint added that would be violated
+    when the Audio table is recreated.
+
+    The index is necessary as this column is used by the Audio object to query
+    for the relevant add on.
+
+    The refresh process will also eventually include cleaning up any potentially
+    dangling audio_add_on rows.
```
zackkrida, line 115: This is a great piece of documentation!

zackkrida (approved):
This looks straightforward and like a nice solution. I'm definitely curious if we'll ever be able to avoid a full refresh of the data tables in the future; until then lgtm!

api/catalog/api/models/audio.py
```diff
@@ -96,6 +97,40 @@ class Meta:
         abstract = True
 
 
+class AudioAddOn(OpenLedgerModel):
+    audio_identifier = models.UUIDField(
```
dhruvkb, line 101: If this is the primary key of this model we should drop the `audio_` prefix.

api/catalog/api/utils/waveform.py
```diff
@@ -128,3 +129,14 @@ def cleanup(file_name):
...
+    file_name = None
+    try:
+        file_name = download_audio(audio.url, audio.identifier)
```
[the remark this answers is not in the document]
dhruvkb: @AetherUnbound wouldn't the `finally` clause below run in the event of an exception?

dhruvkb (approved):
Looks mostly fine, except for one non-blocking suggestion.

api/catalog/api/models/audio.py
dhruvkb: Should also note that leaving it as it is wouldn't be wrong either, as it refers to the Audio model instance.

api/catalog/api/utils/waveform.py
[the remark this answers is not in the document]
AetherUnbound: It does! I just mention `tempfile` because then we wouldn't have to do the cleanup ourselves, in every place where `download_audio` is called.

PR #524, opened 2022-02-21
Send peak data in search results and details

dhruvkb:
## Description

This PR
- adds peaks data to the search results
- adds peaks data to the single results
- enables compression to minify the now huge JSON responses

## Testing Instructions

0. Check out the PR and initialise everything: `just init`
1. Visit the search results page.
2. See no peaks.
3. Visit the waveform page for some audio results to get the waveforms cached.
4. Visit the search results page again.
5. See peaks.
6. Open the details page for the audio with peaks.
7. See peaks.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/models/audio.py
+++ b/api/catalog/api/models/audio.py
@@ -192,11 +192,21 @@ class Audio(AudioFileMixin, AbstractMedia):
         except AudioSet.DoesNotExist:
             return None
 
-    def get_or_create_waveform(self):
+    def get_or_create_waveform(self, short_circuit=False):
+        """
+        Get the waveform peak data for the audio object from the associated
+        addon object. If the associated addon does not exist, it is created. If
+        the waveform is not saved, it is generated.
+
+        :param short_circuit: create the addon instance but not waveform
+        :return: the waveform peaks
+        """
         add_on, _ = AudioAddOn.objects.get_or_create(audio_identifier=self.identifier)
 
         if add_on.waveform_peaks is not None:
             return add_on.waveform_peaks
+        elif short_circuit:
+            return []
 
         add_on.waveform_peaks = generate_peaks(self)
         add_on.save()
--- a/api/catalog/api/serializers/audio_serializers.py
+++ b/api/catalog/api/serializers/audio_serializers.py
@@ -1,12 +1,14 @@
 from catalog.api.controllers.search_controller import get_sources
 from catalog.api.docs.media_docs import fields_to_md
 from catalog.api.models import AudioReport
+from catalog.api.models.audio import Audio
 from catalog.api.serializers.media_serializers import (
     MediaSearchRequestSerializer,
     MediaSearchSerializer,
     MediaSerializer,
     _validate_enum,
 )
+from elasticsearch_dsl.response import Hit
 from rest_framework import serializers
 
 
@@ -170,6 +172,15 @@ class AudioSerializer(MediaSerializer):
         lookup_field="identifier",
         help_text="A link to an endpoint that provides similar audio files.",
     )
+
+    # Add-on data
+    peaks = serializers.SerializerMethodField()
+
+    @staticmethod
+    def get_peaks(obj):
+        if isinstance(obj, Hit):
+            obj = Audio.objects.get(identifier=obj.identifier)
+        return obj.get_or_create_waveform(short_circuit=True)
 
 
 class AudioSearchSerializer(MediaSearchSerializer):
--- a/api/Pipfile
+++ b/api/Pipfile
@@ -49,6 +49,7 @@ python-decouple = "*"
 django-cron = "*"
 gunicorn = "*"
 sentry-sdk = "*"
+django-compression-middleware = "*"
 
 [requires]
 python_version = "3.10"
--- a/api/Pipfile.lock
+++ b/api/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "9e211732f922bb7dd945ee0681569c8fbf12a289b3800bff5f03cec40706da66"
+            "sha256": "47f87f998288dff2bce7d74e80296464233a1ef2582cdb709cfaea7cff8c0168"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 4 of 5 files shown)


zackkrida:
This makes me curious if in the future we should downsample the peak data in the list endpoints and then use the full dataset on the detail view only. Similarly to low-res thumbnails vs. higher quality single images.

AetherUnbound:
This looks good! I assume the compression middleware is plug-and-play?

dhruvkb:
Yeah I tested it locally for about 66% reduction on average.

zackkrida:
Another thought: we might want a follow-up PR to deprecate or outright remove the waveform endpoint.

dhruvkb:
Moving compression out of this PR to

- decrease scope
- prevent having to build containers
- investigate how the API responses are already compressed with Brotli

zackkrida:
Unscientific size comparison:

```bash
# Get the initial size
curl -so /dev/null http://localhost:8000/v1/audio/\?format\=json -w '%{size_download}'
21583

# Generate all waveform data
> curl http://localhost:8000/v1/audio/\?format\=json | jq '.results[].waveform' | xargs -L1 open

# Get the new size
❯ curl -so /dev/null http://localhost:8000/v1/audio/\?format\=json -w '%{size_download}'
201138
```

So about 9.3x bigger.

sarayourfriend:
Sorry I missed the discussion, but why are we removing the waveform endpoint and moving these into the peaks, especially if it will cause read performance issues (at least that was my understanding from the sync conversation we had earlier about this). I just missed the reason why we need to do this is all.

Couldn't a frontend (like our own) just make individual requests for each waveform? At least then it'd be spread across more than 1 worker and wouldn't block the results response while the waveforms are generated. For any UI consuming these they could use a placeholder waveform until the actual one loads (I guess like we're already doing).

But I'm sure I'm missing a lot of context about this anyway.

dhruvkb:
@sarayourfriend the main issue in that case would be that the one search result in the audio would be making N+1 (i.e. 21) requests to get the waveforms for every audio. 

It's usually better to serve a low-fidelity waveform with the results (as @zackkrida suggested [here](https://github.com/WordPress/openverse-api/pull/524#issuecomment-1047243471)) along with the results themselves. On the single audio result page, the frontend can request for the detailed waveform from the API with the rest of the details. But that means the waveform endpoint is no longer used at all.

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -171,6 +173,15 @@ class AudioSerializer(MediaSerializer):
         help_text="A link to an endpoint that provides similar audio files.",
     )
 
+    # Add-on data
+    peaks = serializers.SerializerMethodField()
```
sarayourfriend, line 177: Should we name this something less specific than `peaks` considering in the future other things could potentially live in that response?

api/catalog/api/models/audio.py
```diff
@@ -192,11 +192,21 @@ def audio_set(self):
...
+
+        :param short_circuit: create the addon instance but not waveform
+        :return: the waveform peaks
+        """
         add_on, _ = AudioAddOn.objects.get_or_create(audio_identifier=self.identifier)
 
         if add_on.waveform_peaks is not None:
             return add_on.waveform_peaks
+        elif short_circuit:
+            return []
```
sarayourfriend, line 209: Can you add a unit test for this branch?

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -171,6 +173,15 @@ class AudioSerializer(MediaSerializer):
...
+
+    @staticmethod
+    def get_peaks(obj):
+        if isinstance(obj, Hit):
+            obj = Audio.objects.get(identifier=obj.identifier)
+        return obj.get_or_create_waveform(short_circuit=True)
```
sarayourfriend, line 183: This will still cause n+1 query execution, specifically it will create `n` `create` queries.

sarayourfriend (changes requested):
I don't think this approach actually solves the issue, likely we need to update audio list serializer to batch create the audio add on objects (unless there's some Django magic going on that is already doing this, but I'm not familiar with anything like that).

api/catalog/api/serializers/audio_serializers.py
dhruvkb: This method field is specifically obtaining the audio peaks using the `get_peaks` function. It's not serializing the AddOn object.

api/catalog/api/serializers/audio_serializers.py
dhruvkb: There are potentially N database queries, yes, but those are on the primary key (indexed) and are considerably less expensive than N queries from the frontend to the API over HTTP.

api/catalog/api/serializers/audio_serializers.py
sarayourfriend: Sorry! I realized this later and forgot to remove this comment.

api/catalog/api/serializers/audio_serializers.py
sarayourfriend: > are considerably less expensive than N queries from the frontend to the API over HTTP

Is the concern here the HTTP responses? 20 extra HTTP requests per search request is pretty trivial compared to what Django and gunicorn can handle as far as I understand, but I agree that removing the need for them is better. However, we already make this number of requests for thumbnails anyway; if we're doing this kind of optimization we should reconsider how thumbnails are requested as well.

Likewise, the HTTP requests to the API do use the primary key to access the add on, the audio object identifier on the AddOn table is the PK for that table.

Am I misunderstanding the way that this PR makes is situation better aside from removing the additional HTTP requests though?

sarayourfriend:
> the main issue in that case would be that the one search query for audio would be making N+1 (i.e. 21) requests: 1 to get the results and then 20 for waveforms for every audio.

@dhruvkb can you add this goal to the PR description? Without an accompanying issue it's hard to understand the goal/context of this PR without that piece of information.

api/catalog/api/serializers/audio_serializers.py
dhruvkb: No @sarayourfriend your understanding is correct that the PR only removes 20 additional HTTP requests. It doesn't improve anything else.

sarayourfriend: Okay. I don't feel comfortable approving this approach as it actually adds an actual n+1 query problem to the audio results endpoint. In fact we're creating the objects in a loop that will block each iteration on the database interaction.

If others are fine with this then feel free to go ahead with it but please open an issue to address the n+1 query in the audio results view.

dhruvkb:
@sarayourfriend I've updated the PR because I share your concerns. The behaviour is now as follows:

**Case 1:** for all audio objects that have a waveform pre-computed and cached, the output is served in the results and details

**Case 2:** for all audio objects that do not have a waveform, we serve `[]` in the results and details

No create queries will be run from the serializers. Further we can pre-heat the cache by manually running the waveform computation loop from a shell so that case 1 is the more prevalent case.

zackkrida:
> No create queries will be run from the serializers. 

This makes me more comfortable with the approach in this PR. Otherwise I was going to suggest that we do, in fact, allow 20 network requests to happen on the frontend! 

In either case, we're going to want this peak data to exist first. Fetching 20 static json files of a few kb each from cloudflare is pretty cheap on the frontend; waiting for 20 waveforms to generate dynamically isn't. We have to remember that because openverse usage is relatively small for a search engine, that most queries are 'new', in the sense they've never been made before and aren't cached. Especially when you factor in every possible filter combination for every possible search.

dhruvkb:
Yeah, caching the details page is easier, because those are immutable. But we should not expect a cache hit on a search result page that can have all manners of filter combinations in the query.

zackkrida:
> But we should not expect a cache hit on a search result page that can have all manners of filter combinations in the query.

Just clarifying, each URL with different query params _is_ cached, and since the homepage doesn't apply filters by default anyone who searches from the homepage for common searches _is_ pretty likely to get a cached API response. Still, the more traffic Openverse gets the better our cache hit % should become.

sarayourfriend (approved):
LGTM! Would love to see a unit test for this if you don't mind.

Combined with #530 I think this approach makes sense for now.

zackkrida (approved):
lgtm! thanks again for visiting this issue so quickly and proactively with multiple solutions.

PR #527, opened 2022-02-22
Introducing 🦭 `awf`

dhruvkb:
## Description

This PR decouples the audio waveform generation capability from the API making it a standalone entity. This makes it possible to use the waveform code in the ingestion server and possibly also in the catalog.

Uses `awf`. [GitHub](https://github.com/dhruvkb/awf)

## Testing Instructions

0. Check out the PR and run the Django dev server.
1. Visit the waveform endpoints for some audio files.
2. You should get the expected response.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
Let me go over the advantages pretty quickly.
- <s>The old idea was to have this separate service take some load off the API by working in a manner similar to the image proxy.</s> This no longer applies after waveforms are cached in the API DB.
- It's written in Go using [Fiber](https://gofiber.io) so it's [crazy fast](https://docs.gofiber.io/extra/benchmarks) compared to Python. It can get even faster when we directly integrate with the `audiowaveform` C++ library instead of subprocessing it.
- It's a very small service (about 18MB) so it's not huge, it also helps keep the Python images lean.

A Python package would be awesome for all the reasons you mention but it can never be standalone. It needs the `audiowaveform` binary so it'll always be tightly coupled with where you install it.

As for the latency, in my manual testing with limited samples, I observed a performance gain when using `awf`. I'm not entirely sure but I think it's because `awf` makes up for the network latency by downloading the track and computing waveforms faster.

sarayourfriend:
> A Python package would be awesome for all the reasons you mention but it can never be standalone. It needs the audiowaveform binary and a bunch of libs so it'll always be tightly coupled with where you install it. The API Dockerfile had to be extensively reworked when adding audiowaveform to the API.

To be clear, this is the same situation the go library is in, it's just put into a container that already has those dependencies in it. The go program is just calling to the `audiowaveform` binary in the end and I suppose a go wrapper is probably faster at processing the response from the binary than Python. We're trading a few binary dependencies for a whole new HTTP service and docker container to maintain in our stack. It'd be curious to hear from @rbadillap and probably @AetherUnbound about whether that's an acceptable trade off (maybe it is, I can't speak to it having only observed our infrastructure from afar).

However, I'm surprised there's a measurable difference in performance for these reasons, I'd love to see an actual measurement done:

1. When comparing Go and Python processing what amounts to be around 1000 integers and 1000 floats, I'm struggling to imagine that there's a super clear difference aside from program start-up time (which wouldn't matter in the end, either way the application will have already been running, either as an HTTP service or otherwise), but would be willing to accept this fact, it wouldn't be that surprising
2. Even if 1 is true, that Go is significantly faster than Python at processing that small number of integers and floats, it's still marshaling them to JSON then transferring them over HTTP (which includes it's own overhead) which then has to be parsed on the Python side. Maybe Python HTTP and JSON parsing is blazing fast or something, but it would be strange to me that _that_ would be faster than just doing the number juggling in Python to begin with.

I suppose maybe the Go binary really does download audio files that much faster than `requests` does; in which case great! I do wonder just _how_ much of a difference there is there though and whether it's worth it (more on that at the end of this comment).

However, to these points:

>  It can get even faster when we directly integrate with the audiowaveform C++ library instead of subprocessing it.

You could write Python bindings for the audiowaveform library as well, in which case you'd never have to call out to another binary from our applications at all. Unlike a Go wrapper which (while cool) will always be external to our application, introducing the already-existing overhead we have in calling to another binary. I suppose you could speed up the awf _wrapper_ but I really am surprised there's a performance _gain_ from communicating over HTTP and JSON compared to just calling a local binary and parsing the output.

> It's a very small service (about 18MB) so it's not huge, it also helps keep the Python images lean.

I'm sure the compiled Go binary is small, but it still includes the audiowaveform binary and its dependencies which were already included in our application. If those fit inside an 18MB image then it's not a super significant difference to remove them from our Python images, in particular if we're trading those few MB for additional infrastructure complexity. Is there something else here I'm missing?

In the end, I think this is definitely cool and could be an interesting way to accomplish the decoupling goal, but I'm not super convinced that there's going to be a performance _gain_ from communicating over HTTP and would like to see measurements showing it one way or another (happy to help develop these). I am also concerned about the infrastructure complexity we're adding; in my opinion the performance increase of taking this approach would have to be pretty huge to outweigh the cost of maintaining the infrastructure for it (which includes everything from updating the Go wrapper and docker image when new versions of `audiowaveform` are released as well as actually provisioning the infrastructure to host it).

dhruvkb:
I see how adding an additional service and an additional language is a bad tradeoff for shaving a couple of seconds off. We can continue to use the Python-based solution for now and if need be, just move the Python code to a different part of the infrastructure.

sarayourfriend:
> I see how adding an additional service and an additional language and a non-full-fledged library from outside the main codebase is a bad tradeoff for shaving a couple of seconds off. We can continue to use the Python-based solution for now and if need be, just move the Python code to a different part of the infrastructure.

Well, hold up, I'm not saying that's definitely case! Just wanted to raise that and get input from the folks who are closer to our infrastructure. If we're able to detect a measurable difference and the cost for running it (from a human energy/power perspective) is low then this could be a good solution to this problem.

PR #530, opened 2022-02-23
Django command for generating waveforms

### LINKED ISSUE #529: Audio waveform cache-warming Django command
## Description

Related to #510

We now have a table where audio waveforms are cached, but this table is not yet fully populated. We have plans to move this processing upstream into the catalog (https://github.com/WordPress/openverse-catalog/issues/363). 

For the time being, we should create a Django command that can be run with `python manage.py <command>` which will iterate through all audio records and generate waveforms for them. Here is an outline of the necessary code:

```python
audios = Audio.objects.all()
with tqdm(total=audios.count()) as progress:
    for audio in audios:
        audio.get_or_create_waveform()
        progress.update(1)
```

The above code example uses `tqdm`, which I don't believe is a dependency we have right now. We could add it for this use and remove it after, since the progress bar and estimated time to completion are very useful. 

We have ~14k records, and a test run on staging showed that this could take anywhere from 12-48 hours.
image (3)

## Implementation

AetherUnbound:
## Fixes

Fixes #529 by @AetherUnbound

## Description

This PR adds a Django command, `generatewaveforms`, for generating waveforms of all audio records in order to populate the waveform cache.

I've added a new dependency, [`django-tqdm`](https://pypi.org/project/django-tqdm/), to make the output easier to read. We can remove this dependency once the API no longer generates waveforms.

**Command list**

```bash
root@c219265a4b39:/api# python manage.py --help

Type 'manage.py help <subcommand>' for help on a specific subcommand.

Available subcommands:

[auth]
    changepassword
    createsuperuser

[catalog]
    generatewaveforms

[contenttypes]
    remove_stale_contenttypes

...
```

**Help text**
```bash
root@c219265a4b39:/api# python manage.py generatewaveforms --help
usage: manage.py generatewaveforms [-h] [--version] [-v {0,1,2,3}] [--settings SETTINGS] [--pythonpath PYTHONPATH] [--traceback] [--no-color] [--force-color] [--skip-checks]

Generates waveforms for all audio records to populate the cache

options:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Verbosity level; 0=minimal output, 1=normal output, 2=verbose output, 3=very verbose output
  --settings SETTINGS   The Python path to a settings module, e.g. "myproject.settings.main". If this isn't provided, the DJANGO_SETTINGS_MODULE environment variable will be used.
  --pythonpath PYTHONPATH
                        A directory to add to the Python path, e.g. "/home/djangoprojects/myproject".
  --traceback           Raise on CommandError exceptions
  --no-color            Don't colorize the command output.
  --force-color         Force colorization of the command output.
  --skip-checks         Skip system checks.
```

**Sample output**
```bash
root@c219265a4b39:/api# python manage.py generatewaveforms
Generating waveforms for 5,000 records
  0%|                                                                                                                            | 1/5000 [00:00<16:19,  5.10it/s]
  Unable to process 67acc776-838c-4242-bb99-be9e0be061b6: Can't generate json format output from Unknown file format
  1%|██▍                                                                                                                        | 69/5000 [01:04<45:46,  1.80it/s]
```

## Testing Instructions

1. `just recreate`
1. `just init`
1. `docker-compose exec web bash`
1. `python manage.py generatewaveforms`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/models/audio.py
+++ b/api/catalog/api/models/audio.py
@@ -12,6 +12,7 @@ from catalog.api.models.mixins import FileMixin, ForeignIdentifierMixin, MediaMi
 from catalog.api.utils.waveform import generate_peaks
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
+from django.db import transaction
 from uuslug import uuslug
 
 
@@ -192,6 +193,7 @@ class Audio(AudioFileMixin, AbstractMedia):
         except AudioSet.DoesNotExist:
             return None
 
+<<<<<<< HEAD
     def get_waveform(self) -> list[float]:
         """
         Get the waveform if it exists. Return a blank list otherwise.
@@ -204,6 +206,9 @@ class Audio(AudioFileMixin, AbstractMedia):
         except AudioAddOn.DoesNotExist:
             return []
 
+=======
+    @transaction.atomic
+>>>>>>> d0f86a02 (Paginate generatewaveforms)
     def get_or_create_waveform(self):
         add_on, _ = AudioAddOn.objects.get_or_create(audio_identifier=self.identifier)
 
--- a/api/catalog/api/models/audio.py
+++ b/api/catalog/api/models/audio.py
@@ -11,7 +11,7 @@ from catalog.api.models.media import (
 from catalog.api.models.mixins import FileMixin, ForeignIdentifierMixin, MediaMixin
 from catalog.api.utils.waveform import generate_peaks
 from django.contrib.postgres.fields import ArrayField
-from django.db import models, transaction
+from django.db import models
 from uuslug import uuslug
 
 
--- a/api/catalog/api/models/audio.py
+++ b/api/catalog/api/models/audio.py
@@ -11,8 +11,7 @@ from catalog.api.models.media import (
 from catalog.api.models.mixins import FileMixin, ForeignIdentifierMixin, MediaMixin
 from catalog.api.utils.waveform import generate_peaks
 from django.contrib.postgres.fields import ArrayField
-from django.db import models
-from django.db import transaction
+from django.db import models, transaction
 from uuslug import uuslug
 
 
--- a/api/catalog/api/models/audio.py
+++ b/api/catalog/api/models/audio.py
@@ -192,7 +192,6 @@ class Audio(AudioFileMixin, AbstractMedia):
         except AudioSet.DoesNotExist:
             return None
 
-<<<<<<< HEAD
     def get_waveform(self) -> list[float]:
         """
         Get the waveform if it exists. Return a blank list otherwise.
@@ -205,9 +204,6 @@ class Audio(AudioFileMixin, AbstractMedia):
         except AudioAddOn.DoesNotExist:
             return []
 
-=======
-    @transaction.atomic
->>>>>>> d0f86a02 (Paginate generatewaveforms)
     def get_or_create_waveform(self):
         add_on, _ = AudioAddOn.objects.get_or_create(audio_identifier=self.identifier)
 
--- a/api/catalog/management/commands/generatewaveforms.py
+++ b/api/catalog/management/commands/generatewaveforms.py
@@ -3,6 +3,7 @@ import subprocess
 
 from catalog.api.models.audio import Audio, AudioAddOn
 from django_tqdm import BaseCommand
+from limit import limit
 
 
 def paginate_reducing_query(get_query_set, page_size=10):
@@ -33,6 +34,54 @@ class Command(BaseCommand):
     That should be enough latency to not cause any problems.
     """
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--unlimited", help="Remove self impose rate limits for testing."
+        )
+        parser.add_argument(
+            "--limit", help="Limit the number of waveforms to create.", type=int
+        )
+
+    def get_audio_handler(self, options):
+        if options["unlimited"]:
+            return lambda audio: audio.get_or_create_waveform()
+
+        @limit(limit=1, every=2)  # Call once per two seconds maximum
+        def limited(audio):
+            audio.get_or_create_waveform()
+
+        return limited
+
+    def _process_wavelengths(self, audios, audio_handler, count_to_process):
+        errored_identifiers = []
+        processed = 0
+        with self.tqdm(total=count_to_process) as progress:
+            paginator = paginate_reducing_query(
+                get_query_set=lambda: audios.exclude(identifier__in=errored_identifiers)
+            )
+            for page in paginator:
+                for audio in page:
+                    if processed > count_to_process:
+                        return errored_identifiers
+                    try:
+                        processed += 1
+                        audio_handler(audio)
+                    except subprocess.CalledProcessError as err:
+                        errored_identifiers.append(audio.identifier)
+                        self.error(
+                            f"Unable to process {audio.identifier}: "
+                            f"{err.stderr.decode().strip()}"
+                        )
+                    except KeyboardInterrupt:
+                        errored_identifiers.append(audio.identifier)
+                        return errored_identifiers
+                    except BaseException as err:
+                        errored_identifiers.append(audio.identifier)
+                        self.error(f"Unable to process {audio.identifier}: " f"{err}")
+                    progress.update(1)
+
+        return errored_identifiers
+
     def handle(self, *args, **options):
         # These logs really muck up the tqdm output and don't give us much helpful
         # information, so they get silenced
@@ -44,35 +93,35 @@ class Command(BaseCommand):
         audios = Audio.objects.exclude(
             identifier__in=existing_waveform_audio_identifiers_query
         ).order_by("id")
-        count = audios.count()
-        self.stdout.write(
-            self.style.NOTICE(f"Generating waveforms for {count:,} records")
-        )
-        errored_identifiers = []
-        with self.tqdm(total=count) as progress:
-            paginator = paginate_reducing_query(
-                get_query_set=lambda: audios.exclude(identifier__in=errored_identifiers)
-            )
-            for page in paginator:
-                for audio in page:
-                    try:
-                        audio.get_or_create_waveform()
-                    except subprocess.CalledProcessError as err:
-                        errored_identifiers.append(audio.identifier)
-                        self.stderr.write(
-                            self.style.ERROR(
-                                f"Unable to process {audio.identifier}: "
-                                f"{err.stderr.decode().strip()}"
-                            )
-                        )
-                    except BaseException as err:
-                        errored_identifiers.append(audio.identifier)
-                        self.stderr.write(
-                            self.style.ERROR(
-                                f"Unable to process {audio.identifier}: " f"{err}"
-                            )
-                        )
-                    progress.update(1)
 
-        self.stdout.write(self.style.SUCCESS("Finished generating waveforms!"))
-        self.stdout.write(self.style.WARNING(f"The following Audio identifiers were unable to be processed {','.join(str(identifier) for identifier in errored_identifiers)}"))
+        limit_records = options["limit"]
+        count = audios.count()
+
+        count_to_process = count
+
+        if limit_records is not None:
+            count_to_process = limit_records if limit_records < count else count
+
+        self.info(
+            self.style.NOTICE(f"Generating waveforms for {count_to_process:,} records")
+        )
+
+        audio_handler = self.get_audio_handler(options)
+
+        errored_identifiers = self._process_wavelengths(
+            audios, audio_handler, count_to_process
+        )
+
+        self.info(self.style.SUCCESS("Finished generating waveforms!"))
+
+        if errored_identifiers:
+            errored_identifiers_joined = "\n".join(
+                str(identifier) for identifier in errored_identifiers
+            )
+
+            self.info(
+                self.style.WARNING(
+                    f"The following Audio identifiers were unable "
+                    f"to be processed\n\n{errored_identifiers_joined}"
+                )
+            )
--- a/api/catalog/management/commands/generatewaveforms.py
+++ b/api/catalog/management/commands/generatewaveforms.py
@@ -1,4 +1,5 @@
 import logging
+import subprocess
 
 from catalog.api.models.audio import Audio
 from django_tqdm import BaseCommand
@@ -19,7 +20,15 @@ class Command(BaseCommand):
         )
         with self.tqdm(total=count) as progress:
             for audio in audios:
-                audio.get_or_create_waveform()
+                try:
+                    audio.get_or_create_waveform()
+                except subprocess.CalledProcessError as err:
+                    self.stderr.write(
+                        self.style.ERROR(
+                            f"Unable to process {audio.identifier}: "
+                            f"{err.stderr.decode().strip()}"
+                        )
+                    )
                 progress.update(1)
 
         self.stdout.write(self.style.SUCCESS("Finished generating waveforms!"))
--- a/api/catalog/management/commands/generatewaveforms.py
+++ b/api/catalog/management/commands/generatewaveforms.py
@@ -75,3 +75,4 @@ class Command(BaseCommand):
                     progress.update(1)
 
         self.stdout.write(self.style.SUCCESS("Finished generating waveforms!"))
+        self.stdout.write(self.style.WARNING(f"The following Audio identifiers were unable to be processed {','.join(str(identifier) for identifier in errored_identifiers)}"))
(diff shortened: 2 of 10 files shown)


dhruvkb (approved):

sarayourfriend (commented):
Tested it locally and it worked! I ran the command and then `just dj shell` and confirmed the processed waveforms were in the audio_add_on table.

I've got a couple questions/gut-checks I'd like to verify before approving. Unit tests would also be nice to have for this, especially if my suggestions end up getting implemented (in particular the start/stop/resume aspect) and the complexity of it increases.

api/catalog/management/commands/generatewaveforms.py
```diff
@@ -0,0 +1,34 @@
...
+
+class Command(BaseCommand):
+    help = "Generates waveforms for all audio records to populate the cache"
+
+    def handle(self, *args, **options):
+        # These logs really muck up the tqdm output and don't give us much helpful
+        # information, so they get silenced
+        logging.getLogger("catalog.api.utils.waveform").setLevel(logging.WARNING)
+
+        audios = Audio.objects.all().order_by("id")
```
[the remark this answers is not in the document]
AetherUnbound: Ah drat, I naively assumed that iterating over them in this manner would automatically paginate a cursor on the database. According to the docs[^1], `all()` evaluates the entire `QuerySet`. I think we can use a `Paginator` to improve that.

> Would it be worth excluding any rows that already have corresponding rows in the add_on table?

Absolutely, great call!
[^1]: https://docs.djangoproject.com/en/dev/ref/models/querysets/#when-querysets-are-evaluated

api/catalog/management/commands/generatewaveforms.py
```diff
@@ -0,0 +1,34 @@
...
+        count = audios.count()
+        self.stdout.write(
+            self.style.NOTICE(f"Generating waveforms for {count:,} records")
+        )
+        with self.tqdm(total=count) as progress:
+            for audio in audios:
+                try:
+                    audio.get_or_create_waveform()
```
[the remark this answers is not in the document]
AetherUnbound: I would be okay with adding a delay at this step :slightly_smiling_face: 0.25 seconds is a small fraction of the time it takes to actually retrieve + compute the waveform, so that seems fine.

AetherUnbound:
> Unit tests would also be nice to have for this, especially if my suggestions end up getting implemented (in particular the start/stop/resume aspect) and the complexity of it increases.

Agreed, I'll try to start on those if I have time before I'm out. Initially this was really straightforward, but then I added error handling, and with the other suggestions it definitely makes sense to do!

api/catalog/management/commands/generatewaveforms.py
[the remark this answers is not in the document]
sarayourfriend: I was noticing that as I ran it locally. I _almost_ suggested doing something like a speed _limit_ (1 row per second but include the time spent processing the previous row in the "wait" time between each row... or something?) but decided that was probably too complicated/overkill (though if there's some kind of utility for already doing that easily it would be sweet to use it).

sarayourfriend:
@AetherUnbound If you're okay with handing this off and run out of time, I'm happy to pick this up wherever you leave off, no sweat, just let me know.

AetherUnbound:
@sarayourfriend I don't think I'm going to end up getting to this before I'm out - please feel free to take it over!

sarayourfriend:
@AetherUnbound Sure thing!

sarayourfriend:
@obulat @dhruvkb this should be ready for review now after adding pagination and unit tests

dhruvkb (approved):
This looks good to me. A few lint fixes pointed out by the check and we could merge this!

zackkrida:
@sarayourfriend Seeing three remaining lint errors in the [latest run](https://github.com/WordPress/openverse-api/runs/5394971440?check_suite_focus=true):

```shell
api/test/unit/management/commands/generatewaveforms_test.py:53:5: F841 local variable 'waveforms' is assigned to but never used
api/test/factory/models/media.py:6:1: F401 'catalog.api.models.media.AbstractMedia' imported but unused
api/test/factory/models/media.py:21:58: E741 ambiguous variable name 'l'
```

sarayourfriend:
Thanks @zackkrida, fixed and pushed. Tho now there are some conflicts I need to resolve :+1:

api/catalog/api/models/audio.py
```diff
@@ -192,6 +192,7 @@ def audio_set(self):
         except AudioSet.DoesNotExist:
             return None
 
+<<<<<<< HEAD
```
obulat, line 195: Left-overs from merge. They cause linting failure in CI, too.

obulat (commented):
I first tried it out with the `get_waveform` leftover from the merge, and then with the `get_waveform removed. I guess the latter is the correct way.

It works, but it seems like it's stopping every once in a while. I think that it could be swallowing some errors. Below is an error I got when I was running it at first. The process stalled, and I pressed CTRL+C, and this was the error I saw in the console:
```
Unable to process d495818e-b569-45a6-8e16-64f37f672225: HTTPSConnectionPool(host='mp3d.jamendo.com', port=443): Max retries exceeded with url: /download/track/1204669/mp32 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0xffff948bf460>: Failed to establish a new connection: [Errno 111] Connection refused'))
  0%|                                       | 6/5000 [01:19<25:25:55, 18.33s/it]^CKeyboardInterrupt
2022-03-03T07:54:38Z
Unable to process 2470ec1f-0c29-4063-98d9-8c4da09c3572:
```
The process went on after CTRL+C, only to stop again later.

This error seems to be different from `  1%|▋                                                                                 | 39/5000 [00:55<49:37,  1.67it/s]Unable to process 67acc776-838c-4242-bb99-be9e0be061b6: Can't generate json format output from Unknown file format`, too.

1. Why is the waveform generation not resuming after an error?
2. I think we should record the files that cause the error to check them and possibly remove from the catalog if they are invalid.

api/catalog/management/commands/generatewaveforms.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class Command(BaseCommand):
+    help = "Generates waveforms for all audio records to populate the cache."
+    """
+    Note: We rely on the file download and waveform generation times
+    taking long enough to prevent us from either making too many requests
+    to the upstream provider or inserting into our database too quickly and
+    causing a slow down. In local tests and in tests run on the staging server
+    it appeared to take on average around 6 to 8 seconds for each audio file.
```
obulat, line 32: I think it should depend on the size of the audio file. Many Freesound and Wikimedia files are very small. I sometimes see 2-3 it/s in the console, and the maximum is 11 it/s.

sarayourfriend: Right. I think we probably should use the `limit` library anyway just in case.

api/catalog/management/commands/generatewaveforms.py
```diff
@@ -0,0 +1,127 @@
...
+    That should be enough latency to not cause any problems.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--unlimited", help="Remove self impose rate limits for testing."
+        )
+        parser.add_argument(
+            "--limit", help="Limit the number of waveforms to create.", type=int
```
AetherUnbound, line 42: Just a thought, the fact that these both have `limit` in them is kind of confusing. Maybe we could do:
* `--unlimited` -> `--no-rate-limit`
* `--limit` -> `--max-records`

What do you think?

api/catalog/management/commands/generatewaveforms.py
```diff
@@ -0,0 +1,127 @@
...
+        # These logs really muck up the tqdm output and don't give us much helpful
+        # information, so they get silenced
+        logging.getLogger("catalog.api.utils.waveform").setLevel(logging.WARNING)
+
+        existing_waveform_audio_identifiers_query = AudioAddOn.objects.filter(
+            waveform_peaks__isnull=False
+        ).values_list("audio_identifier", flat=True)
+        audios = Audio.objects.exclude(
+            identifier__in=existing_waveform_audio_identifiers_query
+        ).order_by("id")
```
AetherUnbound, line 95: Are we taking this approach because `Audio` and `AudioAddOn` don't have a foreign key relationship?

api/test/factory/models/audio.py
```diff
@@ -0,0 +1,19 @@
+from test.factory.faker import Faker
+from test.factory.models.media import IdentifierFactory, MediaFactory
+
+from catalog.api.models.audio import Audio, AudioAddOn
+from factory.django import DjangoModelFactory
+
+
+class AudioFactory(MediaFactory):
```
AetherUnbound, line 8: Super neat!

api/test/run_test.sh
```diff
@@ -18,8 +18,6 @@ succeeded=$?
 if [[ $succeeded -eq 0 ]]; then
 	printf "${green}:-) All tests passed${endcol}\n"
 else
-  printf "Full system logs:\n"
-  docker-compose logs
```
AetherUnbound, line 22: I know we have another ticket for this - should we hold off on this change for now?

api/test/unit/management/commands/generatewaveforms_test.py
```diff
@@ -0,0 +1,177 @@
...
+
+@pytest.mark.django_db
+def test_creates_waveforms_for_audio():
+    AudioFactory.create_batch(153)
+
+    assert AudioAddOn.objects.count() == 0
+
+    call_generatewaveforms()
+
+    assert_all_audio_have_waveforms()
```
AetherUnbound, line 44: I love this approach!

api/test/unit/management/commands/generatewaveforms_test.py
```diff
@@ -0,0 +1,177 @@
...
+        test_audio.get_or_create_waveform()
+    test_audio.delete()
+
+    queries_per_iteration = len(capture.captured_queries)
+
+    pagination_queries = pages + 1  # 1 per page + the final empty page's query
+    count_queries = 1  # initializes the count for tqdm
+    interation_queries = (
+        queries_per_iteration * audio_count
+    )  # queries inside get_or_create_waveform
```
AetherUnbound, line 89: I know the format isn't great as a result of this change, but if these comments are on the line above, `black` won't try and split L87-L89 over multiple lines

AetherUnbound (commented):
I can't approve my own PR :sweat_smile: but here's a few comments, nothing to stop a merge though!

api/catalog/management/commands/generatewaveforms.py
sarayourfriend: Yes, I like these better. I hated the ones I chose!

api/catalog/management/commands/generatewaveforms.py
sarayourfriend: By "this approach" what do you mean? The inner select on the add ons table?

The tables don't have FK relationship though, you're right, at least not one the DB knows about.

api/test/run_test.sh
sarayourfriend: Oh sure. I think I did this in rage at not being able to debug the tests, didn't mean to commit it though.

api/test/unit/management/commands/generatewaveforms_test.py
sarayourfriend: Done in 8fb6223843372f6714aaea31dc1ea0545751c9f7

api/catalog/management/commands/generatewaveforms.py
sarayourfriend: Done in e75964f68a12029b70c8c950945dcc0a9f907a95

api/test/run_test.sh
sarayourfriend: Put back in 753182c436281804b2e8508210da8c01f1ac32fc

api/catalog/management/commands/generatewaveforms.py
AetherUnbound: Oh sorry, I meant that I could have sworn we could do something like
```
Audio.objects.filter(audioaddon__waveform_peaks__isnull=True)
```
if these were foreign keys. They're not, so I don't think Django would be able to interpret that relationship appropriately.

_And sorry for the lack of context, I had that initially but second guessed myself and deleted the entire comment :woman_facepalming: _

zackkrida:
@AetherUnbound I think your symbolic approval here would be fine, once you think it's good to go I can merge it, if no one else reviews it by then.

stacimc:
This looks great! Manually testing with interrupts, with the rate limit enabled/disabled, and setting a limit to let the batch complete all seem to be working great 🎉 

Other than merge conflicts, I think the [`test_audio_success_examples`](https://github.com/WordPress/openverse-api/blob/main/api/test/examples_test.py#L25) test may also need to be updated. When I run the api-tests locally I get something like:
```
Differing items:
E         {'peaks': [0.61275, 0.19593, 0.74023, 0.45534, 0.52315, 0.40059, ...]} != {'peaks': []}
```
For the audio detail & search endpoints.

AetherUnbound:
I'm also getting the error Staci is getting, as well as:

```
in_val = '\n# Search for music titled "Wish You Were Here" by The.madpix.project\ncurl  "http://localhost:8000/v1/audio/?title=Wish%20You%20Were%20Here&creator=The.madpix.project"\n'
out_val = {'application/json': {'page': 1, 'page_count': 0, 'page_size': 20, 'result_count': 1, ...}}

    @pytest.mark.parametrize("in_val, out_val", list(audio_mappings.items()))
    def test_audio_success_examples(in_val, out_val):
        res = execute_request(in_val)
>       assert res == out_val["application/json"]
E       AssertionError: assert {'page': 1, '...ount': 1, ...} == {'page': 1, '...ount': 1, ...}
E         Omitting 4 identical items, use -vv to show
E         Differing items:
E         {'results': [{'category': 'music', 'creator': 'The.madpix.project', 'creator_url': 'https://www.jamendo.com/artist/441585/the.madpix.project', 'detail_url': 'http://localhost:8000/v1/audio/8624ba61-57f1-4f98-8a85-ece206c319cf/', ...}]} != {'results': [{'category': 'music', 'creator': 'The.madpix.project', 'creator_url': 'https://www.jamendo.com/artist/441585/the.madpix.project', 'detail_url': 'http://localhost:8000/v1/audio/8624ba61-57f1-4f98-8a85-ece206c319cf/', ...}]}
E         Use -v to get the full diff
```

api/catalog/management/commands/generatewaveforms.py
sarayourfriend: Ohh yes no that's not possible unfortunately because without the explicit relationship field Django won't add the reverse field relationships you'd need to make that JOIN.

sarayourfriend:
@stacimc @AetherUnbound does that happen on a fresh version of the API or only with existing data after running `generatewaveforms`? 

If I do `just recreate && just init && just api-tests` they all pass. But if I run `just dj "generatewaveforms --max_records=3"` and then run `just api-tests` again they fail.

I think this is because our API integration tests rely on the existing database rather than (what would be ideal) building their own data that then gets torn down after each test iteration.

Is fixing that in scope for this issue? I could probably find a workaround but it would likely cheapen the tests :confused: The easiest thing I can think would be to just delete the `peaks` property from both sides of the comparison.

obulat (approved):
I tried generating the waveforms and running together with the frontend locally, and I got myself some instant waveforms! :)
The problem is that there is a mismatch between what the front end assumes and what the API returns when no waveforms are available. If I understand correctly, the API returns an empty list `[]` if the waveform is not available.
The frontend expects either an array with at least 1 item where all values are between 0 and 1, or a falsy value (then it generates random peaks). 
When the frontend gets an empty list for peaks, `[]`, it does not show any waveform,  the validator throws an error, and there are lots of `Nan` errors when it is trying to calculate the waveform peaks.
I am not sure whether it should be fixed in this PR, or a separate one, though, so I'm approving :)

AetherUnbound:
Ah, my bad! I didn't run `recreate` before testing. I'm doing so now, as soon as that finishes OK I'll fix the Pipfile.lock conflicts and merge :smile:

sarayourfriend:
@obulat that's a bug introduced by the previous PR that actually added the peaks data to the API. Let's fix it in a separate issue.

stacimc (approved):
> I think this is because our API integration tests rely on the existing database rather than (what would be ideal) building their own data that then gets torn down after each test iteration.

My bad, I didn't look closely enough at the tests. They _are_ passing after recreating/reinitializing everything. That's definitely not ideal but absolutely a separate issue and shouldn't be fixed here, agreed!

PR #532, opened 2022-02-24
Add logging levels to Slack notifications in ingestion server

### LINKED ISSUE #481: Add "limited reporting" mode for ingestion server
## Problem

When running a data refresh in dev, we really don't need to have the same level of granularity for slack alerts that we do in production. We mostly just need "starting" and "done".

## Description

It would be great to have a simple way of distinguishing that some messages should only be sent in prod.

## Implementation

stacimc:
## Fixes

Fixes #481 by @AetherUnbound 

## Description

Adds Slack logging levels to the ingestion server; this lets us easily control the granularity of logging we need in a given environment. 
* Introduces `slack.error`, `slack.info`, and `slack.verbose` functions
    * `verbose` to be used for 'less important' messages; this allows an environment set at `INFO` to receive the most important info messages (data refresh start/stop) without triggering too many notifications.
        * I didn't use `trace`  for this because my understanding is that implies sensitive data and should not be used in production
    * We could easily add `warning`, `debug` etc if those become useful
* Updates existing slack notifications to use the new levels

The level is set using the `SLACK_LOG_LEVEL` environment variable. 

Notes/open questions:
* If the `SLACK_LOG_LEVEL` is not set, I've chosen to log at the `verbose` level by default. My reasoning was that's currently what we want in production, but I could certainly see a case for the exact opposite. What do folks think?

Because of this we should be able to deploy this without losing any logging in production (although we should still explicitly set the environment variable.) We can set staging to `INFO`.

## Testing Instructions

Run the tests:
`just ing-testlocal`

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/slack.py
+++ b/ingestion_server/ingestion_server/slack.py
@@ -3,21 +3,40 @@ import os
 
 import requests
 from decouple import config
+from enum import Enum
 
 
 log = logging.getLogger(__name__)
 SLACK_WEBHOOK = "SLACK_WEBHOOK"
+LOG_LEVEL = "SLACK_LOG_LEVEL"
+
+class Level(Enum):
+    VERBOSE = 0
+    INFO = 1
+    ERROR = 2
 
 
-def message(text: str, summary: str = None) -> None:
+def _message(text: str, summary: str = None, level: Level = Level.INFO) -> None:
     """
     Send a Slack message to a channel specified by a Slack webhook variable.
 
-    A message is only sent if the SLACK_WEBHOOK environment variable is undefined.
+    A message is only sent if the SLACK_WEBHOOK environment variable is undefined,
+    and the environment is configured to log at this level.
     """
+    environment = config("ENVIRONMENT", default="local")
+
     if not (webhook := os.getenv(SLACK_WEBHOOK)):
         log.debug(
             f"{SLACK_WEBHOOK} variable not defined, skipping slack message: {text}"
+        )
+        return
+    # If no log level is configured in the environment, log everything by default.
+    os_level = os.getenv(LOG_LEVEL)
+    os_level = Level.VERBOSE if os_level is None else Level[os_level]
+    if not (os_level.value <= level.value):
+        log.debug(
+            f"Slack logging level for {environment} set to {os_level.name}, skipping \
+            slack message with priority {level.name}: {text}"
         )
         return
     if not summary:
@@ -26,7 +45,6 @@ def message(text: str, summary: str = None) -> None:
         else:
             summary = text
 
-    environment = config("ENVIRONMENT", default="local")
 
     data = {
         "blocks": [{"text": {"text": text, "type": "mrkdwn"}, "type": "section"}],
@@ -39,3 +57,15 @@ def message(text: str, summary: str = None) -> None:
     except Exception as err:
         log.exception(f"Unable to issue slack message: {err}")
         pass
+
+
+def verbose(text: str, summary: str = None) -> None:
+    _message(text, summary, level = Level.VERBOSE)
+
+
+def info(text: str, summary: str = None) -> None:
+    _message(text, summary, level = Level.INFO)
+
+
+def error(text: str, summary: str = None) -> None:
+    _message(text, summary, level = Level.ERROR)
--- a/ingestion_server/ingestion_server/slack.py
+++ b/ingestion_server/ingestion_server/slack.py
@@ -5,6 +5,7 @@ from enum import Enum
 import requests
 from decouple import config
 
+
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


ingestion_server/ingestion_server/slack.py
```diff
@@ -3,30 +3,48 @@
...
+    environment = config("ENVIRONMENT", default="local")
+
     if not (webhook := os.getenv(SLACK_WEBHOOK)):
         log.debug(
             f"{SLACK_WEBHOOK} variable not defined, skipping slack message: {text}"
         )
         return
+    # If no log level is configured in the environment, log everything by default.
+    os_level = os.getenv(LOG_LEVEL)
+    os_level = Level.VERBOSE if os_level is None else Level[os_level]
```
AetherUnbound, line 35: Would it be easier to use a default here? (Could still be split across two lines if that's clearer)
```suggestion
    os_level = Level[os.getenv(LOG_LEVEL, Level.VERBOSE.name)]
```

ingestion_server/ingestion_server/slack.py
```diff
@@ -3,30 +3,48 @@
...
+    if not (os_level.value <= level.value):
```
AetherUnbound, line 36: Would it be OK to invert this logic?
```suggestion
    if level.value < os_level.value:
```

ingestion_server/test/unit_tests/test_slack.py
```diff
@@ -57,3 +57,62 @@ def test_message(
...
+    ]
+)
+def test_error(log_level, should_log, monkeypatch):
+    monkeypatch.setenv("ENVIRONMENT", "staging")
+    monkeypatch.setenv(slack.SLACK_WEBHOOK, "http://fake")
+    if log_level:
+        monkeypatch.setenv(slack.LOG_LEVEL, log_level)
+    with mock.patch("requests.post") as mock_post:
+        slack.error("text", "summary")
+        assert mock_post.called == should_log
```
AetherUnbound, line 118: FWIW these tests could be collapsed into a single parametrization:
```suggestion
@pytest.mark.parametrize(
    "log_func, log_level, should_log",
    [
        # Verbose logging
        (slack.verbose, "ERROR", False),
        (slack.verbose, "INFO", False),
        (slack.verbose, "VERBOSE", True),
        # Info logging
        (slack.info, "ERROR", False),
        (slack.info, "INFO", True),
        (slack.info, "VERBOSE", True),
        # Error logging
        (slack.error, "ERROR", True),
        (slack.error, "INFO", True),
        (slack.error, "VERBOSE", True),
        # If no log level set, verbose logging is enabled by default
        (slack.verbose, None, True)
        (slack.info, None, True)
        (slack.error, None, True)
    ]
)
def test_verbose(log_func, log_level, should_log, monkeypatch):
    monkeypatch.setenv("ENVIRONMENT", "staging")
    monkeypatch.setenv(slack.SLACK_WEBHOOK, "http://fake")
    if log_level:
        monkeypatch.setenv(slack.LOG_LEVEL, log_level)
    with mock.patch("requests.post") as mock_post:
        log_func("text", "summary")
        assert mock_post.called == should_log
```

AetherUnbound (commented):
This is great! I love the use of enums; I was going to have it be a boolean flag but I think this approach is much more appropriate & versatile. I have a few thoughts on logic/parametrization.

ingestion_server/ingestion_server/slack.py
stacimc: I think so -- I can't remember why I changed it around, but it certainly looks nicer :)

ingestion_server/ingestion_server/slack.py
stacimc: Totally, much easier to read 💯

ingestion_server/test/unit_tests/test_slack.py
stacimc: That's true -- I was resisting collapsing the tests for multiple functions. But they are extremely related and it certainly is a lot of duplication!

stacimc:
I think this is ready for review, but I'd like to make sure to test this end to end locally or on staging.

sarayourfriend (approved):
LGTM! Ran the tests and they pass; also tried to break it by passing a bad log level. This appears to have potentially broken the `just init` script and I can't figure out how to actually run ingestion locally otherwise :sweat_smile:

ingestion_server/test/unit_tests/test_slack.py
AetherUnbound: I hear you - I think if there were anything different within the tests besides the actual function call, it would make sense to have them separate!

AetherUnbound (approved):
Fantastic!

Issue #551, opened 2022-03-04
Random image endpoint


zackkrida:
## Problem

Many of the images in Openverse, particularly from Flickr, have never actually been viewed by someone before.

## Description

It could be compelling to help people find images that have never been viewed before. A `/random` endpoint could be a cool way to do this. We could use the [`random_scoring`](https://www.elastic.co/guide/en/elasticsearch/guide/current/random-scoring.html#random-scoring) feature in ElasticSearch to facilitate this.

sarayourfriend:
Are we psychically connected? I was thinking about this last night!

It'd be nice to add some parameters to it to confine it optionally. For example, it'd be cool to be able to do `/images/random?providers=...` or even better `/images/random?provider_group=glam` for example. Maybe both!

Whatever we use could also be used to create a `random/daily` endpoint as well that caches the response for 24 hours. This could power a "Picture of the day" kind of feature.

dhruvkb:
It might be possible to use the same set of filters as search for this random image endpoint. Filters like resolution (and license) would be very useful for applications like rotating wallpapers.

sarayourfriend:
In https://github.com/WordPress/openverse-api/pull/554#issuecomment-1065142995 @zackkrida noted that `license` and `license_type` parameters can conflict in some sense. If you use the commerical license type filter but then add a non-commercial license filter, what should it do?

My gut tells me that the `license_type` is essentially a mask of licenses and it should just do something like this:

```py
licenses_to_search = get_licenses_for_group(license_type) + licenses_from_param
```

And call it a day. This eliminates the potential for conflict, needing to resolve anything, and allows user refined searches that still benefit from the concept of filter groups.

For example, I believe the following should be a valid search (and I can easily imagine a use case for it):

```
/v1/images/?provider_group=glam&provider=non-glam-provider
```

You could want the GLAM designated providers and also to include a specific provider that is not in that group for whatever reason.

dhruvkb:
Did some research (and built a PoC demo locally to get this working). The key is to use a [`function_score` query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html) which wraps the original search query and then multiplies the document scores with the document's own scores, effectively shuffling them.

```python
s.query = Q("function_score", query=s.query, random_score={})
```

zackkrida:
[...]

PR #554, opened 2022-03-08
Add parameter to exclude certain sources

### LINKED ISSUE #553: Query param to exclude a source
## Problem

As the number of sources in Openverse grows, it may prove beneficial to API users and the frontend to be able to _exclude_ sources, as opposed to limiting results by a list of included sources.

A primary use case I see for this is to omit Flickr, if one doesn't want to see content that can vary greatly in quality, as opposed to the other, more consistent platforms.

dhruvkb:
## Fixes

Fixes #553 by @zackkrida 

## Description

This PR adds an `excluded_source` param that works opposite to `source` in that it drops results from the given providers.

## Testing Instructions

0. Check out the PR and run the server.
1. Send the `excluded_source` param with the search request.
2. Observe that no media from the excluded sources are listed.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -167,6 +167,23 @@ def _apply_filter(s: Search, search_params, param_name, renamed_param=None):
         return s
 
 
+def _apply_exclude(s: Search, search_params, param_name, renamed_param=None):
+    """
+    Same as ``_apply_filter`` above, but works in a manner opposite to that of
+    ``filter`` by using ``exclude``.
+    """
+
+    if param_name in search_params.data:
+        filters = []
+        for arg in search_params.data[param_name].split(","):
+            _param = renamed_param if renamed_param else param_name
+            args = {"name_or_query": "term", _param: arg}
+            filters.append(Q(**args))
+        return s.exclude("bool", should=filters)
+    else:
+        return s
+
+
 def _exclude_filtered(s: Search):
     """
     Hide data sources from the catalog dynamically.
@@ -227,8 +244,11 @@ def search(
         ("license_type", "license__keyword"),
     ]
     for tup in filters:
-        api_field, elasticsearch_field = tup
-        s = _apply_filter(s, search_params, api_field, elasticsearch_field)
+        s = _apply_filter(s, search_params, *tup)
+
+    exclude = [("excluded_source", "source")]
+    for tup in exclude:
+        s = _apply_exclude(s, search_params, *tup)
 
     # Exclude mature content and disabled sources
     s = _exclude_mature_by_param(s, search_params)
--- a/api/catalog/api/serializers/audio_serializers.py
+++ b/api/catalog/api/serializers/audio_serializers.py
@@ -43,6 +43,7 @@ class AudioSearchRequestSerializer(MediaSearchRequestSerializer):
     fields_names = [
         *MediaSearchRequestSerializer.fields_names,
         "source",
+        "excluded_source",
         "categories",
         "duration",
     ]
@@ -54,6 +55,13 @@ class AudioSearchRequestSerializer(MediaSearchRequestSerializer):
     source = serializers.CharField(
         label="provider",
         help_text="A comma separated list of data sources to search. Valid "
+        "inputs: "
+        f"`{list(get_sources('audio').keys())}`",
+        required=False,
+    )
+    excluded_source = serializers.CharField(
+        label="excluded_provider",
+        help_text="A comma separated list of data sources to ignore. Valid "
         "inputs: "
         f"`{list(get_sources('audio').keys())}`",
         required=False,
@@ -79,6 +87,9 @@ class AudioSearchRequestSerializer(MediaSearchRequestSerializer):
         input_sources = [x for x in input_sources if x in allowed_sources]
         input_sources = ",".join(input_sources)
         return input_sources.lower()
+
+    def validate_excluded_source(self, input_sources):
+        return self.validate_source(input_sources)
 
     @staticmethod
     def validate_categories(value):
--- a/api/catalog/api/serializers/image_serializers.py
+++ b/api/catalog/api/serializers/image_serializers.py
@@ -17,6 +17,7 @@ class ImageSearchRequestSerializer(MediaSearchRequestSerializer):
     fields_names = [
         *MediaSearchRequestSerializer.fields_names,
         "source",
+        "excluded_source",
         "categories",
         "aspect_ratio",
         "size",
@@ -29,6 +30,13 @@ class ImageSearchRequestSerializer(MediaSearchRequestSerializer):
     source = serializers.CharField(
         label="provider",
         help_text="A comma separated list of data sources to search. Valid "
+        "inputs: "
+        f"`{list(get_sources('image').keys())}`",
+        required=False,
+    )
+    excluded_source = serializers.CharField(
+        label="excluded_provider",
+        help_text="A comma separated list of data sources to ignore. Valid "
         "inputs: "
         f"`{list(get_sources('image').keys())}`",
         required=False,
@@ -61,6 +69,9 @@ class ImageSearchRequestSerializer(MediaSearchRequestSerializer):
         input_sources = [x for x in input_sources if x in allowed_sources]
         input_sources = ",".join(input_sources)
         return input_sources.lower()
+
+    def validate_excluded_source(self, input_sources):
+        return self.validate_source(input_sources)
 
     @staticmethod
     def validate_categories(value):


api/catalog/api/controllers/search_controller.py
```diff
@@ -167,6 +167,23 @@ def _apply_filter(s: Search, search_params, param_name, renamed_param=None):
         return s
 
 
+def _apply_exclude(s: Search, search_params, param_name, renamed_param=None):
```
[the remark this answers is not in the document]
sarayourfriend: If you take Madison's suggestion it'd be nice to rename some of the parameters as the current names are a little confusing:

* `param_name` -> `django_field_name` or `query_param_name`, not sure which is more accurate
* `renamed_param` -> `elastic_field_name`

api/catalog/api/controllers/search_controller.py
```diff
@@ -227,8 +244,11 @@ def search(
         ("license_type", "license__keyword"),
     ]
     for tup in filters:
```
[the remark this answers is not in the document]
sarayourfriend: Notably I think the error we'd get if one of the `filters` was misconfigured would be clearer as a tuple unpacking error than a "too many arguments" error being passed to `_apply_filter`

sarayourfriend (commented):
What is the expected behavior when the params clash? What should this search do:

`/v1/images/?source=flickr&excluded_source=flickr`

To prevent the possibility of that we could collapse the two into just the `source` parameter and use a prefix like `source=not:flickr,wikimedia`. It's at least clearer that you shouldn't have `not:flickr,flickr` in that list.

My feeling is that it should just return a bad request with some kind of "conflicting params" error or something. What do you think?

dhruvkb:
The way I see it, excluded takes precedence over included. So `source=flickr&excluded_source=flickr` would give no results at all. Your idea for raising 400 also sounds valid to me. We could definitely do that.

The `search_controller` module could use a cleanup/refactor. Even after working with it quite a bit, I find myself often confused in a lot of places.

sarayourfriend:
> So source=flickr&excluded_source=flickr would give no results at all.

If that's the case it'd be good to add a test documenting that expectation.

I think ideally we shouldn't allow `source` and `excluded_source` to be used together at all and return a 400. If used together it's either a programming error or a misunderstanding of how they work.

Even if their values don't overlap, the following request makes no sense either:

`source=wikimedia&excluded_source=flickr`

`excluded_source` is useless here and might as well be ignored. Or else it takes presence and everything _except_ flickr is returned.

sarayourfriend:
@dhruvkb the `not:` prefix should probably go on the param rather than the values, right? like `not:source=flickr` instead of `source=not:flickr`...  mostly to prevent being able to mix them? Though, I wonder if we have any filter categories where some of the values are super-values of other ones... I can't think of any though :thinking:

api/catalog/api/controllers/search_controller.py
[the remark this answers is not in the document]
dhruvkb: Renamed to `serializer_field` (as it's being picked from the `search_params` serializer) and `es_field` (as it's being picked from Elasticsearch).

obulat:
I think I would prefer `not` prefix on the param, not on the values. `not:source=flickr&source=wikimedia` seems clearer than `source=not:flickr&source:wikimedia`.
And returning 400 for clashing responses seems best.

api/test/media_integration.py
```diff
@@ -14,6 +14,14 @@ def search(fixture):
     assert fixture["result_count"] > 0
 
 
+def search_all_excluded(media_path, excluded_source):
+    response = requests.get(
+        f"{API_URL}/v1/{media_path}?q=test&excluded_source={','.join(excluded_source)}"
+    )
+    data = json.loads(response.text)
+    assert data["result_count"] == 0
```
sarayourfriend, line 22: Can you add a test for the validation error as well? Either as an integration test or as a unit test directly for the serializer?

api/catalog/api/serializers/media_serializers.py
```diff
@@ -73,6 +86,61 @@ class TagSerializer(serializers.Serializer):
     )
 
 
+def get_search_request_source_serializer(media_type):
```
sarayourfriend, line 89: I feel like I don't usually see class factories in Python for some reason. Why did you choose this over inheritance? (not suggestion a change, just curious what the thought process is as I almost never reach for class factory in Python but I remember doing it a lot in Scala).

sarayourfriend (commented):
Looking great! Just one more test to add and I think this will be g2g from my perspective :tada:

api/catalog/api/serializers/media_serializers.py
dhruvkb: It was inheritance-based before but that meant just writing the same thing twice because the fields need to be redefined to change the help text.

api/catalog/api/controllers/search_controller.py
```diff
@@ -16,6 +18,11 @@
 from elasticsearch_dsl import Q, Search, connections
 from elasticsearch_dsl.query import Query
 from elasticsearch_dsl.response import Hit, Response
+from rest_framework.request import Request
+
+
+if TYPE_CHECKING:
+    from catalog.api.serializers.media_serializers import MediaSearchRequestSerializer
```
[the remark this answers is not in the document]
dhruvkb: I just recently learned this for `pls`. Side projects are awesome!

api/test/media_integration.py
dhruvkb: Added the test case in 8a97e14.

api/catalog/api/serializers/media_serializers.py
```diff
@@ -73,6 +86,61 @@ class TagSerializer(serializers.Serializer):
...
+        }
+
+        source = serializers.CharField(label="provider", **_field_attrs)
+        excluded_source = serializers.CharField(
+            label="excluded_provider", **_field_attrs
+        )
+
+        @staticmethod
+        def validate_source_field(input_sources):
+            allowed_sources = list(get_sources("image").keys())
```
sarayourfriend, line 120: It seems like `get_sources(media_type)` call could be moved outside of the class definition instead of having to call it multiple times.

```suggestion
            allowed_sources = list(get_sources(media_type).keys())
```

api/catalog/api/serializers/media_serializers.py
sarayourfriend: Oh, I totally glossed over the fact that the help text was changing based on the media type as well :facepalm: Makes a lot of sense!

api/test/media_integration.py
```diff
@@ -14,6 +14,21 @@ def search(fixture):
...
+
+
+def search_source_and_excluded(media_path):
+    response = requests.get(
+        f"{API_URL}/v1/{media_path}?q=test&source=x&excluded_source=x"
```
sarayourfriend, line 27: nit: It might make sense to change this to have different values for the two parameters, to emphasize the fact that the expected error is with _any_ mixing of the two parameters, not just the obvious contradiction of including and excluding a provider.

```suggestion
        f"{API_URL}/v1/{media_path}?q=test&source=x&excluded_source=y"
```

api/catalog/api/serializers/media_serializers.py
```diff
@@ -73,6 +86,61 @@ class TagSerializer(serializers.Serializer):
...
+        def validate(self, data):
+            data = super().validate(data)
+            if "source" in self.initial_data and "excluded_source" in self.initial_data:
+                raise serializers.ValidationError(
+                    "Cannot set both 'source' and 'excluded_source'. "
+                    "Use exactly one of these."
+                )
+            return data
+
+    return MediaSearchRequestSourceSerializer
```
sarayourfriend, line 141: It just occurred to me that the class name for any serializer created with this factory function will be the same, no?

Should we use the `type` constructor to create a dynamically defined class with an appropriate name? I'm recalling having to do that in the past for decorated class objects that needed their name changed to reflected the decorated status.

I'm not sure if this would effect stack traces or not.

sarayourfriend (approved):
LGTM!

I do think it might be interesting to consider implementing this in a generic way using the `not:` prefix. We'd implement it for all parameters at once that way instead of just for sources.

If we commit this then we'd need to start a new API version if we decided to take a more generic approach or else we'd the dangling `excluded_source` as well as the then canonical `not:source`.

I'll leave that up to others to consider. My preference is to take the generic route (based on the work already done in this PR, just extend it to the other fields) but I also don't want to block this if other's are on board with taking the current approach.

zackkrida:
Side note, the discussion of potentially conflicting search parameters here makes me wonder how exactly `license_type` and `license` interact with each other. They can potentially conflict—we protect against this in the front-end but I'm not sure about the API itself.


<table>
<tr>
	<td>license
	<td>string  non-empty  A comma-separated list of licenses. Example: by,cc0. Valid inputs: ['CC0', 'BY-NC-SA', 'BY', 'BY-ND', 'BY-NC', 'PDM', 'BY-SA', 'BY-NC-ND']
<tr>
	<td>license_type
	<td>string  non-empty  A list of license types. Valid inputs: ['all', 'all-cc', 'commercial', 'modification']
</table>

sarayourfriend:
In a sense the `license_type` parameter is an implementation of the filter "groups" we discussed in #551.

api/catalog/api/serializers/media_serializers.py
dhruvkb: While using `type()` will allow us to use dynamic class names, it'll make the code far less readable and much more difficult to understand. I strongly feel that this is not a valuable tradeoff.

obulat (approved):

sarayourfriend:
@dhruvkb before merging do you have any thoughts about whether to reimplement in a more generic way? Or at least in a way compatible with a generic approach in the future?

dhruvkb:
I don't have any concrete thoughts on that. Combining features using flat API params has many limitations.

sarayourfriend:
I guess the worst case if we do implement something with a broader approach in the future `excluded_source` could just have its implementation changed to be an alias of the generic parameter.

Our filtering logic is currently so simple I think we'd need to release a new API version if we decided to do anything sufficiently complicated that we'd run into the limitations of the basic generic approach of being able to inverse any given filter.

Anyway, feel free to merge :+1:

PR #564, opened 2022-03-11
Add a timeout to the `@_loop` recipe

### LINKED ISSUE #562: `just init` sometimes gets stuck when waiting for indices
## Description

Sometimes, quite randomly, `just init` gets stuck waiting for indices to be ready.

```text
curl -X POST -H 'Content-Type: application/json' -d '{"model": "image", "action": "LOAD_TEST_DATA"}' 'http://localhost:8001/task'
{"message": "Successfully scheduled task", "task_id": "ee930da1-5013-4133-b9e7-39c02a669ef3", "status_check": "http://localhost:8001/task/ee930da1-5013-4133-b9e7-39c02a669ef3"}
curl -X POST -H 'Content-Type: application/json' -d '{"model": "audio", "action": "LOAD_TEST_DATA"}' 'http://localhost:8001/task'
{"message": "Successfully scheduled task", "task_id": "e1fa71af-9503-4949-8ce3-039302e83188", "status_check": "http://localhost:8001/task/e1fa71af-9503-4949-8ce3-039302e83188"}
curl -X POST -H 'Content-Type: application/json' -d '{"model": "image", "action": "INGEST_UPSTREAM"}' 'http://localhost:8001/task'
{"message": "Successfully scheduled task", "task_id": "4f1b0c36-603e-4e21-93e6-adb1173e576c", "status_check": "http://localhost:8001/task/4f1b0c36-603e-4e21-93e6-adb1173e576c"}
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
Waiting for index 'image' to be ready...
````

This can also break CI sometimes.

## Reproduction

1. Clean up the Docker space.
    ```shell-session
    $ docker compose down -v
    ```
1. Bring up the services and init the database.
    ```shell-session
    $ just init
    ```
1. Repeat till you see it get stuck.

## Expectation
This should not get stuck, or should have timeout and retry functionality.

## Resolution

### PR DESCRIPTION
**Pull request by dhruvkb**
## Description

Prevents Just recipes for waiting to enter into endless loops. This is part of a larger effort to fix #562.

## Testing Instructions

0. Change the condition in any Just recipe that uses `@_loop` to fail.
1. Run the Just recipe.
2. It will timeout after 5m (or the set timeout duration).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### CODE REVIEW

File: justfile [hunk 1]
```diff
@@ -10,10 +10,10 @@ default:
 ###########
 
 # Sleep for given time showing the given message as long as given condition is met
-@_loop condition message time="5":
-    while [ {{ condition }} ]; do \
-        echo "{{ message }}" && sleep {{ time }}; \
-    done
+@_loop condition message timeout="5m" time="5":
```
sarayourfriend, line 13: 5 minutes is still a long time! Is that a reasonable amount of time to expect ES to take to boot up? I would have thought like 2 minutes max but maybe it could take a lot longer on smaller CPUs (or even in the action runners).

sarayourfriend (commented):
TIL about the `timeout` utility!

Will note that this is _not_ cross-platform.

We could write the script as a Python script instead if we wanted to ensure that: https://github.com/casey/just/#writing-recipes-in-other-languages

dhruvkb:
The `timeout` utility works on Linux and macOS. Considering that the `load_sample_data.sh` is a Bash script, I'm pretty sure we're not supporting Windows as a development environment right now.

sarayourfriend:
That's true, I wonder if we should add further things on the list of stuff to convert to support Windows developers?

Or else we should decide that we don't care about Windows devs (which I don't think is the right answer considering Window's marketshare) and make that explicit in the README and contributor docs.

zackkrida:
I would imagine many devs are using or have access to [WSL](https://docs.microsoft.com/en-us/windows/wsl/about) these days, perhaps we could add some simple docs to suggest it and consider that basic Windows support.

dhruvkb:
Supporting Windows is important. We could definitely put some WSL instructions in the docs but ensuring everything works.

sarayourfriend:
Would it be possible to print the last few lines of docker logs for the elastic container when the operation times out? Might help in debugging failures in CI without having to try reproducing the flakiness locally.

AetherUnbound:
It seems like `timeout` will exit with code 124 if the timeout is hit. Do you think we could use that to echo a message? For instance, if the command's error code is 124, `echo "Command failed after {{ timeout }}"` or something similar. Currently the output isn't super obvious:

```
$ just wait-for-ing
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
error: Recipe `_loop` failed on line 16 with exit code 124
error: Recipe `wait-for-ing` failed on line 134 with exit code 124
```

zackkrida:
@dhruvkb do you need reviews here, or to implement changes? Please feel free to ping the assigned reviewers if the former. Thanks

dhruvkb:
Needs some more work. Will ping reviewers as soon as ready.

sarayourfriend:
@dhruvkb just wanted to check if you needed any support on this PR?

dhruvkb:
@sarayourfriend just need to implement what @AetherUnbound recommended here: https://github.com/WordPress/openverse-api/pull/564#issuecomment-1067387898

AetherUnbound (changes requested):
Hmm, doesn't seem like the current implementation works on linux :( 
```
$ just wait-for-ing
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
Waiting for the ingestion-server to be healthy...
sh: 1: [[: not found
```

AetherUnbound:
Setting the following at the top of the `justfile` worked for me though! `set shell := ["bash", "-uc"]`

```
Waiting for the ingestion-server to be healthy...
Timed out
error: Recipe `_loop` failed on line 22 with exit code 124
error: Recipe `wait-for-ing` failed on line 145 with exit code 124
```

dhruvkb:
@AetherUnbound are you using a shell other than `bash` as the default? I'm guessing based on our error message that says:
```
sh: 1: [[: not found
```

AetherUnbound:
I'm not, `bash` is what I'm using now! :woman_shrugging: I take it this doesn't error out for you when it's run? If you add `echo $0` at the beginning of the loop recipe, what do you get? :open_mouth:

dhruvkb:
I'm getting `sh` too when I add the `echo $0`. But it turns out macOS was being sneaky.
Screenshot 2022-04-06 at 6 08 48 AM

I've updated the script to use `[` instead of `[[` in ed96453 so it should work consistently across shells now without needed to specify one at the top of the `justfile`.

AetherUnbound:
Whaaaaaa :open_mouth: how wild! Well this works for me now, thanks Dhruv!

Fun side note: I found this line while reading the man page for `sh` (apparently `sh` doesn't have a `-version` flag):
> ### Description
> Sh is the standard command interpreter for the system. The current version of sh is **in the process of being changed to conform with the POSIX 1003.2 and 1003.2a specifications for the shell**.  _(emphasis mine)_

I wanted to know how long `sh` has been "in the process of changing its specification", and it turns out that the [oldest man page I can find for `sh` from 1999](https://linux.die.net/man/1/ash) _also has this line_ :joy: ahhh, software.

dhruvkb:
Thanks, just need one more approval. @sarayourfriend or @krysal if you have a few moments, please give it a look.

zackkrida (approved):
lgtm!

PR #568, opened 2022-03-15
Add throttle exemptions

### LINKED ISSUE #565: Create an unrestricted rate limit model
## Problem

For Openverse integrations with large platforms or trusted partners, we need an unlimited rate limit model added.

## Description

> current rate limit groups
CleanShot 2022-03-11 at 16 07 18@2x

https://github.com/WordPress/openverse-api/blob/ab26c741d660707da9647d406e706759c70c5a82/api/catalog/settings.py#L128

Openverse uses the throttling functionality from [django rest framework](https://www.django-rest-framework.org/api-guide/throttling/). It seems we either need to create a class with extremely high limits, or add a way for throttled applications to be _exempted_ from throttling entirely.

sarayourfriend:
## Fixes

Fixes #565 by @zackkrida 

## Description

Adds the ability to easily define exemptions for throttle classes via a general DRF/Django OOP inspired approach.

The idea is that each exemption is defined as its own class which implements the `is_exempt` method which returns a boolean: true when exempt and false otherwise. If any exemption class declares the request as exempt from throttling, then the throttle's `allow_request` logic is short circuited and the request is allowed.

I've added comprehensive unit tests for this. I tried to cover every possible configuration and request situation with redundant checks against the pre-existing state before checking the requests.

The tests are roughly modeled after DRF's own tests for the base throttling classes: https://github.com/encode/django-rest-framework/blob/master/tests/test_throttling.py.

## Testing Instructions

As I mentioned, the unit tests should cover the code 100%.

We can also rely on the integration tests, they should all continue to pass without exception.

Smoke testing the API by running it locally and checking a few requests wouldn't hurt either. If you really feel like it, modifying the local running redis instance to add localhost IP or an API key to the client id allow list would also be good.

For example, on my computer, I exec'd into the redis container using:

```bash
docker-compose exec cache redis-cli
```

Then ran:

```
SADD 'ip-whitelist' '172.20.0.1'
```

And verified that I would not get rate limited.

Now run:

```
SREM 'ip-whitelist' '172.20.0.1'
```

and confirm that you _do_ get throttled.

I think you could do the same using the `AccessTokenFactory` in `just ipython` but there's some extra configuration you need to do to get the locally running app configured for oauth I guess.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Pipfile
+++ b/api/Pipfile
@@ -53,7 +53,6 @@ gunicorn = "*"
 sentry-sdk = "*"
 django-tqdm = "*"
 limit = "*"
-fakeredis = "*"
 
 [requires]
 python_version = "3.10"
--- a/api/Pipfile
+++ b/api/Pipfile
@@ -15,6 +15,7 @@ sphinx-autobuild = "*"
 furo = "*"
 myst-parser = "*"
 factory-boy = "*"
+fakeredis = "*"
 
 [packages]
 aws-requests-auth = "*"
@@ -52,6 +53,7 @@ gunicorn = "*"
 sentry-sdk = "*"
 django-tqdm = "*"
 limit = "*"
+fakeredis = "*"
 
 [requires]
 python_version = "3.10"
--- a/api/catalog/api/utils/throttle.py
+++ b/api/catalog/api/utils/throttle.py
@@ -8,14 +8,29 @@ from rest_framework.throttling import SimpleRateThrottle
 
 log = logging.getLogger(__name__)
 
-
 class ThrottleExemption:
+    """
+    Abstract class describing a given throttle exemption.
+
+    To be included in an iterable of ``ThrottleExemption``s
+    in children of ``ExemptionAwareThrottle``.
+    """
+
     def __init__(self, throttle_class, request):
+        """
+        :param throttle_class: The throttle class the exemption modifies.
+        :param request: The current request against which to evaluate the exemption.
+        """
         self.throttle_class = throttle_class
         self.request = request
 
     @abc.abstractmethod
     def is_exempt(self) -> bool:
+        """
+        Whether the current request is exempt from throttling.
+
+        :return: ``True`` if exempt, ``False`` otherwise.
+        """
         ...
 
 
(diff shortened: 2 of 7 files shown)


api/Pipfile
```diff
@@ -52,6 +53,7 @@ gunicorn = "*"
 sentry-sdk = "*"
 django-tqdm = "*"
 limit = "*"
+fakeredis = "*"
```
AetherUnbound, line 56: Does this need to be in both the dev dependencies and the regular ones?

api/catalog/api/utils/throttle.py
```diff
@@ -7,24 +8,98 @@
 
 log = logging.getLogger(__name__)
 
+class ThrottleExemption:
```
AetherUnbound, line 11: Should this also be an [abstract base class](https://docs.python.org/3/library/abc.html#abc.ABC)?
```suggestion
class ThrottleExemption(abc.ABC):
```

AetherUnbound (approved):
This is a really great approach! I have a few points, but otherwise looks great :+1:

api/Pipfile
sarayourfriend: It should just be in the dev ones, whoops!

api/test/run_test.sh
```diff
@@ -19,7 +19,7 @@ if [[ $succeeded -eq 0 ]]; then
 	printf "${green}:-) All tests passed${endcol}\n"
 else
     printf "Full system logs:\n"
-    docker-compose logs
+    # docker-compose logs
```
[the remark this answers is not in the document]
sarayourfriend: I really think we should just delete it and re-prioritize the issue to make it work in CI. Or else I'm happy to add an `if CI` in here.

This line literally makes local test debugging impossible and I actually know what is happening. For a new contributor it would be baffling.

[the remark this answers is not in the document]
AetherUnbound: I'm fine with that, honestly. It seems like it's been a real pain for API tests!

zackkrida (approved):
I love the way you've handled this. In the future it would be lovely to be able to manage this from within the Django admin UI but this is a really nice approach. I really like how flexible it is towards other future throttle exemptions.

stacimc (approved):
✨ Beautiful, so flexible and very clear to understand. Looks good to me!

PR #569, opened 2022-03-15
Stop destroying test output with automatic logging

### LINKED ISSUE #520: Improve ingestion server test output
## Problem

We call `docker-compose logs` after the tests fail, which I guess is probably helpful sometimes, but most helpful _all the time_ is the actual output of the tests that failed. Because the logs are so long the will remove the test output from many terminal emulator's default scroll back settings.

## Description

Can we spit the logs into a git ignored `ingestion_test.log` file or something? And in the GitHub CI `cat` that file as a separate step so that it's still visible in the CI where it might be useful but doesn't clog up local terminals?

## Implementation

sarayourfriend:
## Fixes

Fixes #520 by @sarayourfriend 

## Description

Stops the destructive logging we were doing but maintains helpful-for-debugging logs in CI.

## Testing Instructions

Run `just api-testlocal` with some tests that will fail (`assert True is False`) and confirm that it does not log away the test output locally.

I'll push a commit with a failing test to confirm that it does still log in CI

## Checklist

- [x[ I tried running the project locally and verified that there are no visible errors.

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -128,8 +128,15 @@ jobs:
 
       - name: Start API, ingest and index test data
         run: just init
+
       - name: Run API tests
+        continue-on-error: true
         run: just api-test -T
+
+      - name: Print API test failure logs
+        if: {{ failure() }}
+        run: just logs
+
       - name: Run analytics tests
         run: just nl-test -T
 
--- a/api/test/run_test.sh
+++ b/api/test/run_test.sh
@@ -18,8 +18,7 @@ succeeded=$?
 if [[ $succeeded -eq 0 ]]; then
 	printf "${green}:-) All tests passed${endcol}\n"
 else
-    printf "Full system logs:\n"
-    docker-compose logs
-	printf "${red}:'( Some tests did not pass${endcol}\n"
+    printf "\n\n${red}:'( Some tests did not pass${endcol}\n"
+    printf "Hint: \`just logs [service]\` will print the docker logs and may be helpful for debugging.\n\n"
 fi
 exit $succeeded


.github/workflows/ci_cd.yml
```diff
@@ -128,8 +128,15 @@ jobs:
...
       - name: Start API, ingest and index test data
         run: just init
+
       - name: Run API tests
+        continue-on-error: true
         run: just api-test -T
+
+      - name: Print API test failure logs
+        if: {{ failure() }}
+        run: just logs
```
AetherUnbound, line 138: Should this stop as well, since the previous step failed?

api/test/run_test.sh
```diff
@@ -18,8 +18,7 @@ succeeded=$?
 if [[ $succeeded -eq 0 ]]; then
 	printf "${green}:-) All tests passed${endcol}\n"
 else
-    printf "Full system logs:\n"
-    docker-compose logs
```
AetherUnbound, line 22: So apparently this has never worked for `just api-test` because `docker-compose` isn't available within the container :upside_down_face:

AetherUnbound (commented):
Looks good, just one note on the GHA flow

.github/workflows/ci_cd.yml
sarayourfriend: I was going to try debugging whether you have to explicitly `exit 1` or something or if the `if: {{ failure() }}` runs just the "failure" jobs and ignores the rest.

api/test/run_test.sh
sarayourfriend: Ahhh, I use `api-testlocal` because it's easier to interface with `pdb` than having to `docker attach openverse-web-1` or whatever.

sarayourfriend:
Just marking ready for review to test CI but it's still not confirmed to be working.

sarayourfriend:
Apparently printing the API test logs times out the GitHub actions :skull: 

I think it'd be best to pipe them to a file and then upload them as an artifact.

sarayourfriend:
Nope, I was wrong. The logs are automatically followed by the justfile. Removing the `-f` default.

justfile
```diff
@@ -44,7 +44,7 @@ recreate:
     @just up "--force-recreate --build"
 
 # Show logs of all, or named, Docker services
-logs services="" args="-f":
+logs services="" args="":
```
AetherUnbound, line 47: I don't love this change, as it'll mean having to specify `-f` every time I run this locally :confused: 

Could we perhaps have this function similar to the `IS_PROD` flag, maybe with `IS_CI`, and that automatically removes the `-f`?

sarayourfriend: Yeah, that'd be good! We could use the same IS_CI flag to automatically add `-T` to the other docker-compose commands as well to avoid having to remember that when adding new actions :tada:

AetherUnbound: That's a great point!

zackkrida:
@obulat could you look at this PR this week?

sarayourfriend:
Rebasing this so the actions actually run and we can verify they work :+1:

justfile
AetherUnbound: @sarayourfriend if you're okay with it, I can work on adding `IS_CI` to this branch :slightly_smiling_face:

dhruvkb (commented):
Can we also export the logs as an artifact? Having the logs locally to scan through would be massively helpful because the GitHub Actions UI is not that good imo.

sarayourfriend:
@dhruvkb Sure, I actually did at one point upload them as an artifact. I've put it back now. Will remove the WIP once the actions are passing as expected.

dhruvkb:
Feel free to use any code regarding logs from 0283304 (now removed from the other PR).

sarayourfriend:
@dhruvkb TIL you can download the raw output of actions:

image

Uploading the logs as an artifact still makes sense though because the raw output includes control characters for color output and makes it very annoying to read :unamused:

sarayourfriend:
This is ready for review now @AetherUnbound @dhruvkb @obulat

justfile
```diff
@@ -44,7 +46,7 @@ recreate:
     @just up "--force-recreate --build"
 
 # Show logs of all, or named, Docker services
-logs services="" args="-f":
+logs services="" args=(if IS_CI != "" { "" } else { "-f" }):
```
AetherUnbound, line 49: woah, TIL! Very cool

justfile
```diff
@@ -1,5 +1,7 @@
 set dotenv-load := false
 
+IS_CI := env_var_or_default("CI", "")
```
AetherUnbound, line 3: [TIL here as well!](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) This is great

AetherUnbound, line 3: [TIL here as well!](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) This is great

AetherUnbound (approved):
I love it! :sparkles: :rocket: 
Small nit: for the API, the steps are print -> upload, but for the ingestion server they're upload -> print. It might be good for consistency to have these both be either one or the other.

.github/workflows/ci_cd.yml
```diff
@@ -126,11 +141,26 @@ jobs:
...
       - name: Start API, ingest and index test data
         run: just init
+
       - name: Run API tests
         run: just api-test -T
 
+      - name: Print API test logs
+        if: ${{ always() }}
+        run: |
+          just logs > api_logs
```
dhruvkb, line 151: `just logs` without a service name will print the logs of all services, not just `web`. Also this will intermix all logs for all services. IMO it would be clearer to call this separately with individual service names and export them to different files.

api/test/run_test.sh
```diff
@@ -18,8 +18,7 @@ succeeded=$?
...
-	printf "${red}:'( Some tests did not pass${endcol}\n"
+    printf "\n\n${red}:'( Some tests did not pass${endcol}\n"
```
dhruvkb, line 21: This bit seems to be using 4 spaces for indentation instead of one tab as used in the rest of the file.

api/test/run_test.sh
```diff
@@ -18,8 +18,7 @@ succeeded=$?
...
+    printf "Hint: \`just logs [service]\` will print the docker logs and may be helpful for debugging.\n\n"
```
dhruvkb, line 22: Minor nit, sorry:
```suggestion
    printf "Hint: \`just logs [service]\` will print the Docker logs and may be helpful for debugging.\n\n"
```

ingestion_server/test/run_test.sh
```diff
@@ -18,8 +18,6 @@ succeeded=$?
 if [[ $succeeded -eq 0 ]]; then
 	printf "${green}:-) All tests passed${endcol}\n"
 else
-  printf "Full system logs:\n"
-  cat test/ingestion_logs.txt
 	printf "${red}:'( Some tests did not pass${endcol}\n"
```
dhruvkb, line 21: Just for consistency between the two `run_test.sh` scripts, we could add the same message here with the hint to use `just logs`.

dhruvkb (changes requested):
I have a few suggestions:

.github/workflows/ci_cd.yml
sarayourfriend: I agree with Madison on this one. Additionally, we'd have to remember to continue to add the services to the list of logs we pull if more relevant services are added.

api/test/run_test.sh
sarayourfriend: The file was already mixed. I've fixed it to use four spaces everywhere.

dhruvkb (approved):

PR #574, opened 2022-03-17
Use `https` for hyperlinked APIs by proxying through NGINX

### LINKED ISSUE #573: Return secure URLs for the fields thumbnail, detail_url and related_url.
_(Framed the verbiage of the title as a feature request)_ 🙏 

## Problem

The response for search and detail requests includes insecure URLs (`http` instead of `https`) in the fields `thumbnail`, `detail_url` and `related_url`.

image

e.g.:

**Search**

https://api.openverse.engineering/v1/images/?q=flower

**Detail:**

https://api.openverse.engineering/v1/images/6c1769b0-a3e5-4dae-8a36-8531a6e1430f/

## Description

When trying to integrate Openverse with some code on the browser I ended up having to replace the scheme part of the URL for avoiding notices like ```xxxx was loaded over HTTPS, but requested an insecure resource 'http://api.openverse.engineering/v1/images/6c1769b0-a3e5-4dae-8a36-8531a6e1430f/'. This request has been blocked; the content must be served over HTTPS.`

### LINKED ISSUE #581: Create a local SSL certificate with `mkcert`
## Problem

Currently if one wants to use `https://` locally, they have to bypass browser warnings about the page being insecure or implement a completely custom solution for creating certificates. Or, perhaps worst, not use SSL at all. It's worth noting that local SSL isn't about _security_, rather parity with production for testing. See this article for a list of [when to use SSL locally](https://web.dev/when-to-use-local-https/).

## Description

Similar to the implementation in the [frontend](https://github.com/WordPress/openverse-frontend/pull/1144), we should use the [mkcert](https://github.com/FiloSottile/mkcert#installation) library to enable contributors to set up a local certificate authority and create a certificate for local development. Some thing we'd likely want to do in this PR:

- Add instructions for installing and configuring mkcert
- Add a just script for generating the certificate
- Add a mechanism to tell Django about the local certificate and key
- Ignore the local certificate and key in `.gitignore`

dhruvkb:
## Fixes

Fixes #573 by @oskosk 
Fixes #581 by @zackkrida 

This is a <s>proposed alternative</s> proper successor to #576.

## Description

This PR configures Django to use the hostname from the `USE_X_FORWARDED_HOST` and treat proxied requests as secure.

This has some security implications as described [here](https://docs.djangoproject.com/en/4.0/ref/settings/#use-x-forwarded-host). To address these implications and to ensure that these headers are present when proxying via NGINX, infrastructure configuration might need to be changed.

## Testing Instructions

0. Check out the PR.
1. Bring up all services using `just up`.
2. Visit the dev server at `https://localhost/v1/images` and observe that all URLs are `https` and not `http`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
It'd be nice if we could actually just use SSL always for all our services, even inside our VPC and on local :thinking:

dhruvkb:
True, but this might need input from folks with infra experience @AetherUnbound and @rbadillap to understand how we can have a setup that maintains prod-dev parity while allowing integration with AWS features like load-balancers and network gateways in prod.

dhruvkb:
I set up a proxy via NGINX (with a self-signed cert) and thus we can verify that 31efb52 fixes the protocol issue.

image

sarayourfriend:
> But as you've mentioned before there are easier alternatives like [Caddy](https://caddyserver.com/) that could remove the pain of configuring/managing certificates

Apparently you're not really supposed (or even allowed) to use certbot for localhost SSL: https://letsencrypt.org/docs/certificates-for-localhost/#for-local-development

I've read some articles about this when I was trying to do the e2e docker network over SSL and you're supposed to issue local certs using a tool like [mkcert](https://github.com/FiloSottile/mkcert) that each developer trusts on their own machine using their own local CA.

In production we could use actual certs (either from certbot or from Cloudflare, wherever we want).

obulat:
As an assigned reviewer, I don't really feel qualified to review this PR, @dhruvkb, especially because it has security implications. Maybe it's better to formally request reviews from people with more infra experience?

dhruvkb:
Yeah @sarayourfriend you're right about that. `certbot` and Let's Encrypt don't work for localhost, which is why the PR uses a self-signed certificate. This causes the browser to warn before allowing you to visit the site (and shows a broken lock icon all the time). But this is much safer than having to share certificates between team members, especially considering the OSS nature of the work and the lack of a boundary between the core committers and contributors.

docker-compose.yml
```diff
@@ -123,6 +123,17 @@ services:
     env_file:
       - ./analytics/env.docker
 
+  proxy:
+    build: ./nginx
+    image: openverse-proxy
+    ports:
+      - "80:80"
+      - "443:443"
```
sarayourfriend, line 131: Should these proxy on the ports we're already accustomed to finding the API on, rather than 80/443?

docker-compose.yml
```diff
@@ -123,6 +123,17 @@ services:
     env_file:
       - ./analytics/env.docker
 
+  proxy:
```
sarayourfriend, line 126: I wonder if adding a proxy locally is the simplest approach. Could we alternatively add the certificates to the built Django container and use [gunicorn's SSL support](https://docs.gunicorn.org/en/stable/settings.html#ssl)?

I suppose that doesn't approximate the production environment and wouldn't solve the issue in production. In that sense I am curious what @rbadillap will say, whether it's worth using SSL even with a proxy (I don't see how it could hurt, and if anything seems wise considering we don't _actually_ control the infrastructure running in AWS and have no idea what HTTP routing in AWS's internal networks is actually doing; that might be overly paranoid though).

docker-compose.yml
dhruvkb: This allows us to use the standard ports for HTTP and HTTPs will continuing to expose the Gunicorn service directly on 8000 (to debug or see logs).

docker-compose.yml
dhruvkb: Yeah, the main reason for using NGINX was to replicate the prod behaviour locally. Otherwise there would definitely be many ways to simply serve the API over HTTPS directly from the Django container.

docker-compose.yml
sarayourfriend: Binding low ports requires root access on Linux (not sure about other OSs). I run rootless docker so this doesn't even work for me with my typical workflow.

I could run just this stack as root, but that obviates the security benefits of using rootless docker :shrug:

dhruvkb: Rootless Docker is good. I hadn't considered that scenario. What ports other than 8000 would be okay, (considering all the occupied ports and 8443 being taken by frontend)?

sarayourfriend: `wp-env` runs the HTTPS version one port higher than the HTTP version. We could do the same here and use 8001 (though preferably we wouldn't be relying on reserved port spaces and move to something in the [unreserved range](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports) :stuck_out_tongue: not necessary to change that for this PR though, it's something we would need to address over the whole stack).

AetherUnbound (commented):
I have to admit, I'm no cert expert! I've mostly been able to fumble my way through them for all the cases I've needed :sweat_smile: I have no strong opinions about whether we should use HTTPS within the VPC or how we'd do so. 

It seems like we wouldn't need to need to set `IS_PROXIED` since it's the default. One thing that concerns me though is that the nginx proxy on the API instances doesn't seem to be using https. It receives requests from the AWS LB on port 8080 and forwards them onto the API on port 8081. I don't know if the second proxy layer will preserve the https status all the way to Django. I can try and test this on staging though and see what happens!

AetherUnbound:
Yea, unfortunately requests are coming into us with a `scheme` of `http`, and from what I read it's a bad security practice to always set the `X-Forwarded-Proto` to `https` :disappointed: 

This may be a really naive question, but would it make sense to do string manipulation on these links instead? That way we don't have to worry about the proxy situation (for now)?

dhruvkb:
@AetherUnbound that is doable and we will can quickly implement it too. But I'm slightly against that practice because that, unlike this, is a hack instead of a true solution.

sarayourfriend:
The string manipulation approach does feel like a hack.

Another option would be to switch to relative URLs instead of indicating an explicit scheme: https://www.django-rest-framework.org/api-guide/serializers/#absolute-and-relative-urls

But this would be a breaking change and require a new version of the API. It'd also complicate client code in a way that would be nice to avoid.

It seems like the most "correct" approach would be to run Django itself on HTTPS. Indeed, that's how I've seen large Django deployments work in the past, even when there were proxies in front of it.

If we did that we'd just use [the built in Django affordances for upgrading all requests and the hyperlink serializer would adopt the correct scheme based on the incoming request](https://docs.djangoproject.com/en/4.0/ref/settings/#secure-ssl-redirect).

Until we can do that though, I think the string manipulation is a good "patch" fix for this.

dhruvkb:
@sarayourfriend if you'd prefer the string hack to be deployed while we approach a long-term fix, you can approve #576 and we'll merge it.

zackkrida:
@rbadillap thanks for the general explanation of https and certificates. 

I think the main counter-argument I would make is that the goal of local SSL _is not_ security, rather parity with production in the following areas ([this article](https://web.dev/when-to-use-local-https/) outlines the use cases well):

- Setting Secure cookies in a consistent way across browsers
- Debugging mixed-content issues
- Using HTTP/2 and later
- Using third-party libraries or APIs that require HTTPS
- Using a custom hostname

All of these with the exception of the last one (which maybe we'd _want to do_ locally in the future) are relevant to Openverse, if not outright requirements.

Regarding your questions @sarayourfriend:

> In that case @zackkrida it seems like for the frontend we should actually be using a proxy there as well, to more closely emulate production? 

I'm not positive, but I don't think so. I think what we're testing is consistency of visiting Openverse in the browser with HTTPS and less concerned about _how_ that certificate is assigned. I think that would be the main concern in the frontend. the implementation of our infra could change and I don't necessarily think we'd want the mechanism for testing SSL on the frontend to need to change at the same time.

> And explicitly not encourage running the Nuxt server directly with SSL?

I don't think we need to discourage this, for the reasons above, but perhaps it means that http should be the default dev experience and that https testing should only be used sparingly to debug issues related to the listed points above.

sarayourfriend:
> I'm not positive, but I don't think so. I think what we're testing is consistency of visiting Openverse in the browser with HTTPS and less concerned about how that certificate is assigned. I think that would be the main concern in the frontend. the implementation of our infra could change and I don't necessarily think we'd want the mechanism for testing SSL on the frontend to need to change at the same time.

I think it means we need to also do the header forwarding that's done in Django but for Nuxt? I'm not sure what currently tells the frontend what scheme to use (perhaps the existing schema on the session?)

dhruvkb:
With the newest changes, the setup looks very good to me.

- Uses `mkcert` for the certificates to bypass the spooky warning.
- Runs Gunicorn in `http` mode because the Gunicorn-NGINX communication is within our trusted network.
- Serves `https` from NGINX with non-root ports.
  - `http://localhost:9080` redirects to `https://localhost:9443`
- URLs in the response body have the right hostname in all scenarios, making them directly usable to navigate the API.

zackkrida (approved):
Works great and I appreciate the parity between dev and prod. One non-blocking suggestion but otherwise LGTM.

api/docs/guides/https.md
```diff
@@ -0,0 +1,13 @@
...
+
+To emulate the proxying behavior of production, we have set up an NGINX proxy that serves the API over `https` by proxying to the Gunicorn server which serves over `http`.
+
+This proxy uses certificates from the `nginx/certs/` directory. These certificates can be generated using `mkcert` with the `just cert` command.
+
+Additionally, to test with the dev hostname `dev.openverse.engineering`, add the following line to your hosts file.
+
+```text
+127.0.0.1 dev.openverse.engineering
+```
```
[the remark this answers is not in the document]
dhruvkb: Fixed in 972433e.

AetherUnbound (approved):
This solution is excellent! Very easy for me to set up, and doesn't interfere with http traffic on port 8000. I love it :tada:

stacimc:
This sounds good to me, but I had trouble testing it. I followed the instructions added here to the `quickstart` guide and installed `mkcert` and `nss`.

`mkcert -V` gives me version 1.5.0, but I get `error: unknown option '-install'` when running `mkcert -install`, and similarly `just cert` throws `error: unknown option '-cert-file'`. I feel like I'm missing something obvious 😅

AetherUnbound:
That's weird! I don't have a `-V` flag, but I did get this:
```
$ mkcert -version
v1.4.3
```

Maybe that's the necessary version? FWIW I installed with `brew install mkcert`

zackkrida:
@stacimc I suspect the wrong `mkcert` was somehow installed. It should be [this one](https://github.com/FiloSottile/mkcert) which is presently on version `1.4.3`.

I do see _this one_ https://github.com/Subash/mkcert is currently on `v1.5.0`!

stacimc:
Thanks @zackkrida, @AetherUnbound -- I had automatically installed with `npm` at first, then realized my mistake, uninstalled and reinstalled correctly with `brew`. It looks like I must not have _uninstalled_ correctly the first time and this was causing my problem 😅

Uninstalled properly and it's all good now! ✅

PR #583, opened 2022-03-23
Replace plural `categories` as field name with singular `category`

obulat:
## Fixes

Fixes [@openverse-frontend#1169](https://github.com/WordPress/openverse-frontend/issues/1169)

## Description

Note: this does not really use category when filtering the search result, hence Draft status. 
But setting category to singular is correct because all other parameters use singular words.
`category` field was using plural and singular forms interchangeably in the codebase. This caused errors when filtering by `category`:
- for audio it always returned 0 results
- for images, it seems to have returned correctly filtered results, but with `category` field sent to the front end being set to `null`.

To ensure that this change doesn't break anyone's code that was using `categories` as an image filter, ~~I added it to the list of deprecated parameters.~~ I removed it from the deprecated parameters, and added a duplicate filter to the search controller to make sure that both parameters work

## Testing Instructions

Run `just up`, `just init`, and try searching for something, filtering it by category.
`http://localhost:8000/v1/audio/?q=look%20back&category=music` should return 1 result

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/test/audio_integration_test.py
+++ b/api/test/audio_integration_test.py
@@ -10,6 +10,7 @@ from test.media_integration import (
     report,
     search,
     search_all_excluded,
+    search_by_category,
     search_consistency,
     search_quotes,
     search_source_and_excluded,
@@ -32,6 +33,11 @@ def audio_fixture():
 
 def test_search(audio_fixture):
     search(audio_fixture)
+
+
+def test_search_category_filtering(audio_fixture):
+    search_by_category("audio", "music", audio_fixture)
+    search_by_category("audio", "pronuntiation", audio_fixture)
 
 
 def test_search_all_excluded():
--- a/api/test/audio_integration_test.py
+++ b/api/test/audio_integration_test.py
@@ -37,7 +37,7 @@ def test_search(audio_fixture):
 
 def test_search_category_filtering(audio_fixture):
     search_by_category("audio", "music", audio_fixture)
-    search_by_category("audio", "pronuntiation", audio_fixture)
+    search_by_category("audio", "pronunciation", audio_fixture)
 
 
 def test_search_all_excluded():
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -240,6 +240,7 @@ def search(
     filters = [
         ("extension", None),
         ("category", None),
+        ("categories", "category"),
         ("aspect_ratio", None),
         ("size", None),
         ("source", None),
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -239,7 +239,7 @@ def search(
     # names are identical.
     filters = [
         ("extension", None),
-        ("categories", None),
+        ("category", None),
         ("aspect_ratio", None),
         ("size", None),
         ("source", None),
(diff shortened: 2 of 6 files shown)


sarayourfriend (commented):
This would require a major version change in it's current state.

If we don't want to create a new API version and we want to have `category` as singular then we need to support both `category` and `categories`.

zackkrida:
I think I'd prefer to support both, with a note that in the next major version one will be removed. Ideally by that time we could email our API consumers or something like that, too.

obulat:
I didn't understand what `deprecated` parameter means, to be honest. I thought it works like the deprecation notices in some python packages: when something is deprecated, it still works, but displays a warning. Apparently, it doesn't work at all now.

krysal (approved):
Works great! The category filter returns exactly the same count from the database 💯

```
http://localhost:8000/v1/audio/?q=&category=music

{
	"result_count": 180,
	...
}
```

```
SELECT COUNT(id) FROM audio WHERE category='music';
```

| count |
| - |
| 180 |

We might need to update the test data for audio, it's difficult to get results for anything without looking at the DB because the majority of titles are in German 😅 and there are only two categories: music and pronunciation.

I added a small test for this in #584, let me know if you want to merge it here. Thanks for finding and addressing this so quickly 👏  It's a relief that we don't have to do migrations for this.

obulat:
> I think I'd prefer to support both, with a note that in the next major version one will be removed. ~Ideally by that time we could email our API consumers or something like that, too.~ Edit: Forgot the entire point of version numbers for a second 🤦

Should I open an issue for allowing for this pattern of deprecation (use both options in current version and display a deprecation warning, and treat it as an error in the future version, as is done now with deprecations)? Or do we already have it available in the API, and I haven't found how to use it? @dhruvkb , what do you think?

dhruvkb:
I don't think we have a deprecation pattern yet. The only API version I've seen was v1 (and it sends error messages when you send the old parameters for licenses)

We do have a bit of a deprecation system here:

https://github.com/WordPress/openverse-api/blob/a5f8b48f87856d6a40971d5c2b55a4ecd026c165/api/catalog/api/serializers/media_serializers.py#L149

sarayourfriend:
Why have a version prefix if we're going to deploy breaking changes?

Unless we're prepared to hold off deploying breaking changes until our small number of API consumers have confirmed that they're ready to handle it, then we're _really_ misleading people by having a version prefix in the API routes.

dhruvkb:
The v1 prefix is kind of misleading anyway. Technically, with all the changes we've made so far, we've already broken backward compatibility with the CC Search's v1 API.

AetherUnbound (approved):
This looks great! I can't comment too much on the API versioning, as that's not something I have too much experience with.

The database is congruent with `category` as a singular word, so from that perspective I don't foresee any issues!

api/test/audio_integration_test.py
```diff
@@ -34,6 +35,11 @@ def test_search(audio_fixture):
     search(audio_fixture)
 
 
+def test_search_category_filtering(audio_fixture):
+    search_by_category("audio", "music", audio_fixture)
+    search_by_category("audio", "pronuntiation", audio_fixture)
```
AetherUnbound, line 40: Hmm, I'm not sure why this test _isn't_ failing because I think this should be:
```suggestion
    search_by_category("audio", "pronunciation", audio_fixture)
```

Maybe we have no pronunciation records in the sample data? :thinking:

PR #560, opened 2022-03-10
Added categories.py in constant modules to store different categories.

ritikBhandari:
Successfully stored the 2 categories in ```categories.py ```module.
Also,imported the 'categories' from ```categories.py``` to use them wherever they were previously used.

Issue #557, opened 2022-03-09
Add serializer unit tests


AetherUnbound:
## Problem

We do not currently have unit tests for any of the API serializers (`api/catalog/api/serializers/*`).

## Description

We should add thorough unit tests to these serializers, focusing on the base functions in the `media_serializers.py` file.

## Additional context

This came up as part of #526.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

sarayourfriend:
A note for whomever picks this up: please continue to extend and use the factories for models. Currently there's only factories for the Audio models and the base media factory needs to be extended further to include more of the fields, especially if we're going to unit test all the serializer fields :slightly_smiling_face:

foroveralls:
Hey would be interested in working on this.  I'm however new to this project and not familiar with factories, or the standard procedure for unit testing here, could someone provide some pointers to reference files?

sarayourfriend:
Hi @foroveralls! Thanks for offering to work on this.

Most of our test coverage is currently in integration tests, so we only have a few unit test examples, but I'll share them here to make it easier to find.

The best factory example is the AudioFactory and AudioAddOnFactory, found in this module: https://github.com/WordPress/openverse-api/blob/c09fd7e16a8eb104c311e8d4f0da08238570067c/api/test/factory/models/audio.py

From there you can inspect the base MediaFactory which will need some expanding in addition to adding a new ImageFactory based on it (similar to AudioFactory).

We use the popular [factoryboy library (docs link here)](https://factoryboy.readthedocs.io) to build Django model factories and other data factories. We've extended this with additional faker providers in this module: https://github.com/WordPress/openverse-api/blob/188b6a0c299a4b79e205c997378e230bf1f96d14/api/test/factory/faker.py

The unit tests themselves are currently limited to a management command and tests for the waveform generation logic on the Audio model. Those are linked here:

https://github.com/WordPress/openverse-api/blob/c09fd7e16a8eb104c311e8d4f0da08238570067c/api/test/unit/management/commands/generatewaveforms_test.py

https://github.com/WordPress/openverse-api/blob/c09fd7e16a8eb104c311e8d4f0da08238570067c/api/test/unit/models/audio_test.py

Generally we follow a pretty standard Django unit testing model, but with pytest instead of the Django unit test classes.

If you need any clarification about the above or additional help, please feel free to ping. And thanks in advance for your contribution! If this all sounds good to you, please confirm and we'll assign the issue to you :grin:

[...]

PR #589, opened 2022-03-24
Move media type categories to constants module

### LINKED ISSUE #558: Define categories as constants
## Problem

For both audio and images, categories are defined as strings in the serializer.

image:
https://github.com/WordPress/openverse-api/blob/88ca7f7172fe2b286091cb6a83a759d2e727d692/api/catalog/api/serializers/image_serializers.py#L37

audio:
https://github.com/WordPress/openverse-api/blob/88ca7f7172fe2b286091cb6a83a759d2e727d692/api/catalog/api/serializers/audio_serializers.py#L61

## Description

These categories should me moved into a constants module for 
- better understanding of available categories
- easier modifications in the future

Relates to #526.

AetherUnbound:
## Fixes

Fixes #558 by @dhruvkb

Supersedes #560

## Description

This PR uses the method described by @dhruvkb in #560 to move the media type categories to a constants module. In doing so, it also changes the category serializer field from a `CharField` to a `MultipleChoiceField`. This has some implications on how we convert the arguments into an Elasticsearch query, as we now need to account for the fact that the field's data is a `set` rather than a string that can be `split`.

I'm not sure this is the correct approach currently because we actually _lose_ some useful information in the validation error. Here's an example on `http://localhost:8000/v1/images/?q=wolf&category=sound`.

**Before**

```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": {
        "category": [
            "Invalid category: sound. Available options: {'illustration', 'photograph', 'digitized_artwork'}"
        ]
    }
}
```

**After**

```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": {
        "category": [
            "\"sound\" is not a valid choice."
        ]
    }
}

```

I don't know enough about DRF, but perhaps we might be able to alter the error message for category so it does display the acceptable values. I think this will require us to make a custom serializer field.

An _entirely different_ approach we could take is just to move the sets that are defined in the `validate_categories` functions on each serializer and move only that to the constants module. This would prevent us from having to handle the `set` type differently since it would still be a `CharField`, and we're already using the `_validate_enums` function elsewhere.

## Testing Instructions

`just api-test` for unit tests

Use the link above to see the above results, then change the category to `photograph` or similar and see that results get returned.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
Although now we don't need the help text, because it's generated for us:
image

obulat (commented):
TIL: the `CharField` and `MultipleChoiceField` have a different behavior when it comes to query parsing, and it's a breaking change here:
If you try to use two categories separated with commas(`http://localhost:8000/v1/images/?q=wolf&category=illustration,photograph`), you get an error:
```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": {
        "category": [
            "\"illustration,photograph\" is not a valid choice."
        ]
    }
}
```

The correct query for several `MultipleChoiceField`s would have each choice to have its own parameter:
`http://localhost:8000/v1/images/?q=wolf&category=illustration&category=photograph`.

So, my vote is for the _entirely different_ approach 😄

AetherUnbound:
Oof, I hadn't even considered that! I think changing this field type would pretty drastically affect the API. How do we feel about _just_ moving the list of strings that are valid categories into `constants/categories.py`?

zackkrida:
cc @dhruvkb and @obulat to discuss @AetherUnbound's follow-up question

zackkrida:
> How do we feel about just moving the list of strings that are valid categories into constants/categories.py?

Let's move forward with this, [as discussed](https://wordpress.slack.com/archives/C02012JB00N/p1649171676474039)!

AetherUnbound:
I've reverted most of the changes and moved the sets out to their own file. I also made a `Categories` subclass of `set` so we'd have a handy `make_help_text` function on it. This might be way overkill...but I wanted to make it so that our help text doesn't drift from the actual category set we have in code.

Lastly, I uncovered an issue that was introduced in #583 - we changed the category field from `categories` to `category`, but didn't change the `validate_categories` function to match. Unfortunately because some of this is DRF-magic, it wasn't immediately obvious that this meant the validation function for `category` was no longer being called. One of the test cases I added in `audio_integration_tests` caught that, so hopefully it won't happen again in the future (at least for this field). Yet another nudge from our repository to work on #557 I suppose :sweat_smile:

AetherUnbound:
Wow, the ordering doesn't seem deterministic! I'm sure it's because they're `set`s. This passed locally for me 😅 #WorksOnMyMachine

I'll fix that!

api/catalog/api/constants/categories.py
```diff
@@ -0,0 +1,35 @@
+class Categories(set):
+    def make_help_text(self) -> str:
```
dhruvkb, line 2: This is partially duplicated here:

https://github.com/WordPress/openverse-api/blob/f2ff9ae525ae060b65608b08f34d6fbc193d4346/api/catalog/api/serializers/media_serializers.py#L68

dhruvkb: Yours is a better implementation, might want to delete usage of this^ one.

AetherUnbound:
I've re-requested review because I've made some significant changes. Instead of having a specific `Categories` class, I've moved that help text functionality into its own module. I then applied that module everywhere that seemed reasonable in the serializers list. Here's a before (left) and after (right) on the help text:
image

api/catalog/api/serializers/image_serializers.py
```diff
@@ -38,34 +44,33 @@ class ImageSearchRequestSerializer(
...
     @staticmethod
     def validate_aspect_ratio(value):
-        valid_ratios = {"tall", "wide", "square"}
-        _validate_enum("aspect ratio", valid_ratios, value)
+        _validate_enum("aspect ratio", ASPECT_RATIOS, value)
+        return value.lower()
+
+    @staticmethod
+    def validate_size(value):
+        _validate_enum("size", IMAGE_SIZES, value)
```
AetherUnbound, line 73: This validation wasn't happening before, so invalid `size` entries were permissible. This is now updated using the values defined in the help text.

dhruvkb:
The updated documentation is dope. 💯 &times; 💯  QoL improvement.

api/catalog/api/constants/field_values.py
```diff
@@ -0,0 +1,31 @@
+AUDIO_CATEGORIES = {
```
dhruvkb, line 1: I think we should segregate the file by media types, or use `AUDIO_`/`IMAGE_` prefix before every variable. Otherwise it'll quickly become confusing as we add more media types.

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -62,35 +64,23 @@ class AudioSearchRequestSerializer(
 
     category = serializers.CharField(
         label="category",
-        help_text="A comma separated list of categories; available categories "
-        "include `music`, `sound_effect`, `podcast`, `audiobook`, "
-        "and `news`.",
+        help_text=make_comma_separated_help_text(AUDIO_CATEGORIES, "categories"),
```
dhruvkb, line 67: This is fantastic, but a small part of me wants to extend `CharField`, give it a keyword argument (like `options` or `values`) and in `__init__`, generate this help text automatically. Again, this is clear and excellent and definitely a huge improvement **to merge as is**, but it is a bit repetitive.

api/catalog/api/utils/help_text.py
```diff
@@ -0,0 +1,16 @@
+from typing import Iterable
+
+
+def make_comma_separated_help_text(items: Iterable, name: str) -> str:
+    """
+    Generate help text that wraps each category in backticks.
```
dhruvkb, line 6: We should document the params and the returned value.
```suggestion
    Generate help text that wraps each category in backticks.
    :param items: ...
    :param name: ...
    :return: ...
```

dhruvkb (approved):
Huge improvement over the present situation. I have some comments but they're not blocking.

api/catalog/api/constants/field_values.py
AetherUnbound: I was thinking about this...but there are some fields that might be relevant across media types. Like `DURATION` - I could see that being used for both audio _and_ video.

api/catalog/api/utils/help_text.py
AetherUnbound: Ah, and this also no longer describes `category` operations! I'll update this.

api/catalog/api/serializers/audio_serializers.py
AetherUnbound: I started toying with this a bit; it gets a little more complicated...which is understandable, since we're turning more of this into "magic". I'd love to get this PR merged, then open a draft PR with an approach to this and see if folks would prefer that 🙂

obulat (approved):

PR #593, opened 2022-03-25
Accurately report status of data_refresh when work is distributed

### LINKED ISSUE #592: Data Refresh inaccurately reports status
## Description

This bug was discovered in the investigation in https://github.com/WordPress/openverse-catalog/issues/437. I'm updating it to high priority because it prevents us from turning on the refresh DAGs at the same time.

When running a data refresh locally, we immediately and synchronously index the table before exiting the data refresh task. In staging/production, [we schedule multiple indexer_worker ](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/indexer.py#L414)instances to perform a distributed index. The data refresh task schedules the workers, then [logs and exits immediately](https://github.com/WordPress/openverse-api/blob/95820a73f090f2627ab516abe0a8b6bed363190d/ingestion_server/ingestion_server/tasks.py#L119).

This causes a bug in the status_check endpoint, because it uses `task.is_alive()` to determine whether the refresh is active -- and this is only accurate in the local env. Not only will it report that the task is no longer active as soon as the work is scheduled, it also [misleadingly reports that there is an error](https://github.com/WordPress/openverse-api/blob/95820a73f090f2627ab516abe0a8b6bed363190d/ingestion_server/ingestion_server/api.py#L138) (because the task 'stopped' (is inactive) before percent_completed == 100).

## Reproduction

Unfortunately this has to be reproduced in production by running one of the audio or image refreshes. You can look at the logs for the audio data refresh's recent runs to see the errors.

## Resolution

stacimc:
## Fixes

Fixes #592 by @stacimc

## Description

This PR updates the ingestion server to track the status of distributed reindexer workers, and more accurately report the 'active' status of the data refresh task.

Locally, the data refresh task is entirely synchronous. In staging and production, the elasticsearch reindexing step is distributed across worker tasks. The data refresh task schedules this work, then immediately exits. When you poll the ingestion server to check on the status of the task, it reports that it is inactive because the task has completed, even though distributed work is ongoing.

This PR basically:
* adds an `active_workers` property to the already tracked `TaskStatus`, which is by default 'False' when a task is created.
* if we're in staging or production, we'll set this to `True` before schedule distributed reindexing workers, and update it to `False` when the final `go_live` step completes, **OR** if one of the workers fails during reindexing
* updates the `TaskStatus` endpoint to:
    *  take active workers into consideration when reporting the status of the task: `active = task.is_alive() or active_workers`
    * report `percent_completed` as the number of workers that have finished successfully, when running in a distributed environment (staging or prod)
* updates the indexer workers to report errors they encounter to Slack
* only attempts to go live with the new indices if no workers encountered errors

## Testing Instructions

This has to be tested in staging.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -87,6 +87,7 @@ class TaskResource(BaseTaskResource):
         # Inject shared memory
         progress = Value("d", 0.0)
         finish_time = Value("d", 0.0)
+        active_workers = Value("i", int(False)) # Tracks whether the task has any active distributed workers
         task = Task(
             model=model,
             task_type=TaskTypes[action],
@@ -94,10 +95,11 @@ class TaskResource(BaseTaskResource):
             progress=progress,
             task_id=task_id,
             finish_time=finish_time,
+            active_workers=active_workers,
             callback_url=callback_url,
         )
         task.start()
-        task_id = self.tracker.add_task(task, task_id, action, progress, finish_time)
+        task_id = self.tracker.add_task(task, task_id, action, progress, finish_time, active_workers)
         base_url = self._get_base_url(req)
         status_url = f"{base_url}/task/{task_id}"
         # Give the task a moment to start so we can detect immediate failure.
@@ -129,9 +131,11 @@ class TaskStatus(BaseTaskResource):
     def on_get(self, req, resp, task_id):
         """Check the status of a single task."""
         task = self.tracker.id_task[task_id]
-        active = task.is_alive()
 
         percent_completed = self.tracker.id_progress[task_id].value
+        active_workers = bool(self.tracker.id_active_workers[task_id].value)
+        active = task.is_alive() or active_workers
+
         resp.media = {
             "active": active,
             "percent_completed": percent_completed,
@@ -139,7 +143,7 @@ class TaskStatus(BaseTaskResource):
         }
 
 
-class WorkerFinishedResource:
+class WorkerFinishedResource(BaseTaskResource):
     """
     For notifying ingestion server that an indexing worker has finished its
     task.
@@ -147,21 +151,23 @@ class WorkerFinishedResource:
 
     @staticmethod
     def on_post(req, _):
-        target_index = worker_finished(str(req.remote_addr))
-        if target_index:
+        task_data = worker_finished(str(req.remote_addr))
+        if task_data:
             logging.info(
                 "All indexer workers finished! Attempting to promote index "
-                f"{target_index}"
+                f"{task_data['target_index']}"
             )
-            index_type = target_index.split("-")[0]
+            index_type = task_data['target_index'].split("-")[0]
             if index_type not in MEDIA_TYPES:
                 index_type = "image"
             slack.verbose(
                 f"`{index_type}`: Elasticsearch reindex complete | "
                 f"_Next: promote index as primary_"
             )
+            task_id = task_data['task_id']
+            active_workers = tracker.id_active_workers[task_id]
             f = indexer.TableIndexer.go_live
-            p = Process(target=f, args=(target_index, index_type))
+            p = Process(target=f, args=(task_data['target_index'], index_type, active_workers))
             p.start()
 
 
@@ -195,7 +201,7 @@ def create_api(log=True):
     _api.add_route("/", HealthResource())
     _api.add_route("/task", TaskResource(task_tracker))
     _api.add_route("/task/{task_id}", TaskStatus(task_tracker))
-    _api.add_route("/worker_finished", WorkerFinishedResource())
+    _api.add_route("/worker_finished", WorkerFinishedResource(task_tracker))
     _api.add_route("/state", StateResource())
 
     return _api
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -160,28 +160,36 @@ class WorkerFinishedResource(BaseTaskResource):
 
     def on_post(self, req, _):
         task_data = worker_finished(str(req.remote_addr), req.media["error"])
-        task_id = task_data["task_id"]
-        # Update progress
-        self.tracker.id_progress[task_id] = task_data["percent_completed"]
+        task_id = task_data.task_id
+        target_index = task_data.target_index
+        active_workers = self.tracker.id_active_workers[task_id]
 
-        if task_data["percent_completed"] == 100:
+        # Update progress
+        self.tracker.id_progress[task_id] = task_data.percent_successful
+
+        if task_data.percent_successful == 100:
             logging.info(
-                "All indexer workers finished! Attempting to promote index "
-                f"{task_data['target_index']}"
+                "All indexer workers succeeded! Attempting to promote index "
+                f"{target_index}"
             )
-            index_type = task_data["target_index"].split("-")[0]
+            index_type = target_index.split("-")[0]
             if index_type not in MEDIA_TYPES:
                 index_type = "image"
             slack.verbose(
                 f"`{index_type}`: Elasticsearch reindex complete | "
                 f"_Next: promote index as primary_"
             )
-            active_workers = self.tracker.id_active_workers[task_id]
             f = indexer.TableIndexer.go_live
             p = Process(
-                target=f, args=(task_data["target_index"], index_type, active_workers)
+                target=f, args=(target_index, index_type, active_workers)
             )
             p.start()
+        elif task_data.percent_completed == 100:
+            # All workers finished, but not all were successful. Mark
+            # workers as complete and do not attempt to go live with the new
+            # indices.
+            active_workers.value = int(False)
+
 
 
 class StateResource:
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -164,7 +164,7 @@ class WorkerFinishedResource(BaseTaskResource):
         target_index = task_data.target_index
         active_workers = self.tracker.id_active_workers[task_id]
 
-        # Update progress
+        # Update global task progress based on worker results
         self.tracker.id_progress[task_id] = task_data.percent_successful
 
         if task_data.percent_successful == 100:
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -157,9 +157,9 @@ class WorkerFinishedResource(BaseTaskResource):
         task_data = worker_finished(str(req.remote_addr), req.media["error"])
         task_id = task_data["task_id"]
         # Update progress
-        self.tracker.id_progress[task_id] = task_data['percent_completed']
+        self.tracker.id_progress[task_id] = task_data["percent_completed"]
 
-        if task_data['percent_completed'] == 100:
+        if task_data["percent_completed"] == 100:
             logging.info(
                 "All indexer workers finished! Attempting to promote index "
                 f"{task_data['target_index']}"
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -154,7 +154,7 @@ class WorkerFinishedResource(BaseTaskResource):
     """
 
     def on_post(self, req, _):
-        task_data = worker_finished(str(req.remote_addr))
+        task_data = worker_finished(str(req.remote_addr), req.media["error"])
         task_id = task_data["task_id"]
         # Update progress
         self.tracker.id_progress[task_id] = task_data['percent_completed']
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -179,16 +179,13 @@ class WorkerFinishedResource(BaseTaskResource):
                 f"_Next: promote index as primary_"
             )
             f = indexer.TableIndexer.go_live
-            p = Process(
-                target=f, args=(target_index, index_type, active_workers)
-            )
+            p = Process(target=f, args=(target_index, index_type, active_workers))
             p.start()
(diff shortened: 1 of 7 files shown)


ingestion_server/ingestion_server/api.py
```diff
@@ -87,17 +87,19 @@ def on_post(self, req, resp):
         # Inject shared memory
         progress = Value("d", 0.0)
         finish_time = Value("d", 0.0)
+        active_workers = Value("i", int(False)) # Tracks whether the task has any active distributed workers
```
stacimc, line 90: Following the model here from `progress` and `finish_time`, although this is new to me.

zackkrida:
Really excited to see how a test goes, I'm impressed by how clean and simple this solution appears 🤩

ingestion_server/ingestion_server/api.py
```diff
@@ -129,39 +131,43 @@ class TaskStatus(BaseTaskResource):
...
-        active = task.is_alive()
 
         percent_completed = self.tracker.id_progress[task_id].value
+        active_workers = bool(self.tracker.id_active_workers[task_id].value)
+        active = task.is_alive() or active_workers
+
         resp.media = {
             "active": active,
             "percent_completed": percent_completed,
             "error": percent_completed < 100 and not active,
```
stacimc, line 142: These changes make `error` slightly more accurate than it was before, but it's still not great:

* `error` will now correctly be `True` if `percent_completed < 100` and the _task died before starting the reindexing workers_
* however, we're currently only setting `active_workers` to False when the `go_live` step completes successfully. 

The indexer workers themselves wrap their work in a try/except and report back to the ingestion server [even if they encounter errors](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/indexer_worker.py#L84). I am fairly certain this means that `go_live` will run even if the re-indexing step fails, and that we expect to catch this at [this healthcheck step](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/indexer.py#L359) within the method. I _think_ what we'll want to do is update `active_workers` to false if that healthcheck times out, but I may be misreading how this works.

stacimc: After looking into it further, I don't understand the current state of our error handling and suggest that it might be best to leave updating that for another ticket -- or to some extent, leaving it to be handled sensibly when we move the logic to Airflow. With this work `error` here should be at least improved.

* As I already stated, I believe the indexer workers simply catch and log any errors they encounter, but when they complete the `go_live` step is called regardless
* `go_live` has a [healthcheck](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/indexer.py#L359) but it only runs in production, not staging. It waits for the es cluster to be GREEN, and will block for up to a 12 hour timeout. But as far as I can tell that check just returns when the timeout is reached and we aren't doing any sort of response checking -- so if it times out with the cluster still unhealthy, I'm pretty sure we just go on to replace the indices regardless 😰 

I can make a ticket/PR for that last point.

stacimc:
@AetherUnbound and I were able to test this (including changes through [bb5fb62](https://github.com/WordPress/openverse-api/pull/593/commits/bb5fb62820c20e1f8fa49bd566510f1ebe9ecf68)) in staging and confirm that the `active` value is now correct (`True` while the distributed workers are running, finally `False` when they finish)

We discovered however that `percent_completed` is _also_ only meaningful in local development and does not work in staging or production. This is because the progress is calculated in the [`replicate`](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/indexer.py#L298) method which is distributed in staging/prod.

I'm updating this with the following:
* In staging and production, the `percent_completed` is now the _percentage of workers that have finished without errors_.
* In staging/prod, errors that happen during the reindexing task will be reported to Slack
* If any workers encounter an error, we will _not_ attempt to `go_live`

ingestion_server/ingestion_server/api.py
stacimc: Per my comment [here](https://github.com/WordPress/openverse-api/pull/593#issuecomment-1082287575), I've updated this so that workers report back to the ingestion server when they encounter errors, and `go_live` is not called unless all workers complete successfully.

ingestion_server/ingestion_server/api.py
```diff
@@ -87,17 +87,23 @@ def on_post(self, req, resp):
         # Inject shared memory
         progress = Value("d", 0.0)
         finish_time = Value("d", 0.0)
+        active_workers = Value(
+            "i", int(False)
+        )  # Tracks whether the task has any active distributed workers
```
AetherUnbound, line 92: If you move this comment to the line above, `black` won't try to line break it:
```suggestion
        # Tracks whether the task has any active distributed workers
        active_workers = Value("i", int(False))
```

ingestion_server/ingestion_server/api.py
```diff
@@ -128,40 +134,53 @@ def on_get(self, req, resp):
...
     def on_get(self, req, resp, task_id):
         """Check the status of a single task."""
-        task = self.tracker.id_task[task_id]
-        active = task.is_alive()
+        try:
+            task = self.tracker.id_task[task_id]
+        except KeyError:
+            resp.status = falcon.HTTP_404
+            resp.media = {"message": f"No task found with id {task_id}"}
+            return
```
AetherUnbound, line 142: My be better to use `.get` rather than a try except?
```suggestion
        task = self.tracker.id_task.get(task_id)
        if task is None:
            resp.status = falcon.HTTP_404
            resp.media = {"message": f"No task found with id {task_id}"}
            return
```

ingestion_server/ingestion_server/api.py
```diff
@@ -128,40 +134,53 @@ def on_get(self, req, resp):
...
     """
 
-    @staticmethod
-    def on_post(req, _):
-        target_index = worker_finished(str(req.remote_addr))
-        if target_index:
+    def on_post(self, req, _):
+        task_data = worker_finished(str(req.remote_addr), req.media["error"])
+        task_id = task_data["task_id"]
+        # Update progress
```
AetherUnbound, line 164:
```suggestion
        # Update global task progress based on worker results
```

ingestion_server/ingestion_server/api.py
```diff
@@ -128,40 +134,53 @@ def on_get(self, req, resp):
...
+        self.tracker.id_progress[task_id] = task_data["percent_completed"]
+
+        if task_data["percent_completed"] == 100:
             logging.info(
                 "All indexer workers finished! Attempting to promote index "
-                f"{target_index}"
+                f"{task_data['target_index']}"
```
AetherUnbound, line 170: Nit: since `task_data["target_index"]` is used a few times in this block, it might be worth setting it to a variable

ingestion_server/ingestion_server/state.py
```diff
@@ -55,31 +57,44 @@ def register_indexing_job(worker_ips, target_index):
...
 
-def worker_finished(worker_ip):
+def worker_finished(worker_ip, error):
     """
     The scheduler received a notification indicating an indexing worker has
     finished its task.
     :param worker_ip: The private IP of the worker.
-    :return: The target index if all workers are finished, else False.
+    :param error: Whether this worker had an error during processing.
+    :return: A dict containing the target index, task_id, and the percent of
```
AetherUnbound, line 70: FWIW it might be worth making this dictionary an explicit `NamedTuple`, that way we can access field attributes easier and have a clear definition of available values/types

ingestion_server/ingestion_server/indexer.py
```diff
@@ -377,6 +387,9 @@ def go_live(write_index, live_alias):
         else:
             es.indices.put_alias(index=write_index, name=live_alias)
             log.info(f"Created '{live_alias}' index alias pointing to {write_index}")
+        # If there were any active workers for this task, we should mark them complete.
+        if active_workers:
+            active_workers.value = int(False)
```
AetherUnbound, line 392: It looks like this is the only place where the `active_workers` value is "turned off", is that correct?

ingestion_server/ingestion_server/state.py
```diff
@@ -55,31 +57,44 @@ def register_indexing_job(worker_ips, target_index):
...
+        completed_workers = 0
         for worker_key in db["worker_statuses"]:
-            if db["worker_statuses"][worker_key] == WorkerStatus.RUNNING:
+            status = db["worker_statuses"][worker_key]
+            if status == WorkerStatus.RUNNING:
                 log.info(f"{worker_key} is still indexing")
-                return False
-        return db["target_index"]
+            elif status == WorkerStatus.FINISHED:
+                completed_workers += 1
```
AetherUnbound, line 92: If I'm understanding this correctly: A worker that failed will report it encountered an error. It's status will be `WorkerStatus.ERROR`, so when these results are accumulated the completed count will _always_ be one less. Then, even if `active_workers` is false and thus `active` is false, `error` will still be true because `percent_completed < 100`?

AetherUnbound (commented):
Whew, this server is a doozy :joy: it looks good! Just a couple of comments

ingestion_server/ingestion_server/indexer.py
stacimc: Yes, at the moment. I was trying to account for the fact that the data refresh isn't really "done" until the `go_live` step completes, which is _after_ the workers finish and thus isn't accounted for if we turn this off as soon as the indexer workers complete.

We could introduce yet another global progress indicator to track just the status of the `go_live` step, which is turned on right before turning `active_workers` off, and turned off when the step completes. Alternatively we could continue considering this its own 'worker', so:
* turn `active_workers` _off_ when the last worker completes, _if_ any of them are in the Error state (because in this case we will not attempt to go live)
* turn `active_workers` off at the end of the `go_live` step as well

ingestion_server/ingestion_server/state.py
stacimc: Yep. That being said I now think there's a problem -- which is that `active_workers` is not set to false until the `go_live` step completes, which will not happen if a worker has errored.

I think at minimum we should turn `active_workers` off when all indexers have completed, [_if_ any of them have errored.](https://github.com/WordPress/openverse-api/pull/593#discussion_r838989489)

This definitely feels a bit shaky around the error states and I would very much prefer to get this into Airflow for better tracking 😓 I think this should at least get us to a better place than we are currently.

ingestion_server/ingestion_server/api.py
stacimc: Visually, I felt like moving the comment inline makes it clear that `active_workers` is part of the group of values including `progress` and `finish_time` that are underneath the `Inject shared memory` comment, without 'breaking up' that block of code:

```
        # Inject shared memory
        progress = Value("d", 0.0)
        finish_time = Value("d", 0.0)
        active_workers = Value(
            "i", int(False)
        )  # Tracks whether the task has any active distributed workers
```

Maybe overthinking and I don't feel strongly about it 😄

ingestion_server/ingestion_server/api.py
stacimc: Agreed! Done

ingestion_server/ingestion_server/indexer.py
stacimc: Update -- done. The result is that:
* if workers fail, we do not attempt to `go_live`. When you query the refresh status it will correctly report that it is no longer active, and that there are errors
* if workers succeed, we do not consider the task inactive until `go_live` is complete

ingestion_server/ingestion_server/api.py
AetherUnbound: Ha, that's fair! I've corrected a lot of these cases in the past where it wasn't necessary, but I agree - here it makes sense to keep it the way you have it :slightly_smiling_face:

ingestion_server/ingestion_server/indexer.py
AetherUnbound: That sounds perfect!

ingestion_server/ingestion_server/state.py
AetherUnbound: > This definitely feels a bit shaky around the error states and I would very much prefer to get this into Airflow for better tracking sweat I think this should at least get us to a better place than we are currently.

I'm so with you :sweat_smile: This just needs to be "good enough" to work on its own _most of the time_ until we can provide better support for it in Airflow

ingestion_server/ingestion_server/state.py
```diff
@@ -27,6 +28,11 @@ class WorkerStatus(enum.Enum):
     ERROR = 2
 
 
+TaskData = namedtuple(
+    "TaskData", ["target_index", "task_id", "percent_successful", "percent_completed"]
+)
```
AetherUnbound, line 33: I have to admit, I like [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) a bit more just because we can have types associated with it as well!
```suggestion
class TaskData(NamedTuple):
    target_index: int
    task_id: str
    percent_successful: float
    percent_completed: float
```

AetherUnbound (approved):
This looks great! Can't wait to have this all report more accurately :sob: :raised_hands: One note on named tuples, but otherwise this is good to go!

ingestion_server/ingestion_server/state.py
stacimc: Sounds good!

zackkrida (approved):
This felt really easy to understand, and there's so many nice features. The slack notification for worker errors will be so helpful (although hopefully we never have to see them 😆).

@dhruvkb you're welcome to look at this too, but I think this looks good.

PR #627, opened 2022-04-05
Reduce flakiness of `just init`

dhruvkb:
## Description

This PR adds steps to the CI+CD workflow that collect the Docker Compose logs and upload them as artefacts so that we can better debug failing jobs.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -137,21 +137,6 @@ jobs:
           name: docs
           path: api/build/html/
 
-      - name: Write logs
-        if: always()
-        run: |
-          mkdir logs/
-          just write-logs web logs/web.log
-          just write-logs ingestion_server logs/ingestion_server.log
-          just write-logs indexer_worker logs/indexer_worker.log
-
-      - name: Upload logs
-        if: always()
-        uses: actions/upload-artifact@v2
-        with:
-          name: logs
-          path: logs/
-
   check-migrations:
     name: Check for uncommited Django migrations
     runs-on: ubuntu-latest
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -137,6 +137,21 @@ jobs:
           name: docs
           path: api/build/html/
 
+      - name: Write logs
+        if: always()
+        run: |
+          mkdir logs/
+          just write-logs web logs/web.log
+          just write-logs ingestion_server logs/ingestion_server.log
+          just write-logs indexer_worker logs/indexer_worker.log
+
+      - name: Upload logs
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: logs
+          path: logs/
+
   check-migrations:
     name: Check for uncommited Django migrations
     runs-on: ubuntu-latest
--- a/ingestion_server/ingestion_server/ingest.py
+++ b/ingestion_server/ingestion_server/ingest.py
@@ -259,12 +259,13 @@ def reload_upstream(
     process involves the following steps.
 
     1. Get the list of overlapping columns: ``_get_shared_cols``
-    2. Create FDW for the data transfer: ``get_fdw_query``
-    3. Import data into a temporary table: ``get_copy_data_query``
-    4. Clean the data: ``clean_image_data``
-    5. Recreate indices from the original table: ``_generate_indices``
-    6. Recreate constraints from the original table: ``_generate_constraints``
-    7. Promote the temp table and delete the original: ``get_go_live_query``
+    2. Create the FDW extension if it does not exist
+    3. Create FDW for the data transfer: ``get_fdw_query``
+    4. Import data into a temporary table: ``get_copy_data_query``
+    5. Clean the data: ``clean_image_data``
+    6. Recreate indices from the original table: ``_generate_indices``
+    7. Recreate constraints from the original table: ``_generate_constraints``
+    8. Promote the temp table and delete the original: ``get_go_live_query``
 
     This is the main function of this module.
 
@@ -288,9 +289,17 @@ def reload_upstream(
     shared_cols = _get_shared_cols(downstream_db, upstream_db, table)
     upstream_db.close()
 
-    with downstream_db.cursor() as downstream_cur:
-        # Step 2: Create FDW for the data transfer
+    with downstream_db, downstream_db.cursor() as downstream_cur:
+        # Step 2: Create the FDW extension if it does not exist
         log.info("(Re)initializing foreign data wrapper")
+        create_fdw = "CREATE EXTENSION IF NOT EXISTS postgres_fdw;"
+        try:
+            downstream_cur.execute(create_fdw)
+        except psycopg2.errors.UniqueViolation:
+            log.warning("Extension already exists, possible race condition.")
+
+    with downstream_db, downstream_db.cursor() as downstream_cur:
+        # Step 3: Create FDW for the data transfer
         init_fdw = get_fdw_query(
             RELATIVE_UPSTREAM_DB_HOST,
             RELATIVE_UPSTREAM_DB_PORT,
@@ -301,7 +310,7 @@ def reload_upstream(
         )
         downstream_cur.execute(init_fdw)
 
-        # Step 3: Import data into a temporary table
+        # Step 4: Import data into a temporary table
         log.info("Copying upstream data...")
         environment = config("ENVIRONMENT", default="local").lower()
         limit_default = 100_000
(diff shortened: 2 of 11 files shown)


sarayourfriend:
@dhruvkb I think this is a duplicate of #569 :slightly_smiling_face:

dhruvkb:
@sarayourfriend while the changes for logging are identical, I'm actually trying to get to the bottom of issues #562.

dhruvkb:
This is the error message that was observed:
```
ingestion_server_1  | Process Task-4:
ingestion_server_1  | Traceback (most recent call last):
ingestion_server_1  |   File "/usr/local/lib/python3.10/multiprocessing/process.py", line 315, in _bootstrap
ingestion_server_1  |     self.run()
ingestion_server_1  |   File "/ingestion_server/ingestion_server/tasks.py", line 113, in run
ingestion_server_1  |     reload_upstream(self.model)
ingestion_server_1  |   File "/ingestion_server/ingestion_server/ingest.py", line 302, in reload_upstream
ingestion_server_1  |     downstream_cur.execute(init_fdw)
ingestion_server_1  | psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "pg_extension_name_index"
ingestion_server_1  | DETAIL:  Key (extname)=(postgres_fdw) already exists.
```

Digging more, I realised the cause. The ingestion server does lots of things in a single transaction. If the ingestion for audio starts before the ingestion for images is complete, it cannot see the extension and run `CREATE EXTENSION` again. When the image transaction is completed the extension exists. When the audio transaction comes to completion the query fails as it tries to create an extension that exists. By splitting the process into two transactions we allow one to complete quickly preventing two identical calls.

While I cannot guarantee this fixes #562, it certainly should make it much less flaky.

.github/workflows/ci_cd.yml
```diff
@@ -137,6 +137,21 @@ jobs:
           name: docs
           path: api/build/html/
 
+      - name: Write logs
+        if: always()
```
zackkrida, line 141: Is there a difference between `if: always()` and not having any conditional on these steps? I think there's something obvious here I'm missing, hah.

dhruvkb: Steps with `if` `always()` runs even if a previous step has failed, sort of like a `finally` in a `try`-`catch` block.

zackkrida: Oh, that's really helpful—thanks!

ingestion_server/ingestion_server/ingest.py
```diff
@@ -288,9 +289,17 @@ def reload_upstream(
...
+        # Step 2: Create the FDW extension if it does not exist
         log.info("(Re)initializing foreign data wrapper")
+        create_fdw = "CREATE EXTENSION IF NOT EXISTS postgres_fdw;"
+        try:
+            downstream_cur.execute(create_fdw)
+        except psycopg2.errors.UniqueViolation:
+            log.warning("Extension already exists, possible race condition.")
+
+    with downstream_db, downstream_db.cursor() as downstream_cur:
+        # Step 3: Create FDW for the data transfer
```
AetherUnbound, line 302: Do you think we could just run `downstream_db.commit()` rather than recreating the cursor context on L301?

ingestion_server/ingestion_server/ingest.py
```diff
@@ -288,9 +289,17 @@ def reload_upstream(
...
-    with downstream_db.cursor() as downstream_cur:
-        # Step 2: Create FDW for the data transfer
+    with downstream_db, downstream_db.cursor() as downstream_cur:
+        # Step 2: Create the FDW extension if it does not exist
         log.info("(Re)initializing foreign data wrapper")
+        create_fdw = "CREATE EXTENSION IF NOT EXISTS postgres_fdw;"
+        try:
+            downstream_cur.execute(create_fdw)
+        except psycopg2.errors.UniqueViolation:
+            log.warning("Extension already exists, possible race condition.")
```
AetherUnbound, line 299: Do you think this logic should exist in `queries.py` for consistency?

AetherUnbound (commented):
What a find! I hadn't seen that error before, but I really hope this does it :crossed_fingers:

ingestion_server/ingestion_server/ingest.py
dhruvkb: Using cursor as a context manager automatically makes a transaction of the entire block. Is it expesive to instantiate cursors?

ingestion_server/ingestion_server/ingest.py
dhruvkb: That seems like a good idea, moved in 166d47c.

sarayourfriend:
Okay @dhruvkb do you want me to just close my PR then? I don't want to waste time working on it if your solution here will just replace it anyway. There appear to be differences in how the logging is being uploaded in this PR vs mine.

If so, then please also remove the local logging output from the `run_test` scripts and feel free to close my PR.

dhruvkb:
@sarayourfriend no, no! I'll remove the logging bit from this PR, please go ahead and merge yours (mine is a very rudimentary one anyway that has served its purpose.

ingestion_server/ingestion_server/ingest.py
AetherUnbound: Creating a new connection and cursor does come with some overhead, mostly on the Postgres end. Each time a connection is made, [Postgres spins up a new process for that connection](https://www.postgresql.org/docs/13/connect-estab.html). For queries like this that will quickly follow one another, it seems prudent to `commit` with the existing cursor rather than tear down everything and reconstruct it for the second query (IMO). In previous applications I've worked on, we've had long-running processes that issue hundreds of queries open a single connection and create cursors when necessary from there.

ingestion_server/ingestion_server/queries.py
```diff
@@ -33,6 +33,16 @@ def get_existence_queries(table):
...
 
 
+def get_create_ext_query():
+    """
+    Get the query for creating the ``postgres_fdw`` extension, if it does not exist.
+
+    :return: the SQL query for creating the FDW extension
+    """
+
+    return "CREATE EXTENSION IF NOT EXISTS postgres_fdw;"
```
AetherUnbound, line 43: Nit: Sorry, I hate to be a nuisance about this :sweat_smile: It seems like every other function in this module returns a `psycopg2.sql.SQL` object, perhaps this should too? On a personal note I think the string is totally fine, that just seems consistent with the rest of the code :woman_shrugging:

ingestion_server/ingestion_server/ingest.py
```diff
@@ -288,9 +290,17 @@ def reload_upstream(
...
     upstream_db.close()
 
-    with downstream_db.cursor() as downstream_cur:
-        # Step 2: Create FDW for the data transfer
+    with downstream_db, downstream_db.cursor() as downstream_cur:
+        # Step 2: Create the FDW extension if it does not exist
         log.info("(Re)initializing foreign data wrapper")
+        try:
+            create_ext = get_create_ext_query()
+            downstream_cur.execute(create_ext)
```
zackkrida, line 298: It's interesting that this throws an exception. In postgres the `IF NOT EXISTS` clause is used to issue a notice instead of an error, and notices don't abort the current transaction so they're not _really_ exceptions. I guess it makes sense to throw on _any_ message from postgres so it can be handled in user code.

zackkrida (approved):
It's scary to touch the ingestion server, but I think this looks good!

ingestion_server/ingestion_server/ingest.py
dhruvkb: The `IF NOT EXIST` part works fine, but in simultaneous threads, two threads can see that the extension does not exist and both try to create it, leading to an integrity error with a repetition in a primary key. So a unique violation is raised.

ingestion_server/ingestion_server/queries.py
dhruvkb: I'm a sucker for consistency, and now that you've pointed it out, I can't unsee this. Updated in 5bac2d8.

ingestion_server/ingestion_server/ingest.py
dhruvkb: We're not reconnecting to Postgres, we're reusing the connection created in L282 to just create new cursors, representing individual transactions.

The singular connection created is terminated at the end of the function in L367.

AetherUnbound: Ah, apologies! I thought that the `with downstream_db` was reopening the connection. This is good as is!

AetherUnbound (approved):
Looks great, really hope this fixes the issue!! :crossed_fingers:

PR #630, opened 2022-04-06
Improve the thumbnail service to support compression and WEBP

### LINKED ISSUE #531: Image thumbnail improvements (dimensions, compression)
## Problem

Currently on the frontend, there are two optimizations we could make with thumbnails:
- Thumbnails are 600px wide at their largest dimension; On the frontend thumbnails are _rarely_ larger than 300px. Display size _is_ a factor here, but one I need to test more.
- Thumbnails could benefit from some lossy compression. Using tinypng.com I was able to give this thumbnail:

https://api.openverse.engineering/v1/images/3fa11800-4907-43d0-9921-52ed5ddb22aa/thumb/

a 56% size reduction with minimal obvious visual differences:

image

> GitHub _may apply compression_, I'm not sure, so you might want to run this test yourself.

|before|after|
|--|--|
|dog-thumb|dog-thumb|

## Description

While images are pretty "cheap" for browsers, compared to something like JS, because they don't need to be parsed, _and_ 
we're employing lazy-loading in Openverse, this would not really effect the _percieved_ performance of the site. Or if it did, it would likely de negligible. However, it would significantly improve the number of real bytes transferred to users, and that's always a wonderful thing.

This could be a good opportunity to give our thumbnail cache some basic sizing options, because then we could _also_ use it on the single result pages, instead of hotlinking to our 3rd party providers.

dhruvkb:
## Fixes

Fixes #531 by @zackkrida 

## Description

This PR replaces [willnorris/imageproxy](https://github.com/willnorris/imageproxy) with [h2non/imaginary](https://github.com/h2non/imaginary), a faster, more feature-rich and more highly-rated library for image manipulations over HTTP. This library allows us to compress thumbnails and reduce bytes.

Now, we can control the dimensions of the thumbnail and the amount of compression using env vars (while also allowing the URL query params `full_size` and `compressed` override the defaults if needed).

To spark joy, this PR adds API documentation for the thumbnail endpoints (both audio and video) with all query params documented.

## Testing Instructions

0. Check out the PR and fetch thumbnails for some audio/image files. They should load normally.
1. Play with the `full_size` and `compressed` query params.
  - With `full_size` set to `'true'`, the image should be larger.
  - With `full_size` set to `'false'`, the image will be smaller.
  - Compression is by default set to the opposite of `full_size`, but can be overridden with the `compressed` param.

## Stats

Here are some randomly sampled cases.

| Full-size no-compression | Full-size compression | Thumbnail no-compression | Thumbnail compression |
|-|-|-|-|
| 307 kB | 84.1 kB | 223 kB | 30.6 kB |
| 385 kB | 96.1 kB | 279 kB | 39.9 kB |
| 387 kB | 108 kB | 278 kB | 42.9 kB |

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
This is such a great find Dhruv! I'm excited to review this tomorrow :slightly_smiling_face:

dhruvkb:
Looking forward to your reviews!

api/catalog/api/views/media_views.py
```diff
@@ -143,24 +152,50 @@ def _get_user_ip(request):
...
-            proxy_upstream = (
-                f"{settings.THUMBNAIL_PROXY_URL}/"
-                f"{settings.THUMBNAIL_WIDTH_PX},fit/"
-                f"{image_url}"
-            )
+    def _thumbnail_proxy_comm(path: str, params: dict):
+        proxy_url = settings.THUMBNAIL_PROXY_URL
+        query_string = urlencode(params)
+        upstream_url = f"{proxy_url}/{path}?{query_string}"
+        log.info(f"Upstream URL: {upstream_url}")
```
AetherUnbound, line 159: Same here:
```suggestion
        log.debug(f"Upstream URL: {upstream_url}")
```

I'll stop with the suggestion, but I think most if not all of the `log.info` calls should be at the `debug` level. Otherwise we're going to get a TON of log messages per request.

api/catalog/api/views/media_views.py
```diff
@@ -143,24 +152,50 @@ def _get_user_ip(request):
...
+        info_res, *_ = MediaViewSet._thumbnail_proxy_comm("info", {"url": image_url})
+        info = json.loads(info_res.read())
+
+        path = "resize"
+        params = {
+            "url": image_url,
+            "width": info["width"] if is_full_size else settings.THUMBNAIL_WIDTH_PX,
+        }
+        if is_compressed:
+            params |= {
```
AetherUnbound, line 189: NICE, I forget you can do that rather than `{**first, **second}`!

AetherUnbound (commented):
This looks great! It sees quite fast too, which is rad. There aren't any pngs in the sample data to test with, but the JPEGs all look good 🙂  I have a few comments about logging

zackkrida:
I'll test this soon, just want to note that we make sure to manually visually diff a few thumbnails against their current, production counterparts, to check for noteworthy quality differences or any other issues.

The new library looks great! It would be an excellent enhancement to figure out how to serve webp images in the future.

dhruvkb:
@zackkrida the `h2non/imaginary` service is capable of converting images sent to it to webp as well, if we send a `type` param in the request. I'll run some tests with that and see if that can further cut down the size.

api/catalog/api/docs/audio_docs.py
```diff
@@ -206,3 +208,17 @@ class AudioComplain(MediaComplain):
...
         "code_examples": code_examples,
     }
+
+
+class AudioThumbnail:
+    desc = f"""
+thumbnail is an API endpoint to retrieve the scaled down and compressed thumbnail
+of the artwork of an audio track or its audio set.
+
+{refer_sample}"""  # noqa
```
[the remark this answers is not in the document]
dhruvkb: That's interesting! I removed it and `flake8` doesn't actually complain at all. Maybe these we're added in old code and never removed after refactoring. I'll remove the other occurrences separately to avoid cluttering this PR.

api/catalog/api/views/media_views.py
dhruvkb: Makes sense, I've combined the log statements and changed to `debug` in c76f1f2.

api/catalog/api/serializers/media_serializers.py
```diff
@@ -420,3 +421,32 @@ class MediaSearchSerializer(serializers.Serializer):
...
+        source="is_compressed",
+        allow_null=True,
+        default=None,
+        required=False,
+        help_text="whether to compress the output image to reduce file size,"
+        "defaults to opposite of `full_size`",
+    )
+
+    def validate(self, data):
+        log.info(f"MediaThumbnailRequestSerializer data: {data}")
```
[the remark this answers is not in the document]
dhruvkb: Done in c76f1f2.

api/catalog/api/docs/audio_docs.py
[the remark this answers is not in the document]
dhruvkb: Removed `#noqa` in f8a0204. Created #635 for the rest.

dhruvkb:
Some more stats:

| Thumnail compressed | Thumbnail compressed WEBP |
|-|-|
| 30.5 kB | 19.2 kB |
| 29.0 kB | 16.9 kB |
| 36.4 kB | 22.6 kB |

With compression _and_ WEBP, we can reduce the size even more than the already compressed thumbnails.

api/catalog/api/views/media_views.py
```diff
@@ -143,24 +157,69 @@ def _get_user_ip(request):
...
+                "compression": settings.THUMBNAIL_PNG_COMPRESSION,
+            }
+        else:
+            params |= {
+                "quality": 100,
+                "compression": 0,
+            }
+
+        if "webp" in accept_header:
+            params["type"] = "auto"  # Use ``Accept`` header to determine output type.
```
dhruvkb, line 217: This converts JPG and PNG to WEBP if the `Accept` header contains it. Otherwise the format stays as it is. This is to prevent issues with Safari which before Big Sur, [preferentially requests PNG](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values). JPG images become quite large if converted to PNG.

docker-compose.yml
```diff
@@ -11,11 +11,14 @@ services:
     volumes:
       - api-postgres:/var/lib/postgresql/data
 
-  thumbs:
-    image: willnorris/imageproxy
+  thumbnails:
+    image: h2non/imaginary:latest
```
sarayourfriend, line 15: Should we pin this to a specific version, like we do the rest of our dependency images? I know the original one wasn't but it seems wrong not to.

api/catalog/api/views/media_views.py
```diff
@@ -143,24 +157,69 @@ def _get_user_ip(request):
...
-                f"{settings.THUMBNAIL_PROXY_URL}/"
-                f"{settings.THUMBNAIL_WIDTH_PX},fit/"
-                f"{image_url}"
-            )
+    def _thumbnail_proxy_comm(
+        path: str,
+        params: dict,
+        headers: List[
+            tuple[str, str]
+        ] = None,  # ``List`` because there is a ``list`` function
```
sarayourfriend, line 165: Setting it to `None` by default would require wrapping this in `Optional` for the type to be accurate, no?

Could just default to an empty list, alternatively. The existing code will work exactly as is, although the `if headers` is a little redundant considering iteration over an empty list is a no-op anyway and doesn't need to be guarded anymore if we don't use the `None` default.

api/catalog/api/views/media_views.py
```diff
@@ -143,24 +157,69 @@ def _get_user_ip(request):
...
+            res_status = upstream_response.status
             content_type = upstream_response.headers.get("Content-Type")
+            log.debug(
+                "Image proxy response "
+                f"status: {res_status}, content-type: {content_type}"
+            )
+
+            return upstream_response, res_status, content_type
         except HTTPError:
             raise get_api_exception("Failed to render thumbnail.")
```
sarayourfriend, line 188: Should we forward the error details from the proxy service?

api/catalog/api/views/media_views.py
```diff
@@ -143,24 +157,69 @@ def _get_user_ip(request):
...
+        is_full_size: bool = False,
+        is_compressed: bool = True,
+    ):
+        info_res, *_ = MediaViewSet._thumbnail_proxy_comm("info", {"url": image_url})
+        info = json.loads(info_res.read())
+
+        params = {
+            "url": image_url,
+            "width": info["width"] if is_full_size else settings.THUMBNAIL_WIDTH_PX,
+        }
```
sarayourfriend, line 203: We can skip the `info` request altogether if `is_full_size` is false:

```suggestion
        width = settings.THUMBNAIL_WIDTH_PX
        if is_full_size:
            info_res, *_ = MediaViewSet._thumbnail_proxy_comm("info", {"url": image_url})
            info = json.loads(info_res.read())
            width = info["width"]

        params = {
            "url": image_url,
            "width": width,
        }
```

api/catalog/api/views/media_views.py
```diff
@@ -124,6 +128,16 @@ def report(self, request, *_, **__):
         serializer = self.get_serializer(report)
         return Response(data=serializer.data, status=status.HTTP_201_CREATED)
 
+    def thumbnail(self, image_url, request, *_, **__):
+        serializer = self.get_serializer(data=request.query_params)
+        if not serializer.is_valid():
+            raise get_api_exception("Invalid input.", 400)
```
sarayourfriend, line 134: Instead of raising our own exception (which will have fewer details) can we just pass `raise_exception` to `is_valid`?

```suggestion
        serializer.is_valid(raise_exception=True)
```

sarayourfriend (changes requested):
Looks really good! Just a few things to clean up/short-cut requests. This is really awesome though.

docker-compose.yml
dhruvkb: I've been waiting on a response for https://github.com/h2non/imaginary/issues/390 to resolve this.

api/catalog/api/views/media_views.py
dhruvkb: Raise exception from serialiser instead of custom

api/catalog/api/views/media_views.py
dhruvkb: Linters generally complain if we use a mutable such as [] as the default value of an argument (for good reason). It might just be better to make the type Optional.

sarayourfriend: Ahh, of course! I'd seen that linting error before but never ingrained why it was bad, it turns out Python re-uses the default array instance between invocations:

```
>>> def arr_default(a = []):
...     a.append(1)
...     return a
... 
>>> a = arr_default()
>>> a
[1]
>>> arr_default()
[1, 1]
>>> a
[1, 1]
```

docker-compose.yml
sarayourfriend: Wow, the last release on GitHub was also 2 years ago :confused:

sarayourfriend: I guess the worst case here is we somehow build it ourselves in a fork?

sarayourfriend:
When I try to access `http://localhost:8000/v1/images/2f3f8563-a52c-473e-ae0b-aac8c9d81c05/thumb/?is_full_size=True` locally I always get the `600px` width result. I tried `is_full_size=true`, `is_full_size` and `full_size` (and variations). Am I misunderstanding how this is meant to work?

api/test/media_integration.py
```diff
@@ -105,6 +105,28 @@ def thumb(fixture):
...
 
+def thumb_compression(fixture):
+    thumbnail_url = fixture["results"][0]["thumbnail"]
+
+    thumbnail_response = requests.get(thumbnail_url)
+    compressed_size = len(thumbnail_response.content)
+    thumbnail_response = requests.get(f"{thumbnail_url}?compressed=no")
+    actual_size = len(thumbnail_response.content)
+
+    assert compressed_size < actual_size
```
sarayourfriend, line 116: Can you add a complimentary test for the `full_size` parameter?

sarayourfriend (commented):
LGTM! Tested locally and it works great.

Along with the `full_size` test, can you add unit tests for the serializer and the `validate` method specifically?

api/catalog/api/views/media_views.py
dhruvkb: As long as you don't mutate the mutable inside the function, there's technically no issue with doing this. But considering that it's a potential pit that's easy to fall into, best to avoid it altogether and use `None`.

sarayourfriend: It'd be nice if Python had a sort of "readonly list" for times like these... Oh wait! A tuple? :grin:

dhruvkb: Tuples aren't readonly.

dhruvkb: Yeah, I was wrong. So I immediately deleted it 😆 . It's certainly feasible to use it instead of an array as the default argument. It would certainly be anti-conventional but there are not _technical_ problems with doing so.

api/catalog/api/views/media_views.py
dhruvkb: Makes sense, fixed in  bb3d89c.

api/catalog/api/views/media_views.py
dhruvkb: Updated code to use tuple in ff1c7df.

sarayourfriend: Ah whoops, I didn't see the comment disappear, sorry!

I wonder if it would even be anti-conventional. If you have access to GitHub codesearch, there's a lot of results of it being used as the default for an iterable argument: https://cs.github.com/?scopeName=All+repos&scope=&q=language%3Apython+%22%3D+tuple%28%29%2C%22

sarayourfriend (approved):
LGTM, just curious about the release issue, whether we'll need to build an image ourselves from the latest code?

api/test/media_integration.py
dhruvkb: Added tests in 3c23121.

api/test/media_integration.py
```diff
@@ -127,6 +129,22 @@ def thumb_webp(fixture):
...
+        return image.size
+
+    thumbnail_url = fixture["results"][0]["thumbnail"]
+    full_w, full_h = _get_image_dimen(f"{thumbnail_url}?full_size=yes")
+    scaled_w, scaled_h = _get_image_dimen(thumbnail_url)
+    if full_w > 600:
+        assert scaled_w == 600
+        assert full_w > scaled_w
+    else:
+        assert scaled_w == full_w  # h2non/imaginary will not scale up
```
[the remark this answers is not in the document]
sarayourfriend: Hmm... This seems like solving the flakiness by hiding it rather than actually making the test _do_ the testing every time though.

Should we choose specific fixture data that we know will test the full size operation correctly? We wouldn't write a unit test using data we know won't actually exercise the unit we want to test, after all. Are integration tests different?

api/catalog/api/views/media_views.py
AetherUnbound: FWIW that link does not work for me 😅 seems I have to sign up for the service this corresponds to. If we wanted to make it generic for both lists and tuples we could use a [`typing.Iterable`](https://docs.python.org/3/library/typing.html#typing.Iterable) type instead. But having it explicitly as a tuple is also good 🙂

docker-compose.yml
dhruvkb: Attempted this in PR #641.

dhruvkb:
Just noting that the process of building and hosting a custom image is being worked out in a separate PR, #641.

api/test/media_integration.py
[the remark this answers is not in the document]
dhruvkb: It does at least test that the image returned using `full_screen` is either larger (when the source material > 600px) or the same size (when the source material is 600px or lower).

PR #648, opened 2022-04-19
Use `::` over `cast` in audioset_view

### LINKED ISSUE #467: Use `::` over `cast` in audioset_view
## Problem

We have a few different forms of the `audioset_view` definition (see WordPress/openverse-catalog#320).

## Description

These should be unified so all definitions are the same, even if [`cast` and `::` serve the same purpose](https://www.postgresql.org/docs/13/sql-expressions.html#SQL-SYNTAX-TYPE-CASTS). That syntax also already exists in the [ingestion server mock schema DDL](https://github.com/WordPress/openverse-api/blob/a7aa53ffc5f9d3b8b6bebf9dbacba248623a5a61/ingestion_server/test/mock_schemas/audioset_view.sql#L22), so we can apply that to the `load_sample_data.sh` script.

## Implementation

parthpatel9414:
Fixes #467

## Fixes

Fixes #[648] by @AetherUnbound 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -80,15 +80,15 @@ docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" /bin/bash -c "psql -U deploy 
 	CREATE VIEW audioset_view
 	AS
 		SELECT DISTINCT
-			audio_set ->> 'foreign_identifier'  :: varchar(1000) as foreign_identifier,
-			audio_set ->> 'title'              :: varchar(2000)) as title,
-			cast(audio_set ->> 'foreign_landing_url' as varchar(1000)) as foreign_landing_url,
-			cast(audio_set ->> 'creator'             as varchar(2000)) as creator,
-			cast(audio_set ->> 'creator_url'         as varchar(2000)) as creator_url,
-			cast(audio_set ->> 'url'                 as varchar(1000)) as url,
-			cast(audio_set ->> 'filesize'            as integer)       as filesize,
-			cast(audio_set ->> 'filetype'            as varchar(80))   as filetype,
-			cast(audio_set ->> 'thumbnail'           as varchar(1000)) as thumbnail,
+			audio_set ->> 'foreign_identifier'   :: varchar(1000) as foreign_identifier,
+			audio_set ->> 'title'               :: varchar(2000) as title,
+			audio_set ->> 'foreign_landing_url' :: varchar(1000) as foreign_landing_url,
+			audio_set ->> 'creator'             :: varchar(2000) as creator,
+			audio_set ->> 'creator_url'         :: varchar(2000) as creator_url,
+			audio_set ->> 'url'                 :: varchar(1000) as url,
+			audio_set ->> 'filesize'             :: integer       as filesize,
+			audio_set ->> 'filetype'             :: varchar(80)   as filetype,
+			audio_set ->> 'thumbnail'           :: varchar(1000) as thumbnail,
 			provider
 		FROM audio_view
 		WHERE audio_set IS NOT NULL;
--- a/load_sample_data.sh
+++ b/load_sample_data.sh
@@ -81,7 +81,7 @@ docker-compose exec -T "$UPSTREAM_DB_SERVICE_NAME" /bin/bash -c "psql -U deploy 
 	AS
 		SELECT DISTINCT
 			audio_set ->> 'foreign_identifier'  :: varchar(1000) as foreign_identifier,
-			cast(audio_set ->> 'title'               as varchar(2000)) as title,
+			audio_set ->> 'title'              :: varchar(2000)) as title,
 			cast(audio_set ->> 'foreign_landing_url' as varchar(1000)) as foreign_landing_url,
 			cast(audio_set ->> 'creator'             as varchar(2000)) as creator,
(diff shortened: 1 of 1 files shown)


AetherUnbound:
Thanks for getting this work started @parthpatel9414! While the error messages aren't super clear, I am seeing this [in the logs](https://github.com/WordPress/openverse-api/runs/6151021640?check_suite_focus=true):
```
DROP VIEW
NOTICE:  view "audioset_view" does not exist, skipping
ERROR:  type "foreign_identifier" does not exist
LINE 4: ...et ->> 'foreign_identifier'   :: varchar(1000) :: foreign_id...
                                                             ^
```

I think there are a couple of things going on here:
1. The `::` syntax can only be used as a replacement for `cast(x as y)` (e.g. `x::y`). We can't use it to replace an *alias*, which is the second `as` on this line. This is discussed in the PostgreSQL documentation

    > By default, a cast can be invoked only by an explicit cast request, that is an explicit CAST(x AS typename) or x::typename construct. _[Source](https://www.postgresql.org/docs/current/sql-createcast.html)_

    And

    > Just as in a table, every output column of a SELECT has a name. In a simple SELECT this name is just used to label the column for display, but when the SELECT is a sub-query of a larger query, the name is seen by the larger query as the column name of the virtual table produced by the sub-query. To specify the name to use for an output column, write AS output_name after the column's expression. (You can omit AS, but only if the desired output name does not match any PostgreSQL keyword (see [Appendix C](https://www.postgresql.org/docs/current/sql-keywords-appendix.html)). _[Source](https://www.postgresql.org/docs/current/sql-select.html)_

2. Another factor in play is the order of operations. When I tested this locally after changing the second `::` to `as`, I got this error:

    ```
    ERROR:  invalid input syntax for type integer: "filesize"
    LINE 10: audio_set ->> 'filesize'             :: integer       as fil...
                                                     ^
    ```

    It looks like this can be resolved by wrapping the JSON access in parenthesis: `(audio_set ->> 'filesize')::integer as filesize`. I assume this error arises because Postgres is trying to convert the string `'filesize'` as an integer rather than casting the field.


Hope this all makes sense, let us know if you have any questions or run into any difficulties!

parthpatel9414:
> Thanks for getting this work started @parthpatel9414! While the error messages aren't super clear, I am seeing this [in the logs](https://github.com/WordPress/openverse-api/runs/6151021640?check_suite_focus=true):
> 
> ```
> DROP VIEW
> NOTICE:  view "audioset_view" does not exist, skipping
> ERROR:  type "foreign_identifier" does not exist
> LINE 4: ...et ->> 'foreign_identifier'   :: varchar(1000) :: foreign_id...
>                                                              ^
> ```
> 
> I think there are a couple of things going on here:
> 
>     1. The `::` syntax can only be used as a replacement for `cast(x as y)` (e.g. `x::y`). We can't use it to replace an _alias_, which is the second `as` on this line. This is discussed in the PostgreSQL documentation
>        > By default, a cast can be invoked only by an explicit cast request, that is an explicit CAST(x AS typename) or x::typename construct. _[Source](https://www.postgresql.org/docs/current/sql-createcast.html)_
>        
>        
>        And
>        > Just as in a table, every output column of a SELECT has a name. In a simple SELECT this name is just used to label the column for display, but when the SELECT is a sub-query of a larger query, the name is seen by the larger query as the column name of the virtual table produced by the sub-query. To specify the name to use for an output column, write AS output_name after the column's expression. (You can omit AS, but only if the desired output name does not match any PostgreSQL keyword (see [Appendix C](https://www.postgresql.org/docs/current/sql-keywords-appendix.html)). _[Source](https://www.postgresql.org/docs/current/sql-select.html)_
> 
>     2. Another factor in play is the order of operations. When I tested this locally after changing the second `::` to `as`, I got this error:
>        ```
>        ERROR:  invalid input syntax for type integer: "filesize"
>        LINE 10: audio_set ->> 'filesize'             :: integer       as fil...
>                                                         ^
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        It looks like this can be resolved by wrapping the JSON access in parenthesis: `(audio_set ->> 'filesize')::integer as filesize`. I assume this error arises because Postgres is trying to convert the string `'filesize'` as an integer rather than casting the field.
> 
> 
> Hope this all makes sense, let us know if you have any questions or run into any difficulties!

Hello @AetherUnbound ,
Thanks for the explanation and sorry for being late to get back to you.
I try and complete the issue by end of this week. 

Thank you again.

AetherUnbound:
Hey @parthpatel9414, just checking in! Do you still have time to tackle this work? We can try and get it across the finish line if not 🙂

AetherUnbound:
Hmm, it looks like #693 unfortunately didn't prevent the docs preview workflow from running @dhruvkb 😞

obulat (approved):
It all looks well, @parthpatel9414. Thank you for your contribution!

stacimc (approved):
Looks good! My understanding is the docs preview is not an issue and that it shouldn't have run here, so this should be okay to merge 🤔

PR #661, opened 2022-04-29
Fix `length` query for audio

### LINKED ISSUE #652: Duration filter is inaccurate
## Description

We have duration filters defined in our Elasticsearch models:

https://github.com/WordPress/openverse-api/blob/a7955c86d43bff504e8d41454f68717d79dd3a44/ingestion_server/ingestion_server/elasticsearch_models.py#L307-L309

Unfortunately, duration does not seem to be accurate (these results are several hours long but appear in all available filters):
* https://api.openverse.engineering/v1/audio/?q=deuteronomy&duration=short
* https://api.openverse.engineering/v1/audio/?q=deuteronomy&duration=long

It seems like this may be due to the nature of the `get_duration` function on the model:

https://github.com/WordPress/openverse-api/blob/a7955c86d43bff504e8d41454f68717d79dd3a44/ingestion_server/ingestion_server/elasticsearch_models.py#L340

It may also be due to an order of magnitude mismatch (e.g. expecting ms when it's actually seconds).

## Resolution

### LINKED ISSUE #653: Medium duration is unavailable
## Description

We have a `medium` duration defined in Elasticsearch: 

https://github.com/WordPress/openverse-api/blob/a7955c86d43bff504e8d41454f68717d79dd3a44/ingestion_server/ingestion_server/elasticsearch_models.py#L308

This value is not specified in the list of valid durations, and will thus fail when used in a query: 

https://github.com/WordPress/openverse-api/blob/fbca822683f4b18a692ac34291d4642309946302/api/catalog/api/constants/field_values.py#L28-L31

## Reproduction

1. Visit https://api.openverse.engineering/v1/audio/?q=whale&duration=medium
2. Observe the following errror

```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": {
        "duration": [
            "Invalid duration: medium. Available options: {'short', 'long'}"
        ]
    }
}
```

## Additional context

Related to #652

## Resolution

obulat:
## Fixes

Fixes #652 by @AetherUnbound
Fixes #653 by @AetherUnbound

## Description

Audio already had a duration field that referred to the duration in milliseconds. Based on the discussion with @dhruvkb, I converted the `duration` Enum field to  a `length` field to be able to filter the search results by 'duration categories' (short, medium, long). I also added the `medium` length to the constants so that it's available as an option.

To not make this a breaking change, I added `duration` as a possible filter query parameter that falls back to `length`. However, since  the `duration` parameter never worked before, this might not be necessary. What do you think?

## Testing Instructions

Run the API using `just start` and `just init`. After the indexes are ready, query for some sound using `length=<short|medium|long>` parameter, eg. `http://localhost:8000/v1/audio/?q=vocal&length=medium,short`. You should see the query working. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/ingestion_server/es_mapping.py
```diff
@@ -137,6 +137,7 @@ def index_settings(table_name):
             "sample_rate": {"type": "integer"},
             "genres": {"fields": {"keyword": {"type": "keyword"}}, "type": "text"},
             "duration": {"type": "integer"},
+            "length": {"fields": {"keyword": {"type": "keyword"}}, "type": "text"},
```
sarayourfriend, line 140: My confusion comes down to this: it seems like we were previously putting a string (from `get_duration`) into an integer field in ES. Did ES not complain about that before? Or am I misunderstanding the code?

I'm also not personally a fan of having length and duration in the index, especially if `length` was getting mapped to the string field before. It'll be pretty confusing.

I think we should "live with" the current name "duration" if it's getting mapped to the string field and correct the type, maybe all we need to make this work (in addition to adding the medium option to the enum, is to correct the type?).

I don't understand what the purpose of the name change is, basically.

obulat: ES didn't complain because we weren't even calling the `get_duration` function . We did have it in the codebase, but we were putting the actual integer value into the index. And since the `duration` wasn't in the search query parameters list, we didn't even notice it.

sarayourfriend: Ahh, I see. Because it wasn't I called I assumed there was some magic going on, like with Django serializer method fields :upside_down_face: That's the worst part of magic to me, it makes you start expecting it in places where it doesn't exist :cry: 

So then I'm still left wondering if pushing the enum into the index is the right way to do this over say, using a [range query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html)? If our definitions of short/medium/long changed, for example, then we'd have to update every audio in the index. Or if we wanted to add an `extra_long` or `extra_short` for example, we'd have to update all the existing `short` or `long` to further discriminate them into the more specific groups.

If we use the range query, then it's just a matter of updating the query configuration against the existing `duration` integer field.

This may also apply to image dimensions or even aspect ratio (if we changed the aspect ratio to a decimal number instead of the enum).

Doing this for the image fields would require re-indexing, we might just have to live with the current design for the image fields. But we don't have to make the same problem for the audio duration (if it is indeed a problem).

obulat: I think it was implemented as an Enum because the image `size` was implemented this way :)
The `range query` looks much more sustainable in the long run, for the reasons you describe. Now I'd have to find out how to implement it with Django.

obulat: I wonder how expensive re-indexing images (replacing Enum size with the range query) would be?

sarayourfriend: My guess is "very" but it's an uninformed one :slightly_smiling_face:

obulat:
Well, the ElasticSearch documentation recommends pre-indexing to improve search performance for cases like duration buckets:
https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-search-speed.html#_pre_index_data

I believe that we should optimize for search speed and not the ease of changing the index. To make our search faster it's better to create a field of type `keyword` called `duration_range` and populate it with one of the duration values `short`, `medium`, `long`.

This would mean that we need to be re-check our assumptions for the duration lengths. I would probably revise the ranges. Make **short** up to 2 or 3 minutes - this would be the category people would use to find sound effects to reuse for their video/audio. 2(3)-7 minutes for **medium** range - this is where most songs would end up. And 7-... for **long** - for most podcasts and ebooks.

What do you think, @WordPress/openverse-developers ?
| Name | Duration range | Possible uses |
| --- | --- | --- |
| Short | 0 ... 3 minutes | Sound effects, pronunciation files
| Medium | 3 ... 7 minutes | Songs
| Long | 7 ... infinity | Ebooks, podcasts

sarayourfriend:
That makes sense Olga, thanks for finding that documentation!

I wonder if we should take a queue from that example you shared and make the range name the range itself, rather than an abstract name that could mean any number of things to different people? To me 3 minutes is the length of a whole pop song, is that short? Maybe, I'm not sure! Personally I would probably have assumed short meant less than 1 minute.

obulat:
We have had a lengthy discussion about the best way to implement the `duration` filter for audio in [WP Make Slack](https://wordpress.slack.com/archives/C02012JB00N/p1651591717953969) (needs an account, [which you can create here](https://make.wordpress.org/chat)). I will summarize it here.

## Distribution

@AetherUnbound ran the SQL queries over the data we have to see the distribution of audio items' durations. Below you can see the distribution for the durations of up to 15 minutes, divided into buckets of 1 minute. The second table shows the distribution of audio durations up to 1 minute, with 15-second buckets.

```
openledger> with durations as (select duration, width_bucket(duration, 0, 15 * 60 * 1000, 14) as bucket from audio where duration < 15 * 60 * 1000)
 SELECT bucket, (bucket-1) || '-' || bucket || ' minutes' as range, count(*) from durations group by 1 order by 1;
+--------+---------------+-------+
| bucket | range         | count |
|--------+---------------+-------|
| 1      | 0-1 minutes   | 81395 |
| 2      | 1-2 minutes   | 12685 |
| 3      | 2-3 minutes   | 20156 |
| 4      | 3-4 minutes   | 26488 |
| 5      | 4-5 minutes   | 16129 |
| 6      | 5-6 minutes   | 7680  |
| 7      | 6-7 minutes   | 3675  |
| 8      | 7-8 minutes   | 1891  |
| 9      | 8-9 minutes   | 1058  |
| 10     | 9-10 minutes  | 676   |
| 11     | 10-11 minutes | 415   |
| 12     | 11-12 minutes | 264   |
| 13     | 12-13 minutes | 212   |
| 14     | 13-14 minutes | 174   |
+--------+---------------+-------+
```
```
openledger> with durations as (select duration, width_bucket(duration, 0, 60 * 1000, 4) as bucket from audio where duration < 60 * 1000)
  SELECT bucket, (bucket-1) * 15 || '-' || bucket * 15 || ' seconds' as range, count(*) from durations group by 1 order by 1;
+--------+---------------+-------+
| bucket | range         | count |
|--------+---------------+-------|
| 1      | 0-15 seconds  | 67929 |
| 2      | 15-30 seconds | 6020  |
| 3      | 30-45 seconds | 3416  |
| 4      | 45-60 seconds | 2721  |
```

Among the audio items we currently have, the vast majority are in the shortest,  0-15 seconds,  bucket. Then there is another peak at the 3-4 minutes range bucket. Most of the shortest items are either `sound effect` or the `pronunciation` (the recordings of word pronunciations) type.

## Usage of duration range query
If we go back to the why of the duration query, we can note that it is used not only to get the items of a specific length but also as a proxy for the category of the audio. For example, if I am looking for a song, I would probably look for something in the 3-5 minutes range, and if I need a sound effect to reuse in my video, I would probably select some filter that is up to 1 minute in length. This usage pattern is especially important because our `category` data is based on the mix of data we get (or don't get) from the providers, and is not always very accurate.

## Other considerations
We noted that using a duration query, it can be difficult to understand what the exact term means. A user searching for `short` audio may expect 15-second audio or a 2-minute one. To clarify, we should add the explanation to the query serializer, and we can also use a clarifying phrase on the front end.

## Duration ranges

We decided on the following ranges in the Slack thread:
`0-30s`, `30s-3m`, `3m-10m`, `10m-inf.`
Now, looking at the distributions one more time, and seeing the peak at 2-3 minutes, maybe we should revise it to:
`0-30s`, 30s-**2m**, **2m**-10m, `10m-inf` ?

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -81,8 +81,9 @@ def validate_category(value):
         return value.lower()
 
     @staticmethod
-    def validate_duration(value):
-        _validate_enum("duration", DURATION, value)
+    def validate_length(value):
+        print(f"Validating length: {value}")
+        _validate_enum("length", LENGTHS, value)
```
obulat, line 86:
```suggestion
    def validate_length(value):
        _validate_enum("length", LENGTHS, value)
```

api/catalog/api/constants/field_values.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import annotations
```
AetherUnbound, line 1: What made this line necessary?

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -320,20 +320,23 @@ def database_row_to_elasticsearch_doc(row, schema):
...
             min_boost=min(popularity or 1, authority_boost or 1),
             **attrs,
         )
 
     @staticmethod
-    def get_duration(duration):
+    def get_length(duration):
         if not duration:
             return None
         for length in Audio.Durations:
```
AetherUnbound, line 342: I have to be honest, I don't quite understand this function 😅 wouldn't this `duration < length` check include files at a certain size category _and all files below that size category_ in a search? It seems that if I searched for "long" files, I would get all "medium" and below too since those are `< "long"`.

AetherUnbound (changes requested):
I would be fine with either a 2 min barrier or a 3 min barrier. I think 2 mins might be better because that's the lower end of the "song" length.

I think that we will need to add `SHORTEST` to the Audio model in `elasticsearch_models.py` L307, right?

ingestion_server/ingestion_server/elasticsearch_models.py
sarayourfriend: I had that same though too, but it relies on the order of the audio durations enum and also the fact that if something is labelled "medium" it won't get included in the query for specifically ones labelled "long" because it's a plain keyword search (the search doesn't know the ordinal relationship between the terms).

The ordering matters because it iterates and finds the first one that the duration is _shorter_ than. If it went longest to shortest enum then everything would be long, of course. Otherwise they get checked in order and return the ones that they're shorter than.

However, there's a bug in this PR that will prevent the shortest time from working, it isn't present in the `Audio.Durations` enum.

krysal (changes requested):
This effectively adds the "medium" category to the filter, but it still doesn't work correctly. Doing a call with any of the values `short|medium|long` always returns the same amount of results (5000, the total sample data). 

```
GET /v1/audio/?duration=short

{
    "result_count": 5000,
    "page_count": 250,
    "page_size": 20,
    "page": 1,
    "results": [...]
}
```

Checking with the sample data in the local db most of the audio fall under 'short' and there are no 'long' audio.

```sql
SELECT COUNT(id) FROM audio WHERE duration < 4*60*1000;  -- short
-- count: 4923

SELECT COUNT(id) FROM audio
WHERE duration >= 4*60*1000 AND duration < 20*60*1000;  -- medium
-- count: 77

SELECT COUNT(id) FROM audio WHERE duration >= 20*60*1000;  -- long
-- count: 0
```

The bug seems to be around the function @AetherUnbound pointed to.
---
Edit: I see it works correctly with `length` as the filter param, I got confused by the title 😅 It is still the intention to make the duration filter work? I'll be okay just replacing it with `length` since audio is still in beta and it never worked as you said.

obulat:
I've updated this PR to use `duration` as the query parameter, e.g. `/search/audio?q=song&duration=medium` would return all audio items between 2 and 10 minutes. However, the search result and single result would return the `duration` as a numeric field, in milliseconds, and would not return the `length` enum value (shortest/short/medium/long). Does that make sense, or would it be simpler to use `length` for the range, and `duration` for the numeric value?
The ranges used are `0-30s`, `30s-2m`, `2m-10m`, `10m-inf`.

I'm not sure how we should go about merging this PR considering that we are re-factoring the serializer and the ES indexing.

api/catalog/api/constants/field_values.py
obulat: I have no idea 🤷 It's the IDE...

ingestion_server/ingestion_server/elasticsearch_models.py
obulat: Sorry for not explaining earlier! I have fixed the bug by adding the missing range.

stacimc (commented):
(For others testing) I ran into some errors while trying to `just init` which were resolved with a rebase 😄 

The mismatch between `duration` as a string query parameter and the numeric field returned by the API does feel a little strange to me, though I agree there's trade-offs either way. I probably lean toward using `length` as the query parameter just to avoid overloading the term 🤔

I'm getting some odd results that don't quite match up, for example:
* I get **0** results for `?duration=short` and **188** for `?duration=medium`, but **198** when I query `?duration=short,medium`
* I get **1** result for `?duration=long` and **188** for `?duration=medium`, but **190** for `?duration=long,medium`

obulat:
> (For others testing) I ran into some errors while trying to `just init` which were resolved with a rebase 😄
> 
> The mismatch between `duration` as a string query parameter and the numeric field returned by the API does feel a little strange to me, though I agree there's trade-offs either way. I probably lean toward using `length` as the query parameter just to avoid overloading the term 🤔

I've thought about it more, and I do agree with you. Having two different things be referred to by the same name is confusing me now when I'm writing it, it will definitely confuse everyone in the future. I'm going to re-write it again :)

> I'm getting some odd results that don't quite match up, for example:
> 
>     * I get **0** results for `?duration=short` and **188** for `?duration=medium`, but **198** when I query `?duration=short,medium`
> 
>     * I get **1** result for `?duration=long` and **188** for `?duration=medium`, but **190** for `?duration=long,medium`

This _might_ be caused by the filtering of the dead links. What query do you search for? If you add `filter_dead=false`, you will get the raw results (i.e., before we send the `head` requests to providers to confirm that the items still exist at their URLs). I'll test again to confirm, too.

AetherUnbound:
We're tackling two issues with this PR. One (#653) is presently marked as "critical" priority and the other (#652) is marked as "high". Since we're still having active discussion on aspects of #652, would it be possible to open a PR exclusively to fix #653 while we discuss the other aspects at play in this one? That way we can deploy an API fix next week for the currently broken filter while allowing the discussion that the duration filter inaccuracy & duration vs length topics are due.
If we're close to a resolution on both fronts than my comment can be ignored 🙂

obulat:
I have finally renamed the query parameter to `length`, and I think now it's good to go :)
Unless you still think we should discuss the length categories. 

If we deploy it as it is, though, the `duration` query parameter in the frontend will simply be discarded. We'll need similar changes applied to the frontend.

AetherUnbound (approved):
This looks great to me! And I think the categories are solid. Forgive me if I'm misinterpreting this though - isn't this a breaking change? Since `duration` used to work and now it won't?

stacimc:
@AetherUnbound My understanding was that the frontend is using `duration`, but that it never worked (except for a while in this PR, before it was changed to `length`). So I believe there's no breaking change, but the frontend would need to be updated to use the `length` query param still.

AetherUnbound:
Huh, that's an interesting philosophical question: is it breaking the API if it never worked in the first place 😅 Thanks for the explanation, I'm personally fine with it 🙂

PR #684, opened 2022-05-05
Clean up ES index mappings

dhruvkb:
## Description

This PR
- changes the ES mapping mode from dynamic to <s>strict</s> false (this means unmapped fields are stored but not indexed <s>this prevents adding documents with fields other than those defined in the mapping schema</s>)
- removes unnecessary schema mappings, changing them to purely stored fields
- converts some fields from text+keyword to just keyword
- moves hardcoded dictionary to external JSON files

## Testing Instructions

Since this is a change that makes the ES mappings stricter, passing CI should be adequate.

Additionally, you can do this.

1. On `main`, see the initial mappings by visiting the following URLs:
  - http://localhost:9200/image/_mapping
  - http://localhost:9200/audio/_mapping
2. Check out the PR, drop everything `just down -v` and reinitialise with `just init`.
3. Check the mappings again by visiting the above URLs and compare against the old value. Observe the above mentioned changes in the mapping fields.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
I think I prefer the hand written configuration :sweat_smile: It doesn't seem like it incurs a very big maintenance burden and is much easier to grok than the imperatively generated.

What is the benefit to moving towards the strict mapping?

sarayourfriend:
> this was indeed the case earlier as 2 fields were being inserted into ES using a dynamic mapping

Which two fields were those? Was it causing an issue? Switching to strict mode sounds good to me, I'm just hoping we can document the change.

Do you know if this will cause any necessary infrastructure changes? Do we need to force a reindex after we deploy this or will it automatically do that?

And thanks for the explanation. I prefer the verbosity and ease of groking the hand written one but don't have a strong enough opinion to block this if other folks prefer this approach.

dhruvkb:
The fields were `creator_url` and `thumbnail`. 

We should not need any infrastructure changes because I've tried to keep the final state the same as before. But yeah, a reindex might be needed.

sarayourfriend (commented):
I wonder if it would help this module to be re-written using the Elasticsearch DSL for Python: https://elasticsearch-dsl.readthedocs.io/en/latest/persistence.html#document

It's already in our dependencies anyway, and provides very Pythonic ways for defining our document types, index settings, everything top to bottom basically. It even lets you write unit tests against things like the custom analyzer (which would be pretty cool).

Do you have any thought's about that @dhruvkb?

dhruvkb:
@sarayourfriend the DSL-based models are already present in the https://github.com/WordPress/openverse-api/blob/270b086095ba125fa26e04bf27dff3ab6994e025/ingestion_server/ingestion_server/elasticsearch_models.py file. I'm not completely sure about the relationship between these two files but my understanding is that this file is used to define the indices and then the `elasticsearch_models.py` file is used when creating documents from a PostgreSQL row and pushing it into ES.

The addition of strict mode prevents the `elasticsearch_models.py` from pushing any field that not defined in the `es_mappings.py` file.

sarayourfriend (commented):
I appreciate the clean up. I can't fully articulate why I dislike this approach so much over hand-writing it. You suggested moving it into a JSON file could be an alternative and I personally would prefer that over this approach.

However, I'm fully prepared to accept that I am in the minority here. I have no real criticism other than preferential ones for this. Once it's undrafted I will gladly test it and approve (averting my gaze all the while :stuck_out_tongue:)

sarayourfriend:
Based on our conversation from yesterday, should we draft this PR and add a blocked label to it?

ingestion_server/ingestion_server/indices/mappings/audio.json
```diff
@@ -0,0 +1,14 @@
+{
```
obulat, line 1: I think it makes sense to add `integer` mappings only if we plan to add some filtering/search for these fields. Do [we plan]/[you think we should] to add `bit_rate` or `sample_rate` queries?

ingestion_server/ingestion_server/indices/mappings/common.json
```diff
@@ -0,0 +1,108 @@
...
+    "fields": {
+      "keyword": {
+        "ignore_above": 256,
+        "type": "keyword"
+      }
+    },
+    "similarity": "boolean",
+    "type": "text"
+  },
+  "extension": {
```
obulat, line 35: When does it make sense to have full-text search for extensions? I would think that keyword search for extension should be enough...

ingestion_server/ingestion_server/indices/mappings/common.json
```diff
@@ -0,0 +1,108 @@
...
+        "ignore_above": 8,
+        "type": "keyword"
+      }
+    },
+    "type": "text"
+  },
+  "id": {
+    "type": "long"
+  },
+  "identifier": {
```
obulat, line 47: The same as the above.

ingestion_server/ingestion_server/indices/mappings/common.json
dhruvkb: This was for some very nice cases like `jpg/jpeg`. I might be overthinking this though.

obulat: Since we are processing the file name to get the extension, we can also do replacing there. I don't think that for the user it matters if they find `jpg` and not `jpeg` as they asked, for instance. Does it??

obulat: I'm not sure about the existing index, though. We could add both `jpg` and `jpeg` to the filter as an interim solution, before re-indexing, if necessary.

ingestion_server/ingestion_server/indices/mappings/common.json
dhruvkb: Good catch, fixed in 71e74d9.

ingestion_server/ingestion_server/indices/mappings/common.json
dhruvkb: That was a very uncommon example, we might not even have to remediate this. Changed to `keyword` in 71e74d9.

ingestion_server/ingestion_server/indices/mappings/audio.json
dhruvkb: Dropped fields in 2045eee.

dhruvkb:
Download and install "Elasticvue". It's a browser extension for visualising and modifying your Elasticsearch setup.

On `main`, initialise things using `just init`. You should now have two indices. Alias them to `image-old` and `audio-old`. Using the [clone API](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-clone-index.html) take backups of these indices named `image-bkp` and `audio-bkp`.

Note that these indices are aliased to `image` and `audio`, meaning that they're the live instances on which search is performed.

Then checkout `es_improv`, this branch. Verify that searching works on this branch.  Also exec into the `web` container using `just ipython` and run these Python commands:
```python
from django.conf import settings
print(settings.ES_INDEX_FORMAT)  # should be 'old'
```

Open L386 in `indexer.py` and comment out the line:
```python
es.indices.delete(index=old)
```
Run a new ingest session using `just ingest-upstream` and `just ingest-upstream audio`. You will now have two new indices. Alias them to `image-new` and `audio-new`.

Now the aliases `image` and `audio` point to these new indices. Run the same IPython commands again and you should see the output `'new'`.

api/catalog/api/examples/audio_responses.py
```diff
@@ -36,6 +36,10 @@
     "related_url": f"{origin}/v1/audio/{identifier}/related/",
     "category": "music",
     "duration": 270000,
+    "bit_rate": 128000,
+    "sample_rate": 44100,
+    "filesize": "7139840",
```
obulat, line 41: So great that we can add `filesize` and `filetype` here!
Why is filesize a string, though?

ingestion_server/ingestion_server/indices/mappings/common.json
```diff
@@ -0,0 +1,96 @@
...
+        "type": "keyword"
+      }
+    },
+    "similarity": "boolean",
+    "type": "text"
+  },
+  "extension": {
+    "type": "keyword"
+  },
+  "id": {
```
obulat, line 38: I wonder why we need the separate `id` field, if we already have `_id`?

obulat: I found this in the docs: _In case sorting or aggregating on the _id field is required, it is advised to duplicate the content of the _id field into another field that has doc_values enabled_
And we use the ids of documents to determine whether we should ingest documents (by comparing last id in ES and the DB), but does comparing fall into `sorting` or `aggregating`?

obulat: Apparently, it does:
https://github.com/WordPress/openverse-api/blob/6cc0551c639081f895d974b30cc465713d0ce360/ingestion_server/ingestion_server/indexer.py#L200

ingestion_server/ingestion_server/indices/mappings/common.json
```diff
@@ -0,0 +1,96 @@
...
+    "type": "long"
+  },
+  "identifier": {
+    "type": "keyword"
+  },
+  "license": {
+    "type": "keyword"
+  },
+  "license_version": {
```
obulat, line 47: I don't think we will ever search for licenses by version. We could remove this mapping.

obulat:
> This is an explanation for the commit [aaef5d4](https://github.com/WordPress/openverse-api/commit/aaef5d45557228b959e685d7be6397e8aa638d6d) (and its bugfix [fb154a7](https://github.com/WordPress/openverse-api/commit/fb154a75b0bea4e55e2824f55f7067b514679013)).

Could you also make an issue to remove the `_determine_index_format()  # May 2022 Elasticsearch migration` calls after index migration?

sarayourfriend:
Why do we have to comment out `es.indices.delete(index=old)`. How will this work when we deploy it?

Generally I'm not a fan of using the worlds "old" and "new" to describe what essentially amount to versions (because the "new" eventually becomes "old" :sweat_smile:).

In this case we'll be removing the logic immediately in the next deployment after this one, as @obulat suggested?

Ideally this part would happen automatically:

> Run a new ingest session using just ingest-upstream and just ingest-upstream audio. You will now have two new indices. Alias them to image-new and audio-new using Elasticvue. Now the aliases image and audio point to these new indices. Run the same IPython commands again and you should see the output 'new'.

One of the things I was hoping for out of this plan was to (1) establish a pattern we can rely on in the future that (2) does not add additional steps or complication (and therefore risk) to the people running the deployment that they have to run _during_ the deployment rather than after to have a successful deployment. Additional tasks during deployment can make an already annoyingly long (min 1 hour) deployment that we currently suffer riskier and more likely to go longer.

Would it be possible to follow a pattern somewhat like this?

1. Leave the existing API referencing the plain `{type}` indexes (no version numbers)
2. The new code, to support zero downtime index changes, ingests and references a new index `{type}-{version}`.
3. Deploy ingestion and run ingestion, permanently removing the code that deletes the old index, and run ingestion to create `{type}-{version}` index.
4. Deploy the API that consumes the `{type}-{version}` index that now exists.
    * Notable, steps 3 and 4 can happen on completely different days without causing problems. The existing API will continue to point to the existing `{type}` index, there will just be a second index `{type}-{version}` unused by anything.
5. Manually delete the previous version index (this first case `{type}` but in the future the previous `{type}-{version}`).

Aside from significantly reducing the pressure on deployment to do manual (easily typo'd) manipulations to the live ES quickly, this also has the advantage that the deployed API code doesn't have to "handle" both indexes.

ingestion_server/ingestion_server/indices/mappings/common.json
dhruvkb: This is good, fix pushed in 134a743.

api/catalog/api/examples/audio_responses.py
dhruvkb: It's stored as a number in ES, it's just a serializer bug that makes it a string when sending the response. #696 fixes this by removing the manual definition altogether.

dhruvkb:
I was just testing if things work out, the commenting bit is for local development, testing and debugging purposes only. For the production, I'm woking on a separate PR that updates the ingestion step to be more customisable.

Old and new are used here for just this one instance because as of now they have very clear definitions:
- **old**: ES indices as they exist on `main` and in prod
- **new**: ES indices proposed by this PR which have yet to be deployed
Again all these names and aliases are just for this one migration job. In the future I hope to standardise a proper versioning system involving numbers.

The pattern you suggest @sarayourfriend is correct. I'm in the process of both developing a PoC PR and documenting my findings in a P2 post. It's taking a little longer because I'm new to this but it'll be ready soon. In the meantime, my instruction comment was just to allow us to see how the API can deal with 2 index versions existing simultaneously.

sarayourfriend:
Gotcha, sorry for misunderstanding the goal behind your original comment. Looking forward to what you're able to come up with for this problem :rocket:

AetherUnbound (commented):
I like the approach! Agreed that versioning of the indices is probably the way to go, but as an example of what non-dynamic mappings look like I think it's solid 😄 I have a few code comments, but they're not important if we're jettisoning this anyway.

sarayourfriend:
@dhruvkb What is the status of this PR? Can work on it continue? Does it need to be restarted from fresh given the number of conflicts now. Is there anything you need from anyone else to be able to continue this work?

dhruvkb:
@sarayourfriend perfect timing! I've been updating this PR today and should be able to do it by tomorrow.

dhruvkb:
Closing in favour of #850. Will also address feedback and comments in that PR itself.

ingestion_server/ingestion_server/es_mapping.py
```diff
@@ -1,145 +1,65 @@
...
+    Read and parse the JSON file at the given path.
 
-    :param table_name: The name of the table in the upstream database.
-    :return:
+    :param json_path: the path to the JSON file to be read
+    :return: the parsed contents of the JSON file
+    """
+
+    with json_path.open("r", encoding="utf-8") as json_file:
+        return json.load(json_file)
```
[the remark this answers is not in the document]
dhruvkb: Even better, you can `json.load(json_file)`.

PR #685, opened 2022-05-05
Refactor license constants and utils

dhruvkb:
## Description

This PR refactors and cleans up in the implementation of license constants and utils and their usage in the code.

This PR increases the code's parity with the frontend. A similar effort on the frontend is happening in WordPress/openverse-frontend#1393.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/constants/licenses.py
```diff
@@ -0,0 +1,23 @@
+# This differs from the frontend which does not consider CC0 as a CC license
```
sarayourfriend, line 1: Why this difference?

dhruvkb: I'm not sure, but this has no difference in the outcome. The frontend puts CC0 with the public domain licenses and then in the `isCc` function adds an `OR` condition to include CC0. The backend places it with CC licenses and then doesn't need the `OR` condition. Either way, the difference is historical and the outcome is the same in both cases.

sarayourfriend: I wish there was an easy way to unify these implementations so we don't have to maintain those differences :confused:

dhruvkb: I agree but changing it on the backend side would be wrong as that's a versioned API and also philosophically correct (in my non-lawyer opinion) to consider CC0 a CC license.

sarayourfriend: Well, it's just not _always_ a license, I guess that's the distinction the frontend code chose to make: https://wiki.creativecommons.org/wiki/CC0_FAQ#How_does_it_work.3F

However, I'm a little wary that the distinction matters. It's good to have "technically correct variable names" but it's also good to have easy to understand and obvious code :slightly_smiling_face: Either way, it'd be nice to unify these implementations into a _single implementation_, not just in the actual implementation details. That's more what I meant.

api/test/unit/utils/attribution_test.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.mark.parametrize(
+    "args, expectation",
+    [
+        (
+            ("Title", "Creator", "xy", "0.0", "https://license/url"),
+            '"Title" by Creator is licensed under CC XY 0.0. '
+            "To view a copy of this license, visit https://license/url.",
+        ),
+        (
+            ("", "Creator", "xy", "0.0", "https://license/url"),  # No title
```
AetherUnbound, line 17: Is it possible any of these values could be `None`? Is it worth testing for those cases if so?

dhruvkb: Everything except `license` can be null; `license` is a non-nullable field in the model so we can assume it will always be present. Added the corresponding test case in 7f080f1.

AetherUnbound (approved):
This looks great! Thanks for adding the new test cases, and the notes about frontend parity

sarayourfriend (approved):
LGTM!

PR #696, opened 2022-05-12
Overhaul serializers to eliminate manually defined fields

dhruvkb:
## Description

To further streamline the process of adding new media, this PR updates serializers to almost eliminate manually defined serializer fields.

This PR
- moves `help_text` from serializers to models
- drops fields from serializers that can be inferred from models

## Testing Instructions

<s>Being a refactoring PR, the differences at the surface level should be minimal. To test, visit each endpoint once (sorry for the big ask) and see that the server never raises any errors or sends a 500 status code.</s>

You can also see the changes to `api/catalog/api/examples/image_responses.py` and `api/catalog/api/examples/audio_responses.py` to see the final output changes.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/serializers/audio_serializers.py
```diff
@@ -86,85 +87,51 @@ def validate_duration(value):
...
-        "related_url",
-        "category",
-    ]
-    """
-    Keep the fields names in sync with the actual fields below as this list is
-    used to generate Swagger documentation.
-    """
+    class Meta:
+        model = Audio
+        fields = [  # keep this list ordered logically
```
sarayourfriend, line 98: What does "logically" mean here?

sarayourfriend (commented):
The clean up so far looks really nice but I wonder if we could wait to do this until we have unit tested the serializers in #557?

dhruvkb:
@sarayourfriend since the output is not being changed, I'd recommend #557 proceed with the tests based on serializers as on `main` and we can update this PR to include those tests when ready.

api/catalog/api/serializers/audio_serializers.py
dhruvkb: By `logically` I meant an order that humans would like to read the output JSON. For example ID first, then title, then URL, foreign landing URL, more attrs, finally hyperlinks to the other endpoints etc.

sarayourfriend:
Okay! I was mostly trying to get around at least three of us going through each API endpoint and verifying it works locally :sweat_smile:

dhruvkb:
@sarayourfriend fair point, I anticipated this. So I'll try to that a little easier. You won't have to manually go to each API endpoint and diff the output.

api/catalog/api/serializers/audio_serializers.py
dhruvkb: Added the sort in b107152.

api/catalog/api/models/image.py
```diff
@@ -19,8 +19,16 @@ class ImageFileMixin(FileMixin):
     mixin. Do not use this as the sole base class.
     """
 
-    width = models.IntegerField(blank=True, null=True)
-    height = models.IntegerField(blank=True, null=True)
+    width = models.IntegerField(
+        blank=True,
+        null=True,
+        help_text="The height of the image in pixels. Not always available.",
```
obulat, line 25:
```suggestion
        help_text="The width of the image in pixels. Not always available.",
```

api/catalog/api/models/image.py
```diff
@@ -19,8 +19,16 @@ class ImageFileMixin(FileMixin):
...
+    )
+    height = models.IntegerField(
+        blank=True,
+        null=True,
+        help_text="The width of the image in pixels. Not always available.",
```
obulat, line 30:
```suggestion
        help_text="The height of the image in pixels. Not always available.",
```

api/catalog/api/models/media.py
```diff
@@ -70,19 +87,19 @@ class AbstractMedia(
     meta_data = models.JSONField(blank=True, null=True)
 
     @property
-    def license_url(self):
+    def license_url(self) -> str:
```
obulat, line 90: The license_url should really be available in all media. I know catalog now requires every media item to have it, I wonder how many items in the current DB don't have the URL value...

api/catalog/api/serializers/fields.py
```diff
@@ -0,0 +1,63 @@
...
+        # Only rewrite URLs if a fixed scheme is provided
+        if self.scheme is not None:
+            url = re.sub(r"^\w+://", f"{self.scheme}://", url, 1)
+
+        return url
+
+
+class EnumCharField(serializers.CharField):
+    """
+    This field extends the ``CharField`` to add enum validation.¡
```
obulat, line 31:
```suggestion
    This field extends the ``CharField`` to add enum validation.
```

api/catalog/api/models/media.py
dhruvkb: This is a computed field, so if the license name and version are present, the URL can be determined.

obulat: I know, but the catalog is computing this field, too, if it doesn't get it from providers. And it now saves all three items: license name, version, and URL. So we shouldn't be doing computing here as well.

dhruvkb: The API database does not have this value as a model field, so the ingestion server must not be copying from the catalog. We should make a new issue to add this field to the model and then we can gradually retire this computed function to be a fallback for fields that don't have this value.

obulat: Oh, I wasn't entirely correct.  The license_url *is* now saved, but it is saved as a property of metadata, and not as a separate column.

api/catalog/api/serializers/media_serializers.py
```diff
@@ -264,163 +167,263 @@ def validate(self, data):
...
+        """
+
+    id = serializers.CharField(
+        help_text="Our unique identifier for an open-licensed work.",
+        source="identifier",
     )
 
     tags = TagSerializer(
-        allow_null=True,
+        allow_null=True,  # replaced with ``[]`` later
```
AetherUnbound, line 294: What requirements are necessary for `later`? When can we do this replacement? 😄

api/catalog/api/serializers/media_serializers.py
```diff
@@ -264,163 +167,263 @@ def validate(self, data):
...
+            **_field_attrs,
+        )
+
+        @staticmethod
+        def validate_source_field(input_sources):
+            allowed_sources = list(search_controller.get_sources(media_type).keys())
+            input_sources = input_sources.split(",")
+            input_sources = [x for x in input_sources if x in allowed_sources]
+            input_sources = ",".join(input_sources)
+            return input_sources.lower()
```
AetherUnbound, line 373: Do we want to perform this `lower` step before any comparisons are made to normalize input vs allowed?

api/catalog/api/serializers/base.py
```diff
@@ -1,24 +1,25 @@
...
+
+        :param field_name: the name of the property for which the field is being built
+        :param model_class: the ``class`` instance for the Django model
+        :return: the Field subclass to use and the keyword arguments to pass to it
+        """
+
+        klass, kwargs = super().build_property_field(field_name, model_class)
+        kwargs |= {
+            "allow_null": True,
+        }
```
AetherUnbound, line 22: Why is this step necessary?

api/catalog/api/serializers/fields.py
```diff
@@ -0,0 +1,63 @@
...
+        url = super().get_url(*args, **kwargs)
+
+        # Only rewrite URLs if a fixed scheme is provided
+        if self.scheme is not None:
+            url = re.sub(r"^\w+://", f"{self.scheme}://", url, 1)
+
+        return url
+
+
+class EnumCharField(serializers.CharField):
```
AetherUnbound, line 29: This is great! I love that this is moved out into it's own field.

AetherUnbound (commented):
This is a really cool simplification! I've got a couple of questions:

api/catalog/api/serializers/media_serializers.py
dhruvkb: Later as in lower in the source code inside the `to_representation` function, not as in later in time.

api/catalog/api/serializers/media_serializers.py
dhruvkb: This is good, addressed in 3d4f26b.

api/catalog/api/serializers/base.py
dhruvkb: This is because `Hit` objects don't have the same properties as the Django media models. In those cases, the serializer can exception and return a 500. Setting `allow_null` to `True` allows the same serializer to handle both search results from ES hits and details from the media model instances.

api/catalog/api/serializers/media_serializers.py
```diff
@@ -117,13 +117,14 @@ def validate_q(self, value):
     def validate_license(value):
         """Checks whether license is a valid license code."""
 
-        licenses = value.lower().split(",")
+        value = value.lower()
+        licenses = value.split(",")
```
AetherUnbound, line 121: Nit: I would just combine these lines
```suggestion
        licenses = value.lower().split(",")
```

api/catalog/api/serializers/base.py
AetherUnbound: Got it! Would it be possible to add a comment about that here? Or is there somewhere else that behavior is documented? (Just so I know what to look at later when I forget 😅)

AetherUnbound (approved):
Looks good!

api/catalog/api/serializers/base.py
dhruvkb: Added in 0bb2301.

api/catalog/api/models/mixins.py
```diff
@@ -47,26 +51,47 @@ class MediaMixin(models.Model):
...
+        max_length=2000,
+        blank=True,
+        null=True,
+        help_text="A direct link to the media creator.",
+    )
 
     # Because all forms of media have a thumbnail for visual representation
     thumbnail = models.URLField(
-        max_length=1000, blank=True, null=True, help_text="The thumbnail for the media."
+        max_length=1000,
```
obulat, line 91: We could add a note about `thumbnail` for image being generated by the thumbnail service, and the audio thumbnail being provided by the sources or being null.

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -64,107 +45,94 @@ class AudioSearchRequestSerializer(
...
+            [
+                *MediaSerializer.Meta.fields,
+                *AudioHyperlinksSerializer.field_names,
+                "genres",
+                "alt_files",
+                "audio_set",
+                "duration",
+                "bit_rate",
+                "sample_rate",
+                "waveform",
```
obulat, line 106: What is the difference between `waveform` and `peaks`? `peaks` is the array of numbers, and `waveform`?

api/catalog/api/constants/field_order.py
```diff
@@ -0,0 +1,46 @@
...
+    # Hyperlinks
+    "thumbnail",
+    "detail_url",
+    "related_url",
+    "waveform",  # audio-only
+    # Add-ons
+    "peaks",  # audio-only
+]
+
+field_pos_map: dict[str, int] = {field: idx for idx, field in enumerate(json_fields)}
```
obulat, line 45: Can we use the full name? does this variable mean `field_position_map`?

dhruvkb: Yeah it maps each field to its sorted position. Expanded the name in 3d68558.

api/catalog/api/serializers/fields.py
```diff
@@ -0,0 +1,63 @@
...
+
+from django.conf import settings
+from rest_framework import serializers
+
+from catalog.api.utils.help_text import make_comma_separated_help_text
+
+
+class SchemableHyperlinkedIdentityField(serializers.HyperlinkedIdentityField):
+    """
+    This field returns the link but allows the option to replace the URL scheme.
```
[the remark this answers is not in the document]
dhruvkb: Added explanation in 3d68558.

api/catalog/api/serializers/audio_serializers.py
dhruvkb: Added explanatory comments in 3d68558.

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -64,107 +45,94 @@ class AudioSearchRequestSerializer(
...
+    category = EnumCharField(
+        plural="categories",
+        enum_var=AUDIO_CATEGORIES,
         required=False,
     )
-    length = serializers.CharField(
-        label="length",
-        help_text=make_comma_separated_help_text(LENGTHS, "audio lengths"),
+    length = EnumCharField(
+        plural="audio lengths",
```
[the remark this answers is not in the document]
dhruvkb: Sure! This is just the text that appears in the documentation so we can change that without any side-effects. Changed in 3d68558.

api/catalog/api/models/mixins.py
dhruvkb: Added this information in 3d68558.

api/test/run_test.sh
```diff
@@ -8,7 +8,7 @@ endcol="\e[0m"
 # ```
 # $	./test/run_test.sh test/audio_integration_test.py
 # ```
-TEST_ARG="${1:-test/}"
+if [ $# -ge 1 ]; then TEST_ARG="$@"; else TEST_ARG="test/"; fi
```
sarayourfriend, line 11: Might as well split this across multiple lines :sweat_smile: 

```suggestion
if [ $# -ge 1 ]; then
    TEST_ARG="$@"
else
    TEST_ARG="test/"
fi
```

sarayourfriend (commented):
Why are there new fields being returned/present on the image respones?

api/catalog/api/models/media.py
```diff
@@ -56,8 +64,17 @@ class AbstractMedia(
         default=0,
     )
 
-    tags = models.JSONField(blank=True, null=True)
-    tags_list = ArrayField(models.CharField(max_length=255), blank=True, null=True)
+    tags = models.JSONField(
+        blank=True,
+        null=True,
+        help_text="Tags with detailed metadata, such as accuracy.",
```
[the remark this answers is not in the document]
obulat: Accuracy is present on old tags that have `clarifai` as provider. The docstring for is is quite confusing, though:

https://github.com/cc-archive/cccatalog-api/blob/731074ee543d50edac9aacb9aa0362d03a6bec41/cccatalog-api/cccatalog/api/serializers/image_serializers.py#L252

api/catalog/api/models/mixins.py
```diff
@@ -27,6 +27,10 @@ class ForeignIdentifierMixin(models.Model):
     """
     This mixin adds fields related to the external unique ID to any model. Do
     not use this as the sole base class.
+
+    This mixin adds
+
+    - foreign_identifier: CharField
```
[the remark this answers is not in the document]
dhruvkb: Yes, this is helpful when hovering you're reading a class and hover over the name of it's mixins to see what fields they're adding to the model.

api/catalog/api/serializers/fields.py
```diff
@@ -0,0 +1,66 @@
...
+class EnumCharField(serializers.CharField):
+    """
+    This field extends the ``CharField`` to add enum validation.
+    """
+
+    default_error_messages = serializers.CharField.default_error_messages | {
+        "outside_enum": "Invalid value: {given}. Allowed values: {allowed}"
+    }
+
+    def __init__(self, plural: str, enum_var: set[str], **kwargs):
```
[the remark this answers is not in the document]
dhruvkb: Prefixing `_` is generally associated with public/private. We can change that to `enum_class` though.

api/catalog/api/serializers/image_serializers.py
```diff
@@ -43,82 +48,55 @@ class ImageSearchRequestSerializer(
...
 
-    @staticmethod
-    def validate_aspect_ratio(value):
-        _validate_enum("aspect ratio", ASPECT_RATIOS, value)
-        return value.lower()
+class ImageReportRequestSerializer(MediaReportRequestSerializer):
+    class Meta:
+        model = ImageReport
+        fields = MediaReportRequestSerializer.Meta.fields
+        read_only_fields = MediaReportRequestSerializer.Meta.read_only_fields
```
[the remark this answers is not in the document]
dhruvkb: I agree with you that inheritance would be better. I had forgotten that `Meta` classes had to explicitly set their base and was wondering why it was not working earlier.

dhruvkb:
^ That was an accident, sorry.

api/catalog/api/models/mixins.py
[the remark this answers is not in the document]
sarayourfriend: Gotcha. My concern is that because it is essentially a 1:1 mirror of the implementation that it will easily go out of date quickly (as it already had, with this line needing to be move from one serializer to another without any changes on the actual implementation of either one).

If our base mixin classes are so abstract that it's not easy to remember (or learn) what fields they provide based on the names and reading the declarative code inside their definitions, maybe it's a sign that things are too complex?

It could be I'm over thinking this though. I'm just wary of whenever we start to write documentation that says in literal terms what the code immediately below it says.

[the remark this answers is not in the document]
dhruvkb: I don't like this either, and I'm 100% in agreement with you that the mixins are super-confusing and can be improved. If we did spend a lot of time in the API models (which we don't), we'd soon remember what field was where.

But at the moment, it's a bit hard to see a class inheriting from 4-5 mixins and having to guess which fields are coming from which mixin (or going to each of their definitions to check).

sarayourfriend:
> The goal was to ensure that the search results and the details endpoint should follow as close a schema as possible by avoiding the use of required=False. Extra fields can be ignored by the client but a serializer choosing to omit fields if they're null can be chaotic.

Makes sense, thanks for the explanation.

What's the testing strategy for this? I saw that the "visit each endpoint and verify it" is crossed out. But the sample responses aren't automatically generated are they? Is anything comparing those against the actual data coming from the API this PR produces?

dhruvkb:
@sarayourfriend the example responses are compared by actually making a request to the API and checking if the JSON matches exactly. It tests the API by checking both search results and detail pages.

dhruvkb:
I'm merging this to unblock API deployment. We can surely fix any smaller issues that come up in hotfixes.

sarayourfriend:
Sounds good @dhruvkb :+1:

PR #699, opened 2022-05-16
Refactor search controller for consistency and clarity

dhruvkb:
## Description

This PR refactors the search controller, splitting it up into smaller, easier to understand chunks with proper documentation and type hinting.

This PR 
- removes frivolous arguments to search, using the `MediaSearchRequestSerializer` as the single source of truth for search params
- applies the excluded provider check only when there are non-zero excluded providers
- updates the `get_sources` function to use DSL instead of raw query to ES
- updates usage of `Search` to be idiomatic with Elasticsearch DSL conventions
- moves globally accessible ES connection to `settings.ES`
- reduces code duplication 

These refactor PRs I've been making recently is to make the process of adding media types (like 3D models and video) smoother and more intuitive. It also helps to understand what the code does when you can confidently modify it.

## Testing Instructions

This is a refactor, passing CI should be sufficient proof that no functionality is impacted.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+            fields=["title"],
+            query=f'"{quotes_stripped}"',
+            boost=10000,
+        )
+        s.query = Q("bool", must=s.query, should=exact_match_boost)
+    else:
+        query_bases = ["creator", "title", ("tags", "tags.name")]
+        for query_basis in query_bases:
+            if isinstance(query_basis, tuple):
+                ser_field, es_field = query_basis
```
obulat, line 141: Can we rename the variable to `serializer_field`? This way it makes reading the code more difficult.

dhruvkb: It was `serializer_field` before but that was causing Black to introduce weirdly positioned line breaks. But I understand your point and will make the change :+1:

api/catalog/api/controllers/elasticsearch/utils.py
```diff
@@ -0,0 +1,160 @@
...
+        raise ValueError("Deep pagination is not allowed.")
+    return start_slice, end_slice
+
+
+def _post_process_results(
+    s, start, end, page_size, search_results, filter_dead
+) -> List[Hit]:
+    """
+    After fetching the search results from the back end, iterate through the
+    results, perform image validation, and route certain thumbnails through our
```
obulat, line 103: This is not really in the scope of this PR, but where exactly is this done: `route certain thumbnails thumbnails through our proxy` in `_post_process_results` function?

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+    if isinstance(basis, tuple):
+        serializer_field, es_field = basis
+    else:
+        serializer_field = es_field = basis
+    if serializer_field in search_params:
+        filters = []
+        for arg in search_params[serializer_field].split(","):
+            filters.append(Q("term", **{es_field: arg}))
+        method = getattr(s, behaviour)  # can be ``s.filter`` or ``s.exclude``
+        return method("bool", should=filters)
```
obulat, line 68: From my limited understanding of ES, I _think_ we can replace the `should` with `must` here, because we don't need the results that do not contain the filter parameter (i.e., when we are filtering by some value, we don't need the score that `should` can bring).
This is not a suggestion for this PR, I'm just putting my thoughts in words here :)

api/catalog/api/utils/pagination.py
```diff
@@ -1,48 +1,18 @@
...
-
-    @page.setter
-    def page(self, value):
-        if value is None or not str(value).isnumeric():
-            value = 1
-        value = int(value)  # convert str params to int
-        if value <= 0:
-            raise get_api_exception("Page must be greater than 0.", 400)
-        self._page = value
+        self.page_size = 20
```
obulat, line 14: Why are we removing the page size and page setting?
Wouldn't this be a breaking change for the API?

api/catalog/api/utils/pagination.py
```diff
@@ -1,48 +1,18 @@
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
-from catalog.api.utils.exceptions import get_api_exception
-
 
 class StandardPagination(PageNumberPagination):
-    page_size_query_param = "page_size"
-    page_query_param = "page"
+    page_size_query_param = None
```
obulat, line 6: If we are not going to use these, then we can simply remove the `page_size_query_param`. It is already set to to `None` in the parent class.

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+    search request.
+
+    :param query_ser: the ``MediaSearchRequestSerializer`` instance with search query
+    :param index: The Elasticsearch index to search (e.g. 'image')
+    :param ip: the users' hashed IP to consistently route to the same ES shard
+    :return: the list of search results with the page and result count
+    """
+
+    s = Search(using="default", index=index)
+    search_params = query_ser.data
```
obulat, line 89: Nit: can we rename it to the full `query_serializer`?

api/catalog/api/utils/pagination.py
obulat: And this is the only minor thing blocking this PR, I guess.

obulat (approved):
I really love how much clearer the code is now. It's great to have ES global with the Elasticsearch connection, too!

api/catalog/api/utils/pagination.py
dhruvkb: Yeah, by adding `page` and `page_size` to the serializer, we're able to 
- validate their limits using `min_value` and `max_value` in the serializer
- have that information passed directly to the search controller without manual effort

api/catalog/api/utils/pagination.py
dhruvkb: Ah this is a good catch, I saw that `page_query_param` was `'page'` but I missed that `page_size_query_param` was `None`.

dhruvkb: Fixed in c461450.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: Okay, I like short variable names but this is much clearer. Expanded in ca45eb1.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: I thought so too, but I don't know enough ES to be confident enough to make the change. I'll update it if you're sure that it should be `must` instead of `should`.

api/catalog/api/controllers/elasticsearch/utils.py
dhruvkb: This code is in `api/utils/validate_images.py`. It's not exactly what the comment describes though, it checks for broken images but it does not route any images through our proxy. I think we used to do that before for `http://` images but I think we stopped doing sometime and didn't update the comment.

api/catalog/api/controllers/elasticsearch/related.py
```diff
@@ -0,0 +1,42 @@
+from elasticsearch_dsl import Search
+
+from catalog.api.controllers.elasticsearch.utils import (
+    _exclude_filtered_providers,
+    _get_query_slice,
+    _get_result_and_page_count,
+    _post_process_results,
```
AetherUnbound, line 7: If we're importing these functions to other modules, I think it might be best to have them be public rather than "private"

api/catalog/api/controllers/elasticsearch/related.py
```diff
@@ -0,0 +1,42 @@
...
+
+def related_media(uuid, index, filter_dead):
+    """
+    Given a UUID, find related search results.
+    """
+    search_client = Search(using="default", index=index)
+
+    # Convert UUID to sequential ID.
+    item = search_client
+    item = item.query("match", identifier=uuid)
```
AetherUnbound, line 19: Is this variable declaration necessary?
```suggestion
    item = search_client.query("match", identifier=uuid)
```

api/catalog/api/controllers/elasticsearch/related.py
```diff
@@ -0,0 +1,42 @@
...
+    _id = item.execute().hits[0].id
+
+    s = search_client
+    s = s.query(
```
AetherUnbound, line 23: (same as above)
```suggestion
    s = search_client.query(
```

api/catalog/api/controllers/elasticsearch/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    filtered_providers = cache.get(key=filter_cache_key)
+    if not filtered_providers:
+        filtered_providers = ContentProvider.objects.filter(filter_content=True).values(
+            "provider_identifier"
+        )
+        cache.set(
+            key=filter_cache_key,
+            timeout=FILTER_CACHE_TIMEOUT,
+            value=filtered_providers,
+        )
```
AetherUnbound, line 39: If the cache doesn't have any filtered providers, and no `ContentProvider`s have the field `filter_content=True`, it seems like this will attempt to update the cache every time the function is called. Is there a way to differentiate between "the cache doesn't have this value" and "this value is an empty list"?

api/catalog/api/controllers/elasticsearch/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    :param s: The elasticsearch Search object
+    :param page_size: How big the page should be.
+    :param page: The page number.
+    :return: Tuple of start and end.
+    """
+    query_hash = get_query_hash(s)
+    query_mask = get_query_mask(query_hash)
+    if not query_mask:
+        start = 0
+        end = ceil(page_size * page / (1 - DEAD_LINK_RATIO))
```
AetherUnbound, line 62: I'm confused as to what `DEAD_LINK_RATIO` is being used here for, do you mind elaborating on it? (If this is just inherited code that's been moved around and we don't quite understand what's happening feel free to ignore 😅)

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+            "aspect_ratio",
+            "size",
+            "length",
+            "source",
+            ("license", "license.keyword"),
+            ("license_type", "license.keyword"),
+        ],
+        "exclude": [
+            ("excluded_source", "source"),
+        ],
```
AetherUnbound, line 105: This seems like something we'll have to edit when adding more media types or changing available columns - do you think it might be best in a `constants` module or somewhere alongside other lists of available columns? I worry it'll be overlooked when adding something.

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+        s,
+        search_params["page_size"],
+        search_params["page"],
+        search_params["filter_dead"],
+    )
+    s = s[start:end]
+
+    try:
+        if settings.VERBOSE_ES_RESPONSE:
+            log.info(pprint.pprint(s.to_dict()))
```
AetherUnbound, line 176: I don't know if we have an issue open for this already, but this is great! It'll surely reduce the noise in our logs 💯

AetherUnbound (approved):
This is awesome! I have a few comments/potential improvements, but this is good to go as-is IMO.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: This already existed before, it's not new or added as a part of this PR.

api/catalog/api/controllers/elasticsearch/utils.py
dhruvkb: It was existing code. I don't know what this variable is or does. We can probably leave it untouched for now.

api/catalog/api/controllers/elasticsearch/related.py
dhruvkb: Good catch, fixed in 64a2ed9.

api/catalog/api/controllers/elasticsearch/related.py
dhruvkb: Good catch, fixed in 64a2ed9.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: I'll make a separate issue to improve this filtering setup, this PR is blocking others.

api/catalog/api/controllers/elasticsearch/search.py
AetherUnbound: Oh 😂

api/catalog/api/controllers/elasticsearch/utils.py
AetherUnbound: Sounds good!

zackkrida:
@WordPress/openverse can we explore re-merging this work and getting it in an upcoming API deploy? I'd hate to 1. lose this work or 2. have the API drift so far from the changes here that it becomes difficult or impossible to merge.

AetherUnbound:
Fully support that post-v2.5.6!

PR #707, opened 2022-05-19
Separate go-live step from indexing

dhruvkb:
## Description

This PR
- splits reindexing into a two step process
  - reindex / ingest upstream
  - point alias
- updates the init script to run unchanged with manual go-live
- fixes a bug preventing use of indexer workers locally
- deletes unused chunks code
  - CLI interface for `indexer.py`
  - listen mode for `indexer.py`
  - `_index_table` function in `indexer.py`
- refactors the API and task system for code quality
 
## Testing Instructions

If CI passes, it proves that the manual go-live and indexing are all running as intended. The API tests also prove that data is making it to the serializers.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal:
Could you create the issue for the bug with indexer workers? That way we can follow the reproduction steps and test if it's fixed here.

justfile
```diff
@@ -114,7 +114,7 @@ cert:
 
 # Check if the media is indexed in Elasticsearch
 @check-index index="image":
-    -curl -sb -H "Accept:application/json" "http://localhost:9200/_cat/aliases/{{ index }}" | grep -c "{{ index }}-"
+    -curl -sb -H "Accept:application/json" "http://localhost:9200/_cat/indices/{{ index }}" | grep -o "{{ index }}" | wc -l | xargs
```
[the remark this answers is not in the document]
dhruvkb: `wc -l` has leading spaces, `xargs` trims it.

ingestion_server/ingestion_server/api.py
```diff
@@ -47,124 +47,170 @@ def _get_base_url(req):
...
-    def _validate_create_task(request):
+    @jsonschema.validate(
+        req_schema={
+            "type": "object",
+            "properties": {
+                "model": {"type": "string", "enum": MEDIA_TYPES},
+                "action": {
+                    "type": "string",
+                    "enum": list(task_type.name for task_type in TaskTypes),
+                },
```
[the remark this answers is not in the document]
dhruvkb: There was some manual validation earlier, this just formalises the API spec.

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
+        """
+
+        def _time_fmt(timestamp: int) -> Optional[str]:
+            """
+            Format the timestamp into a human-readable date and time notation.
+            :param timestamp: the timestamp to format
+            :return: the human-readable form of the timestamp
+            """
+
+            if not timestamp:
```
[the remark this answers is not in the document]
dhruvkb: Actually the timestamp for unfinished jobs is set to a value of `0` (falsy). This handles that case and returns finish time as `None` if it encounters that situation. The timestamp cannot be `None`.

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
+
+    @staticmethod
+    def serialize_task_info(task_info: dict) -> dict:
+        """
+        Generate a response dictionary containing all relevant information about a task.
+        :param task_info: the stored information about the task
+        :return: the details of the task to show to the user
+        """
+
+        def _time_fmt(timestamp: int) -> Optional[str]:
```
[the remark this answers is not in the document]
dhruvkb: It's only being used in this function and nowhere else. So it sort of made sense to put it inside the function as one would put a `lambda`.

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
+        """
+        Get the statuses of all tasks.
+        :return: the statuses of all tasks
+        """
+
+        results = [self.list_task_status(task_id) for task_id in self.tasks.keys()]
+        results.sort(key=lambda task: task["finish_timestamp"])
+        return results
+
+    def list_task_status(self, task_id) -> dict:
```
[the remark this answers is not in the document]
dhruvkb: Good catch, renamed in 00b5c86.

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
-    UPDATE_INDEX = 1
-    # Download the latest copy of the data from the upstream database, then
-    # completely reindex the newly imported data.
-    INGEST_UPSTREAM = 2
-    # Create indices in Elasticsearch for QA tests.
-    # This is not intended for production use, but can be safely executed in a
-    # production environment without consequence.
-    LOAD_TEST_DATA = 3
+    @staticmethod
+    def _generate_next_value_(name: str, *args, **kwargs) -> str:
```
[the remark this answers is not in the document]
dhruvkb: Added the docstring in 0919b01.

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
+        if model == "audio":
+            reload_upstream("audioset", approach="basic")
+        indexer.reindex(model, **kwargs)
+
+    try:
+        locs = locals()
+        if func := locs.get(action.value):
+            func()  # Run the corresponding task function
+        elif func := getattr(indexer, action.value):
+            func(model, **kwargs)
```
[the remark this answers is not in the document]
dhruvkb: This flexibility allows us to define new methods in `indexer.py` without having to declare mapping for them in a different file. So as long as your `task.py` only involves forwarding the call to `TableIndexer`, there isn't a need to change anything in this file at all.

Added comments in c255241 if you think they'd help?

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
+
+        self._prune_old_tasks()
+
+        task_info = self.tasks[task_id]
+        return {"task_id": task_id} | self.serialize_task_info(task_info)
+
+
+def perform_task(
+    task_id: str,
+    model: Literal["image", "audio", "model_3d"],
```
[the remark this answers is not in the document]
dhruvkb: That's true, added the type in c2e4698.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -309,135 +300,81 @@ def replicate(self, table, dest_index, query):
...
-            self._index_table(model_name, dest_idx=destination_index)
-            slack.verbose(
-                f"`{model_name}`: Elasticsearch reindex complete | "
-                f"_Next: promote index as primary_"
-            )
-            self.go_live(destination_index, model_name)
 
-    def update(self, model_name: str, since_date):
-        log.info(f"Updating index {model_name} with changes since {since_date}")
+        destination_index = f"{model_name}-{index_suffix}"
```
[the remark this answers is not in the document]
dhruvkb: So the idea is that all of the tasks' public APIs will receive the model name as the first parameter, and any additional params from the API directly. See `reindex`, `update_index`, `load_test_data` and `point_alias` (even the `delete_alias` added in the other PR).

Since the API accepts `index_suffix`, that's the argument this function gets. It's not a very expensive operation to join the model name and the suffix.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -169,64 +172,49 @@ def get_last_item_ids(table):
...
-                table=Identifier(table),
-                last_es_id=Literal(last_es_id),
-                last_pg_id=Literal(last_pg_id),
-            )
-            self.es.indices.create(index=dest_idx, body=index_settings(table))
-            self.replicate(table, dest_idx, query)
+            model = database_table_to_elasticsearch_model[origin_table]
+        except KeyError:
+            log.error(f"Table {origin_table} is not defined in elasticsearch_models.")
+            return []
```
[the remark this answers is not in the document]
dhruvkb: I hadn't modified this code at all, just rearranged it in the file (guess Git isn't very good when it comes to that). But your suggestion makes sense. Changed in 789b4e4.

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,189 @@
...
+        }
+
+    def list_task_statuses(self) -> list:
+        """
+        Get the statuses of all tasks.
+        :return: the statuses of all tasks
+        """
+
+        results = [self.list_task_status(task_id) for task_id in self.tasks.keys()]
+        results.sort(key=lambda task: task["finish_timestamp"])
```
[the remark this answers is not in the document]
dhruvkb: This situation can never arise
- `finish_timestamp` is always present (sent in the `add_task` call from `api.py`)
- `finish_timestamp.value` is always a number (it is set to `Value("d", 0.0)`)

ingestion_server/ingestion_server/constants/media_types.py
```diff
@@ -2,8 +2,13 @@
...
+
+
 AUDIO_TYPE = "audio"
 IMAGE_TYPE = "image"
 MODEL_3D_TYPE = "model_3d"
 
-MEDIA_TYPES = [AUDIO_TYPE, IMAGE_TYPE, MODEL_3D_TYPE]
+MediaType = Literal["audio", "image", "model_3d"]
+
+MEDIA_TYPES: list[MediaType] = list(get_args(MediaType))
```
AetherUnbound, line 14: Nice 💯

AetherUnbound (approved):
Sweet!!

ingestion_server/ingestion_server/tasks.py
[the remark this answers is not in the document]
dhruvkb: Made the change in 151b74e.

ingestion_server/env.docker
```diff
@@ -9,3 +9,5 @@ UPSTREAM_DB_PORT="5432"
 
 LOCK_PATH="/worker_state/lock"
 SHELF_PATH="/worker_state/db"
+
+INDEXER_WORKER_HOST="indexer_worker"
```
obulat, line 13: Will this require any changes in the infrastructure repo?

ingestion_server/ingestion_server/api.py
```diff
@@ -47,124 +47,170 @@ def _get_base_url(req):
...
+                {
+                    "if": {
+                        "properties": {"action": {"const": TaskTypes.POINT_ALIAS.name}}
+                    },
+                    "then": {"required": ["index_suffix", "alias"]},
+                },
+            ],
+        }
+    )
+    def on_post(self, req, res):
```
obulat, line 82: What happens if `req` is not validated?
I don't quite understand how this validation works and where `get_media` method is in `req`, if it's not in the schema...

ingestion_server/ingestion_server/tasks.py
```diff
@@ -2,155 +2,199 @@
...
-        self.id_action[task_id] = action
-        self.id_progress[task_id] = progress
-        self.id_start_time[task_id] = dt.datetime.utcnow().timestamp()
-        self.id_finish_time[task_id] = finish_time
-        self.id_active_workers[task_id] = active_workers
-        return task_id
+        self.tasks = {}
 
     def _prune_old_tasks(self):
+        # TODO: Populate, document or delete function stub
```
obulat, line 63: What is the purpose of this function now?

obulat (dismissed):
I ran the updated API, and all worked great (after `recreate`, because at first it complained about missing `jsonschema` dependency). 
I love the more detailed task view! 
It does not add `finish_time`, though: 
```
{
  "task_id": "75dbff29f5104a8d97d3399b97a9a371",
  "active": false,
  "model": "image",
  "action": "POINT_ALIAS",
  "progress": 100.0,
  "start_timestamp": 1654159447.424355,
  "start_time": "2022-06-02 08:44:07.424355",
  "finish_timestamp": 0.0,
  "finish_time": null, 
  "active_workers": false, 
  "error": false
}
```

ingestion_server/env.docker
dhruvkb: No it should not. It's a change specific to the local Docker setup.

ingestion_server/ingestion_server/api.py
dhruvkb: Falcon provides a built-in decorator `jsonschema.validate` to validate the body of the incoming request. It will run the validation before the function body executes and return 400 in case of validation failure. You can refer to its documentation [here](https://falcon.readthedocs.io/en/stable/api/media.html#validating-media).

`get_media` is a method defined on the Falcon `Request` object. You can refer to its documentation [here](https://falcon.readthedocs.io/en/stable/api/request_and_response_wsgi.html#falcon.Request.get_media).

ingestion_server/ingestion_server/tasks.py
dhruvkb: To be honest, we could delete it (it doesn't do anything) but I wanted to be sure that there was nothing planned for it before we removed it from the codebase.

load_sample_data.sh
```diff
@@ -102,10 +102,14 @@ just load-test-data "image"
...
 # Ingest and index the data
-just ingest-upstream "audio"
+just ingest-upstream "audio" "init"
+just wait-for-index "audio-init"
+just point-alias "audio" "init" "audio"
 just wait-for-index "audio"
 
-just ingest-upstream "image"
+just ingest-upstream "image" "init"
+just wait-for-index "image-init"
```
krysal, line 111: `just wait-for-index ...` throw an error for audio and image to me: `sh: timeout: command not found`.

Which causes an error at the search list endpoint:

```python
NotFoundError(404, 'index_not_found_exception', 'no such index [audio]', audio, index_or_alias)
```

Unless I point the alias manually, then I get no audio results 😓

dhruvkb: Ah, `timeout` has to be installed separately I suppose. I must have got that installed as a dependency of some other package.

Screenshot 2022-06-30 at 7 51 58 AM
 
Could you try installing [`coreutils` using Homebrew](https://formulae.brew.sh/formula/coreutils#default)? If that fixes the problem, I'll update the docs with this prerequisite, or change the timeout script to use Python as the base language.

ingestion_server/ingestion_server/api.py
```diff
@@ -47,124 +47,170 @@ def _get_base_url(req):
...
+
+        # Required fields
+
+        model: MediaType = body[MODEL]
+        action = TaskTypes[body[ACTION]]
+
+        # Optional fields
+        callback_url = body.get("callback_url")
+        since_date = body.get("since_date")
+        index_suffix = body.get("index_suffix", task_id)
```
stacimc, line 103: I re-tested with #578 and I think I'm seeing what @krysal  is seeing -- `NotFoundError(404, 'index_not_found_exception', 'no such index [audio]', audio, index_or_alias)`.

I did a small amount of investigating and strangely, it looks like we're not getting the same `index_suffix` here for the two POSTs. In both cases `index_suffix` is in the request body, but I'm seeing a different uuid. Therefore the `POINT_ALIAS` task fails to find the index.

When I initially tested this I generated a uuid locally, then used `curl` commands and manually passed the same one. This all works! So while it's a strange problem, I don't think it's an issue with this code 🤔

stacimc:
This may not be part of the intention of this PR, but do you think promoting the temporary table/deleting the original index should be considered part of the "go live" step and moved out of the `ingest_upstream` step? That currently happens [here](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/ingest.py#L369) in `reload_upstream`.

load_sample_data.sh
krysal: It worked! Though it would be nice if we could spare that dependency. Thanks for looking at this Dhruv!

krysal:
Installing [`coreutils`](https://formulae.brew.sh/formula/coreutils#default) worked for me (🎉) I can take a better look at this tomorrow but if folks are eager to merge this I suggest cutting a release before and after, that way we can deploy changes in isolation.

dhruvkb:
@stacimc that's a good point that I hadn't given enough thought and factored in properly. I'll mark this PR as a draft and work on that. Might also have to update the state diagram.

@krysal there's no eagerness to merge it (especially with @stacimc having found a major point I had not considered earlier) so take as much as you need!

ingestion_server/ingestion_server/ingest.py
```diff
@@ -326,26 +320,48 @@ def reload_upstream(
...
+    7. Recreate constraints from the original table: ``_generate_constraints``
+    8. Promote the temp table and delete the original: ``get_go_live_query``
+
+    :param table: the upstream table to copy
+    :param progress: multiprocessing.Value float for sharing task progress
+    """
+
+    log.info(f"`{table}`: Starting table promotion | _Next: recreate-indices_")
+    downstream_db = database_connect()
+    _, index_mapping = _generate_indices(downstream_db, table)
```
stacimc, line 360: This is also defined on L365

ingestion_server/ingestion_server/indexer.py
```diff
@@ -169,64 +171,46 @@ def get_last_item_ids(table):
...
-                last_es_id=Literal(last_es_id),
-                last_pg_id=Literal(last_pg_id),
-            )
-            self.es.indices.create(index=dest_idx, body=index_settings(table))
-            self.replicate(table, dest_idx, query)
+        # Map column names to locations in the row tuple
+        schema = {col[0]: idx for idx, col in enumerate(columns)}
+        model = database_table_to_elasticsearch_model.get(model_name)
+        if model is None:
+            log.error(f"Table {model_name} is not defined in elasticsearch_models.")
```
stacimc, line 201: I'm slightly confused by table/model variable names in this method. It looks like `database_table_to_elasticsearch_model` expects a `table_name` rather than a `model_name`. Should this method take `table_name` instead and pass that here?

ingestion_server/ingestion_server/indexer.py
```diff
@@ -286,8 +274,8 @@ def replicate(self, table, dest_index, query):
                 es_batch = self.pg_chunk_to_es(
                     pg_chunk=chunk,
                     columns=server_cur.description,
-                    origin_table=table,
-                    dest_index=dest_index,
+                    model_name=model_name,
```
stacimc, line 277: Same question as above, should we pass `table_name` here instead?

ingestion_server/ingestion_server/indexer.py
```diff
@@ -309,135 +297,89 @@ def replicate(self, table, dest_index, query):
...
-                f"_Next: promote index as primary_"
-            )
-            self.go_live(destination_index, model_name)
 
-    def update(self, model_name: str, since_date):
-        log.info(f"Updating index {model_name} with changes since {since_date}")
+        destination_index = f"{model_name}-{index_suffix}"
+
+        log.info(f"Updating index {destination_index} with changes since {since_date}.")
         deleted, mature = get_existence_queries(model_name)
```
stacimc, line 383: Do we need to pass `table_name` here now too?

ingestion_server/ingestion_server/api.py
```diff
@@ -47,124 +47,175 @@ def _get_base_url(req):
...
-        active_workers = self.tracker.id_active_workers[task_id]
+        task_info = self.tracker.tasks[task_id]
+        active_workers = task_info["active_workers"]
 
         # Update global task progress based on worker results
-        self.tracker.id_progress[task_id].value = task_data.percent_successful
+        task_info["progress"].value = task_data.percent_successful
 
         if task_data.percent_successful == 100:
             logging.info(
```
stacimc, line 221: The log below (`Attempting to promote index {target_index}`) is no longer accurate, now that the promotion step has been moved out of `ingest_upstream`.

stacimc (changes requested):
When I run `just recreate` and `just init`, I get several errors along the lines of:
```
ERROR tasks.py:204 - Error processing task `PROMOTE` for `image`: relation "public.temp_import_image" does not exist
```

I can run `ingest_upstream` with a new suffix (`just ingest-upstream "image" "test"`), but when I try to run `promote`, or if I re-run `ingest-upstream` with the default "init" suffix, I'm getting disk space errors:

```
elasticsearch.exceptions.TransportError: TransportError(429, 'cluster_block_exception', 'index [audio-test] blocked by: [TOO_MANY_REQUESTS/12/disk usage exceeded flood-stage watermark, index has read-only-allow-delete block];')
```

😱 Any ideas what's going on there? I have had some environment issues with the api recently, so I wonder if someone else could try to repro.

ingestion_server/ingestion_server/indexer.py
dhruvkb: So existence queries will reuse the model name as the table name if you don't explicitly pass it (which it basically accurate).

ingestion_server/ingestion_server/indexer.py
dhruvkb: `pg_chunk_to_es` does not need the table name (it already has the chunk of rows from the table) but rather the model name so that it can map those rows to the right ES fields. This is correct and works as expected.

ingestion_server/ingestion_server/ingest.py
dhruvkb: Good find! Removed the redundant line in a8468c7.

ingestion_server/ingestion_server/indexer.py
dhruvkb: The `database_table_to_elasticsearch_model` identifier is badly named, it actually takes the name of a media type and returns the ES models for that media. It uses the model name. Updated the name in 9e70ce0.

ingestion_server/ingestion_server/indexer.py
stacimc: This makes much more sense now that `database_table_to_elasticsearch_model` has been renamed! Got it, thank you :)

ingestion_server/ingestion_server/indexer.py
stacimc: It will definitely work now, since the model names all match the table names. I'm wondering if this would break if we introduced a model/table where that was not the case?

stacimc:
That's interesting -- I actually have more resources allocated 🤔 At any rate I think that was a red herring -- the real issue was with the first error message:

```
ERROR tasks.py:204 - Error processing task `PROMOTE` for `image`: relation "public.temp_import_image" does not exist
```

I just took another look and found that the `wait-for-index` steps in `load_sample_data` were failing, causing my problems. I'm not sure when this got borked in my environment, but in the event this happens to someone else: I just needed to `brew install coreutils`.

stacimc (approved):

PR #714, opened 2022-05-20
Add task for deleting indices

dhruvkb:
## Description

This PR adds a new task type `DELETE_INDEX` useful to delete unused indices via the API.

## Testing Instructions

1. Try deleting an index that is not associated with an alias. It should succeed.
2. Try deleting an index that is associated with an alias. It should fail.
3. Try deleting an index that is associated with an alias using `force_delete` set to `true`. It should succeed. 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -200,7 +200,7 @@ class TaskResource(BaseTaskResource):
                 "task_id": task_id,
                 "status_check": status_url,
             }
-        elif is_bad_request.value:  # set to 1 for bad tasks
+        elif is_bad_request.value == 1:
             res.status = falcon.HTTP_400
             res.media = {
                 "message": "Failed during task execution due to bad request. Check "
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -145,6 +145,7 @@ class TaskResource(BaseTaskResource):
         progress = Value("d", 0.0)
         finish_time = Value("d", 0.0)
         active_workers = Value("i", int(False))
+        is_bad_request = Value("i", 0)
 
         task = Process(
             target=perform_task,
@@ -156,6 +157,7 @@ class TaskResource(BaseTaskResource):
                 "progress": progress,
                 "finish_time": finish_time,
                 "active_workers": active_workers,
+                "is_bad_request": is_bad_request,
                 # Task-specific keyword arguments
                 "since_date": since_date,
                 "index_suffix": index_suffix,
@@ -174,6 +176,7 @@ class TaskResource(BaseTaskResource):
             progress=progress,
             finish_time=finish_time,
             active_workers=active_workers,
+            is_bad_request=is_bad_request,
         )
 
         base_url = self._get_base_url(req)
@@ -196,6 +199,12 @@ class TaskResource(BaseTaskResource):
                 "message": "Successfully completed task",
                 "task_id": task_id,
                 "status_check": status_url,
+            }
+        elif is_bad_request.value:  # set to 1 for bad tasks
+            res.status = falcon.HTTP_400
+            res.media = {
+                "message": "Failed during task execution due to bad request. Check "
+                "scheduler logs."
             }
         else:
             res.status = falcon.HTTP_500
--- a/ingestion_server/ingestion_server/api.py
+++ b/ingestion_server/ingestion_server/api.py
@@ -15,7 +15,8 @@ from falcon.media.validators import jsonschema
 
 from ingestion_server import slack
 from ingestion_server.constants.media_types import MEDIA_TYPES, MediaType
-from ingestion_server.indexer import TableIndexer, elasticsearch_connect
+from ingestion_server.es_helpers import elasticsearch_connect, get_stat
+from ingestion_server.indexer import TableIndexer
 from ingestion_server.state import clear_state, worker_finished
 from ingestion_server.tasks import TaskTracker, TaskTypes, perform_task
 
@@ -31,6 +32,25 @@ class HealthResource:
     def on_get(_, resp):
         resp.status = falcon.HTTP_200
         resp.media = {"status": "200 OK"}
+
+
+class StatResource:
+    @staticmethod
+    def on_get(_, res, name):
+        """
+        Handles an incoming GET request. Provides information about the given index or
+        alias.
+
+        :param _: the incoming request
+        :param res: the appropriate response
+        :param name: the name of the index or alias
+        :return: the information about the index or alias
+        """
+
+        elasticsearch = elasticsearch_connect()
+        stat = get_stat(elasticsearch, name)
+        res.status = falcon.HTTP_200
+        res.media = stat._asdict()
 
 
 class BaseTaskResource:
@@ -61,6 +81,7 @@ class TaskResource(BaseTaskResource):
                 "since_date": {"type": "string"},
                 "index_suffix": {"type": "string"},
                 "alias": {"type": "string"},
+                "force_delete": {"type": "boolean"},
             },
             "required": ["model", "action"],
             "allOf": [
@@ -80,6 +101,17 @@ class TaskResource(BaseTaskResource):
                         "properties": {"action": {"const": TaskTypes.UPDATE_INDEX.name}}
                     },
                     "then": {"required": ["index_suffix", "since_date"]},
+                },
+                {
+                    "if": {
+                        "properties": {"action": {"const": TaskTypes.DELETE_INDEX.name}}
+                    },
+                    "then": {
+                        "oneOf": [
+                            {"required": ["alias"]},
+                            {"required": ["index_suffix"]},
+                        ]
+                    },
                 },
             ],
         }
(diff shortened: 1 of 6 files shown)


AetherUnbound (approved):
This looks great! Nice addition, I'm sure that we have tons of unused indices.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -463,3 +463,44 @@ def point_alias(self, model_name: str, index_suffix: str, alias: str, **_):
...
+                # No alias associated imp lies unused index.
+                self.es.indices.delete(index=index_name)
+                message = f"Index {index_name} was deleted."
+                log.info(message)
+                slack.info(message)
+        except NotFoundError:
+            # Index does not exist
+            message = f"Index {index_name} does not exist and cannot be deleted."
+            log.info(message)
+            slack.info(message)
```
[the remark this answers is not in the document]
dhruvkb: This is actually a great idea! I'll make a separate issue for that though and not try to address it in this PR.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -191,6 +191,28 @@ def __init__(
     # Helpers
     # =======
 
+    def get_stat(self, name_or_alias: str) -> tuple[bool, bool, Union[str, list[str]]]:
```
AetherUnbound, line 194: Totally not crucial, but it might be good to have this return type be a `NamedTuple`. I always get a little confused with functions returning more than two fields 😅 That way we can reference the fields under their expected names too!

dhruvkb: This is an awesome idea. I always forget that named tuples exist and my mind goes from structureless tuples to overkill data classes.

dhruvkb: Implemented in 63ee1da.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -191,7 +201,7 @@ def __init__(
     # Helpers
     # =======
 
-    def get_stat(self, name_or_alias: str) -> tuple[bool, bool, Union[str, list[str]]]:
+    def get_stat(self, name_or_alias: str) -> Stat:
```
AetherUnbound, line 204: The place where this function is called can also use the `stat` object now instead of unpacking it into 3 separate vars!

ingestion_server/ingestion_server/indexer.py
```diff
@@ -169,6 +169,16 @@ def get_last_item_ids(table):
     return last_added_pg_id, last_added_uuid
 
 
+class Stat(NamedTuple):
```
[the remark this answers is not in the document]
AetherUnbound: Ah! Okay that makes sense :)

ingestion_server/ingestion_server/indexer.py
dhruvkb: Cool, implemented in a595497.

dhruvkb:
Raising priority to 🟧 **high** as it blocks WordPress/openverse-catalog#578.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -469,3 +411,60 @@ def point_alias(self, model_name: str, index_suffix: str, alias: str, **_):
...
+        target = alias if alias is not None else f"{model_name}-{index_suffix}"
+
+        target_stat = get_stat(self.es, target)
+        if target_stat.exists:
+            if target_stat.is_alias:
+                if not force_delete:
+                    # Alias cannot be deleted unless forced.
+                    message = (
+                        f"Alias {target} might be in use so it cannot be deleted. Use"
+                        f"the `force_delete` parameter."
```
sarayourfriend, line 441: If we can give more specific instructions for how to check this too (maybe another request or a link to a document), that would be really helpful for making it as easy as possible for folks to confirm this before using force_delete.

```suggestion
                        f"Alias {target} might be in use so it cannot be deleted. Confirm "
                        " that it is not in use, then use the `force_delete` parameter."
```

ingestion_server/ingestion_server/indexer.py
```diff
@@ -469,3 +411,60 @@ def point_alias(self, model_name: str, index_suffix: str, alias: str, **_):
...
+
+            self.es.indices.delete(index=target)
+            message = f"Index {target} was deleted."
+            log.info(message)
+            slack.info(message)
+        else:
+            # Cannot delete as target does not exist.
+            message = f"Target {target} does not exist and cannot be deleted."
+            log.info(message)
+            slack.info(message)
```
sarayourfriend, line 466: Should this generate a bad request response?

ingestion_server/test/integration_test.py
```diff
@@ -267,6 +267,30 @@ def _promote(self, model, suffix="integration", alias=None):
...
+            "model": model,
+            "action": "DELETE_INDEX",
+            "callback_url": bottle_url,
+        }
+        if alias is None:
+            req |= {"index_suffix": suffix}
+        else:
+            req |= {
+                "alias": alias,
+                "force_delete": True,
```
sarayourfriend, line 281: Can you add a test for when force_delete is not passed and also when it is not True?

sarayourfriend (changes requested):
LGTM! Just one more unit test I think would be helpful.

ingestion_server/ingestion_server/indexer.py
dhruvkb: The issue with that is that the ingestion server registers requests, returns the response and then processes tasks in a separate process. So unless the task is executed, every request with a valid schema is a good request.

dhruvkb: Tasks with invalid input will eventually fail so the task status endpoint can be used after submitting to see if the task passes (and completes) or fails.

dhruvkb: I'll rescind the above statements, I've added a framework 72cc511 for tasks to communicate bad requests to the main API process, at least for the failing-fast cases where it's immediately possible to determine if the input is invalid.

ingestion_server/test/integration_test.py
dhruvkb: Added these tests cases in fb4fa02 (and got them to work 6 commits later 😆).

ingestion_server/ingestion_server/indexer.py
dhruvkb: Updated the message in 44e0d0a.

ingestion_server/ingestion_server/api.py
```diff
@@ -163,6 +200,12 @@ def on_post(self, req, res):
                 "task_id": task_id,
                 "status_check": status_url,
             }
+        elif is_bad_request.value:  # set to 1 for bad tasks
```
sarayourfriend, line 203: Should we just compare it directly against 1 instead of having to have an explanatory comment?

```suggestion
        elif is_bad_request.value == 1:
```

ingestion_server/ingestion_server/api.py
```diff
@@ -163,6 +200,12 @@ def on_post(self, req, res):
...
+            res.status = falcon.HTTP_400
+            res.media = {
+                "message": "Failed during task execution due to bad request. Check "
+                "scheduler logs."
```
sarayourfriend, line 207: Just a nit. I'm surprised the linter allows this form actually, it's quite unreadable in my opinion :confused: 

```suggestion
                "message": (
                  "Failed during task execution due to bad request. Check "
                  "scheduler logs."
                )
```

ingestion_server/test/integration_test.py
```diff
@@ -267,6 +271,43 @@ def _promote(self, model, suffix="integration", alias=None):
...
+            es.indices.get(index=f"{model}-{suffix}")
+
+    def _soft_delete_index(self, model, alias, suffix="integration", ambiguous=False):
+        req = {
+            "model": model,
+            "action": "DELETE_INDEX",
+            "alias": alias,
+        }
+        if not ambiguous:
+            req |= {"force_delete": False}
```
sarayourfriend, line 305: What does `ambiguous` mean in this context? Could the parameter be named `omit_force_delete` instead? Maybe a doc comment would help explain this. It took me a couple read throughs to understand what this test utility does.

ingestion_server/test/integration_test.py
```diff
@@ -457,3 +498,45 @@ def test_update_index_audio(self):
 
         # Wait for the task to send us a callback.
         assert self.__class__.cb_queue.get(timeout=120) == "CALLBACK!"
+
+    @pytest.mark.order(12)
```
sarayourfriend, line 502: Out of curiosity, do you know why the order mark is needed for these? My understanding was that pytest ran tests in the order that they appear in the module. Is that different for class-method based tests? My understanding is confirmed by the `pytest-order` docs as well: https://pytest-dev.github.io/pytest-order/stable/intro.html#quickstart

sarayourfriend (approved):
LGTM! Just a couple suggestions and clarification questions but nothing blocking.

ingestion_server/ingestion_server/api.py
dhruvkb: Fixed, along with the other such string in d220079.

ingestion_server/test/integration_test.py
dhruvkb: My memory is also fuzzy in that regards but, as I recall, Pytest was running them alphabetically. It'll be a simple change to remove the order and see if the tests fail (I'll do that in a separate PR after merging this).

PR #727, opened 2022-05-27
Make thumbnail link null for audio files without artwork

### LINKED ISSUE #528: Possibly make `thumbnail` null for audio files without artwork
## Description

Currently the frontend tries to fetch thumbnails for all audio files regardless of whether the audio file in question has one or not. 
I noticed that the API returns the thumbnail URL for all tracks. That makes sense, but could we improve this to be `null` for audio tracks without artwork? Then we could check the field in the frontend before making a network request.

krysal:
## Fixes

Fixes #528 by @zackkrida.
Should also fix WordPress/openverse-frontend#1475

## Description

Adds a check to the `AudioSerializer` if the underlying audio row has artwork and removes the link in the case is not found. I believe we should make sure the cover of the album (if exists) is duplicated in each track if that track does not have its own artwork. In fact, I find it rare the case where the audio will have a different cover than their album.

This also seems to be the common case with our current live data, the artwork of the `audio_set` will never be required unless you are watching at the `audio_set` directly.

```sql
-- Audio with thumbnail & audio set
 SELECT COUNT(id) FROM audio
 WHERE thumbnail IS NOT NULL AND audio_set_foreign_identifier IS NOT NULL;
 
+-------+
| count |
|-------|
| 0     |
+-------+
SELECT 1
Time: 0.179s


-- Audio with thumbnail & without audio set
 SELECT COUNT(id) FROM audio
 WHERE thumbnail IS NOT NULL AND audio_set_foreign_identifier IS NULL;
 
+-------+
| count |
|-------|
| 89138 |
+-------+
SELECT 1
Time: 0.179s


-- Audio without thumbnail & with audio set
 SELECT COUNT(id) FROM audio
 WHERE thumbnail IS NULL AND audio_set_foreign_identifier IS NOT NULL;
 
+-------+
| count |
|-------|
| 0     |
+-------+
SELECT 1
Time: 0.177s
```

## Testing Instructions

Find an audio row without a thumbnail in the database and check if is null in the `audio_detail` response.
Try with this search for the current data sample:
```
localhost:8000/v1/audio?q=zaus&filter_dead=false
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/serializers/audio_serializers.py
+++ b/api/catalog/api/serializers/audio_serializers.py
@@ -135,6 +135,17 @@ class AudioSerializer(AudioHyperlinksSerializer, MediaSerializer):
             obj = Audio.objects.get(identifier=obj.identifier)
         return obj.get_waveform()
 
+    def to_representation(self, instance):
+        # Get the original representation
+        output = super().to_representation(instance)
+
+        if isinstance(instance, Hit):
+            audio = Audio.objects.get(identifier=instance.identifier)
+            if not audio.thumbnail:
+                output["thumbnail"] = None
+
+        return output
+
 
 class AudioSearchSerializer(MediaSearchSerializer):
     """
--- a/api/catalog/api/serializers/audio_serializers.py
+++ b/api/catalog/api/serializers/audio_serializers.py
@@ -140,6 +140,7 @@ class AudioSerializer(AudioHyperlinksSerializer, MediaSerializer):
         output = super().to_representation(instance)
 
         if isinstance(instance, Hit):
+            # TODO: Remove when updating ES indexes
             audio = Audio.objects.get(identifier=instance.identifier)
             if not audio.thumbnail:
                 output["thumbnail"] = None
--- a/api/test/audio_integration_test.py
+++ b/api/test/audio_integration_test.py
@@ -81,6 +81,14 @@ def test_audio_thumb(audio_fixture):
     thumb(audio_fixture)
 
 
+def test_audio_without_thumb():
+    """The first audio of this search should not have a thumbnail."""
+    resp = requests.get(f"{API_URL}/v1/audio?q=zaus&filter_dead=false")
+    assert resp.status_code == 200
+    parsed = json.loads(resp.text)
+    assert parsed["results"][0]["thumbnail"] is None
+
+
 def test_audio_thumb_compression(audio_fixture):
     thumb_compression(audio_fixture)
 


obulat (approved):
This is a great fix for the problem that's been annoying me for so long! 🎉
I wonder if we should the check to the indexing phase, though. It seems that this will add one extra db request per audio now, wouldn't it?

krysal:
Yes, that's correct. I wanted to avoid it but seems related to Elasticsearch mapping, it doesn't return the thumbnail in the `Hit` instance so I had to look for the row in the database to solve the problem here (similar case to `peaks`), while we figure out the best way to add those non-search fields. We should reduce the number of times we look for a DB row in the future for sure.

dhruvkb:
This seems quite suboptimal performance wise. IIRC this n+1 requests was the main reason we precomputed all the waveforms.

Can we push this till we make our ES indices improvements and add thumbnails to ES `_source`?

krysal:
> This seems quite suboptimal performance wise. IIRC this n+1 requests was the main reason we precomputed all the waveforms.

I agree that is suboptimal. This is not intended to be a permanent solution but to fix the many requests we're making in the [frontend](https://github.com/WordPress/openverse-frontend/issues/1475) asking for non-existent audios' cover, as the ES indices changes got delayed by API outages. The serializer is still making a query per audio to add `peaks` so the heavy task is on the actual generation of the waveform.

> Can we push this till we make our ES indices improvements and add thumbnails to ES _source?

I'm afraid is taking too long but I'm really curious to explore the `_source` field. I hope to make time this week to explore an alternative solution.

api/catalog/api/serializers/audio_serializers.py
```diff
@@ -135,6 +135,17 @@ def get_peaks(obj) -> list[int]:
             obj = Audio.objects.get(identifier=obj.identifier)
         return obj.get_waveform()
 
+    def to_representation(self, instance):
+        # Get the original representation
+        output = super().to_representation(instance)
+
+        if isinstance(instance, Hit):
+            audio = Audio.objects.get(identifier=instance.identifier)
```
AetherUnbound, line 143: Given that we know this isn't a permanent solution, do you mind adding a `# TODO: Remove when <condition>` comment here so we can ensure this is removed when we make the ES changes?

AetherUnbound (approved):
Looks good, just one request! 🚀

PR #741, opened 2022-06-01
Revert "Refactor search controller for consistency and clarity"

sarayourfriend:
Reverts WordPress/openverse-api#699


sarayourfriend:
The deployed API is currently completely broken and unable to communciate with ES. Based on conversations I've seen from @AetherUnbound my understanding is that the root cause is this changeset, so I am reverting this and redeploying the API

PR #755, opened 2022-06-14
Limit  `page_size` to 20 for unauthenticated requests

### LINKED ISSUE #715: Ingestion of images stops unexpectedly after cleanup
## Description

All CI flakiness has boiled down to the ingestion of images stopping unexpectedly after the cleanup step. Rerunning ingestion almost surely fixes the issue.

We should investigate the issue and find out the cause. If it is incurable, we should add a step to retry the ingestion which is almost guaranteed to succeed the second time.

This also causes #562.

## Reproduction
See failed runs:
- https://github.com/WordPress/openverse-api/runs/6527207725?check_suite_focus=true#step:8:702
- https://github.com/WordPress/openverse-api/runs/6527646327?check_suite_focus=true#step:11:480

## Resolution

### LINKED ISSUE #751: Limit unauthenticated API requests to 20 page size
## Problem

Currently it's possible to send page_size up to 500. This makes it trivial for someone to DDoS the API, even if they're getting throttled.

Note: I'm marking this high instead of critical because we can't deploy the API right now anyway to fix it (although maybe this would help prevent the API issues we were seeing, something to investigate).

## Description

Limit the page_size to 20 for requests without authorization to do otherwise (which currently is none!)

## Implementation

krysal:
## Fixes

Fixes #751 by @sarayourfriend 

## Description

Adds the restriction for list views.

## Testing Instructions

Make authenticated and unauthenticated requests to try the new limit.

```sh
curl -s "http://localhost:8000/v1/images/?page_size=22"
```

You should get: `{
	"detail": "Page size must be between 0 & 20 for unauthenticated requests."
}` and for

```sh
curl -H "Authorization: Bearer <access_token>" \
  "http://localhost:8000/v1/images/?page_size=22"
```
you should get results normally.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/serializers/media_serializers.py
+++ b/api/catalog/api/serializers/media_serializers.py
@@ -7,6 +7,7 @@ from catalog.api.controllers import search_controller
 from catalog.api.models.media import AbstractMedia
 from catalog.api.serializers.base import BaseModelSerializer
 from catalog.api.serializers.fields import SchemableHyperlinkedIdentityField
+from catalog.api.utils.exceptions import get_api_exception
 from catalog.api.utils.help_text import make_comma_separated_help_text
 from catalog.api.utils.url import add_protocol
 
@@ -39,6 +40,7 @@ class MediaSearchRequestSerializer(serializers.Serializer):
         "extension",
         "mature",
         "qa",
+        "page_size",
     ]
     """
     Keep the fields names in sync with the actual fields below as this list is
@@ -104,6 +106,11 @@ class MediaSearchRequestSerializer(serializers.Serializer):
         required=False,
         default=False,
     )
+    page_size = serializers.IntegerField(
+        label="page_size",
+        help_text="Number of results to return per page.",
+        required=False,
+    )
 
     @staticmethod
     def _truncate(value):
@@ -148,6 +155,14 @@ class MediaSearchRequestSerializer(serializers.Serializer):
 
     def validate_title(self, value):
         return self._truncate(value)
+
+    def validate_page_size(self, value):
+        request = self.context.get("request", None)
+        if request.user and request.user.is_anonymous and value > 20:
+            raise get_api_exception(
+                "Page size must be between 1 & 20 for unauthenticated requests.", 401
+            )
+        return value
 
     @staticmethod
     def validate_extension(value):
--- a/api/catalog/api/serializers/media_serializers.py
+++ b/api/catalog/api/serializers/media_serializers.py
@@ -157,8 +157,9 @@ class MediaSearchRequestSerializer(serializers.Serializer):
         return self._truncate(value)
 
     def validate_page_size(self, value):
-        request = self.context.get("request", None)
-        if request.user and request.user.is_anonymous and value > 20:
+        request = self.context.get("request")
+        is_anonymous = bool(request and request.user and request.user.is_anonymous)
+        if is_anonymous and value > 20:
             raise get_api_exception(
                 "Page size must be between 1 & 20 for unauthenticated requests.", 401
             )
--- a/api/catalog/api/views/media_views.py
+++ b/api/catalog/api/views/media_views.py
@@ -60,7 +60,7 @@ class MediaViewSet(ReadOnlyModelViewSet):
             and int(raw_page_size) > 20
         ):
             raise get_api_exception(
-                "Page size must be between 0 & 20 for unauthenticated requests.", 400
+                "Page size must be between 0 & 20 for unauthenticated requests.", 401
             )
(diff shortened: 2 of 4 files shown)


api/catalog/api/views/media_views.py
```diff
@@ -52,7 +52,17 @@ def get_queryset(self):
...
-        self.paginator.page_size = request.query_params.get("page_size")
+        raw_page_size = request.query_params.get("page_size")
+        if (
+            request.user
+            and request.user.is_anonymous
+            and raw_page_size is not None
+            and int(raw_page_size) > 20
+        ):
+            raise get_api_exception(
+                "Page size must be between 0 & 20 for unauthenticated requests.", 400
```
sarayourfriend, line 63:
```suggestion
                "Page size must be between 0 & 20 for unauthenticated requests.", 401
```

401 is Unauthorized, which I think is more accurate here than "bad request" :slightly_smiling_face:

sarayourfriend (commented):
This approach looks good to me :slightly_smiling_face:

AetherUnbound (commented):
I love this approach! Initially I was thinking about just reducing any larger page size to 20, but sending a 401 back is an excellent idea 💯

api/test/auth_test.py
```diff
@@ -80,3 +78,25 @@ def test_auth_rate_limit_reporting(client, test_auth_token_exchange, verified=Fa
...
+@pytest.mark.django_db
+def test_page_size_limit_authed(client, test_auth_token_exchange):
+    time.sleep(1)
+    token = test_auth_token_exchange["access_token"]
+    query_params = {"filter_dead": False, "page_size": 21}
+    res = client.get("/v1/images/", query_params, HTTP_AUTHORIZATION=f"Bearer {token}")
+    assert res.status_code == 200
+
+    query_params = {"filter_dead": False, "page_size": 500}
+    res = client.get("/v1/images/", query_params, HTTP_AUTHORIZATION=f"Bearer {token}")
```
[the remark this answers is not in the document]
krysal: Oops! That was the intention! Thank you for pointing it out. Adding it.

dhruvkb (commented):
Can we add this validation inside the serializer using `validate_page_size`? That would be better imo as it keeps all our validation localised in one place.

krysal:
Thank you, @dhruvkb! I was looking for something like that, it is much cleaner this way ✨

api/catalog/api/serializers/media_serializers.py
```diff
@@ -149,6 +156,14 @@ def validate_tags(self, value):
     def validate_title(self, value):
         return self._truncate(value)
 
+    def validate_page_size(self, value):
+        request = self.context.get("request", None)
+        if request.user and request.user.is_anonymous and value > 20:
+            raise get_api_exception(
+                "Page size must be between 1 & 20 for unauthenticated requests.", 401
+            )
+        return value
```
sarayourfriend, line 165: The `None` default is redundant on dictionary's `get` method, but more importantly we need to actually handle the None case here or we'll end up with 500s:

```suggestion
        request = self.context.get("request", None)
        authenticated = request and request.user and not request.user.is_anonymous
        if not authenticated and value > 20:
            raise get_api_exception(
                "Page size must be between 1 & 20 for unauthenticated requests.", 401
            )
        return value
```

That seems like a reasonable and safe way to handle None. What you do think? Just treat it the same as an anonymous request?

sarayourfriend (changes requested):
LGTM! Dhruv's suggestion makes it way cleaner, kudos @dhruvkb! Just one edge case we should handle but otherwise this is g2g for me :slightly_smiling_face:

api/catalog/api/serializers/media_serializers.py
krysal: I don't think that case will happen but I like this! Though it's a bit tricky to check using the condition for the authenticated user (it becomes `None` for some reason 🤷🏻‍♀️). I modified it a bit, LMKWYT.

AetherUnbound (approved):
Looks great! Great suggestion to add this to page size validation @dhruvkb 💯

api/catalog/api/serializers/media_serializers.py
sarayourfriend: Looks good to me! I also don't foresee it happening but "something something [Murphy's Law](https://en.wikipedia.org/wiki/Murphy%27s_law)" :sweat_smile:

sarayourfriend (approved):

Issue #729, opened 2022-05-31
Invalid UUID breaks related images query


AetherUnbound:
## Description

It appears that if an invalid UUID-like string is supplied to a `/related/` URL, the UUID passes filters and an `IndexError` is raised.

```python
IndexError: list index out of range
  File "django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
  File "catalog/api/views/media_views.py", line 99, in related
    results, num_results = related_media(
  File "catalog/api/controllers/elasticsearch/related.py", line 19, in related_media
    _id = item.execute().hits[0].id
  File "elasticsearch_dsl/utils.py", line 84, in __getitem__
    l = self._l_[k]
```

Here are some examples of invalid yet UUID-like identifiers:

<details> <summary>Invalid UUIDs</summary>

- `95-9297ada227947df116f43b87f0978ae-2`
- `3975690f43-7d1a9b6613-b25492d442--fe`
- `00000000000000000000000000000000770-`
- `00000000000000000000000000000000771-`
- `00000000000000000000000000000000777a`
- `000000000000000000000000000000007779`
- `000000000000000000000000000000007775`
- `000000000000000000000000000000007773`
- `000000000000000000000000000000007771`
- `000000000000000000000000000000007770`
- `00000000000000000000000000000000666-`
- `00000000000000000000000000000000555-`
- `00000000000000000000000000000000333-`
- `00000000000000000000000000000000111-`
- `00000000000000000000000000000000000-`
- `77700000000000000000000000000000000-`
- `77700-000000000000000000000000000000`
- `00000000000000000000000000000000777-`
[...]

PR #766, opened 2022-06-22
Fix invalid UUID breaking related images query

obulat:
## Fixes

Related to #729 by @AetherUnbound 

## Description

This PR uses @krysal 's [suggestion](https://github.com/WordPress/openverse-api/issues/729#issuecomment-1163336180) to add an IndexError handling for when there are no `hits` in the search controller:
https://github.com/WordPress/openverse-api/blob/7414174a16ba4cbd6af70773ce6d19b92b0e96de/api/catalog/api/controllers/search_controller.py#L344

## Testing Instructions

On main, if you go to `http://localhost:8000/v1/images/000000000000000000000000000000000000/related/`, you would get a 500 error.
On this branch, you should get a 404.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/views/media_views.py
+++ b/api/catalog/api/views/media_views.py
@@ -116,6 +116,9 @@ class MediaViewSet(ReadOnlyModelViewSet):
             self.paginator.page_size = 10
         except ValueError as e:
             raise get_api_exception(getattr(e, "message", str(e)))
+        # If there are no hits in the search controller
+        except IndexError:
+            raise get_api_exception("Could not find items.", 404)
 
         serializer = self.get_serializer(results, many=True)
         return self.get_paginated_response(serializer.data)
--- a/api/test/api_live_integration.py
+++ b/api/test/api_live_integration.py
@@ -446,6 +446,13 @@ def test_page_consistency_removing_dead_links(search_without_dead_links):
     assert no_duplicates(ids)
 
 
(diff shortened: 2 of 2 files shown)


dhruvkb (approved):
This is a good fix. We should still tighten up the lookup regex for good measure.

obulat:
> Also, do you know why something that looks like a valid UUID (follows the pattern) doesn't cause the 500? Why doesn't it throw an IndexError as well like the same-length-but-not-UUID-pattern string does when queried for?

@sarayourfriend, actually, that's not the case, but it took me some time to figure it out :)
The example you gave in the issue (`da5cb478-c093-4d62-b721-cda18797e3fz`) looks like UUID for humans, because we don't notice that `z`(the last letter of the string) is not in the pattern, it's only `a-f` for the hex numbers.

sarayourfriend:
Ah!! I forgot that UUID had rules :sweat_smile: Good catch! Sorry for the false lead :scream:

obulat:
> Ah!! I forgot that UUID had rules 😅 Good catch! Sorry for the false lead 😱

It was actually a great opportunity to learn more about what's happening when we search for something 👍

PR #769, opened 2022-06-23
Include alt_files extensions in filtering

### LINKED ISSUE #615: Filtering by extension for audio doesn't work as expected
## Problem

When you filter audio search by extension, you don't get results that you would expect. Only `mp3` category returns any results.

## Description

To get the audio extension field in ElasticSearch, we use the URL's last part.
https://github.com/WordPress/openverse-api/blob/5c9b654a9e3ab55932de42aa76ae2c17f5e85d1d/ingestion_server/ingestion_server/elasticsearch_models.py#L136-L144

This doesn't work well for audio because many URLs don't have an extension at all. Fortunately, most of the audio files should have the `filetype` attribute saved in the database, so we could use that as the `extension` value.

Marking this as high priority because having a category filter in the UI that is not working properly is confusing for the users.

## Alternatives

Not really alternative, but additional solutions would be to use the extensions of the `alt_files` in addition to the main file extension.

## Implementation

krysal:
## Fixes

Fixes #615 by @obulat

## Description

I added Freesound records with downloadable files to test the filter with those. The audio search endpoint also returns the `filetype` of the playable audio file now.

## Testing Instructions

Check out the first commit (`d94763c1`), recreate the container to save the new data, and make a request with the filter. Notice the response is empty results, though there are some `flac` files as `alt_files`.

```
http://localhost:8000/v1/audio/?extension=flac
```
Check out the branch normally, trigger and ingestion of audio `just ingest-upstream audio` and observe the existing results.

```json
{
    "result_count": 304,
    "page_count": 16,
    "page_size": 20,
    "page": 1,
    "results": [...]
}
```
You can also try with `aiff` and `m4a`. These do not currently work in production.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -280,6 +269,19 @@ def get_aspect_ratio(height, width):
...
 
+    @staticmethod
+    def get_extension(url):
+        """
+        Get the extension from the last segment of the URL separated by a dot.
+        TODO: Use the `filetype` field once the following issue is completed:
+        https://github.com/WordPress/openverse-catalog/issues/536
+        """
+        extension = url.split(".")[-1].lower()
+        if "/" in extension or extension is None:
```
sarayourfriend, line 280: For null safety we need the None check to go first:

```
>>> '/' in None
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'NoneType' is not iterable
```

```suggestion
        if not extension or "/" in extension:
```

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -330,12 +336,22 @@ def database_row_to_elasticsearch_doc(row, schema):
...
             **attrs,
         )
 
+    @staticmethod
+    def get_extensions(filetype, alt_files):
+        if not alt_files:
+            return filetype
+
+        extensions = [filetype] + [file["filetype"] for file in alt_files]
+        return extensions
```
sarayourfriend, line 353: Given that the preview audio is not always available for downloading... I wonder if we should _only_ index alt_files for extensions when there are alt_files available (and filetypes in the future, etc)? :thinking: Might also be nice to always return a list instead of `str | list[str]`

```suggestion
    @staticmethod
    def get_extensions(filetype, alt_files):
        if not alt_files:
            return [filetype]

        return [file["filetype"] for file in alt_files]
```

ingestion_server/ingestion_server/es_mapping.py
```diff
@@ -101,6 +101,7 @@ def index_settings(table_name):
                 "fields": {"keyword": {"ignore_above": 256, "type": "keyword"}},
                 "type": "text",
             },
+            "filetype": {"type": "keyword"},
```
sarayourfriend, line 104: Can we name this `extension` so that it's clear we're not actually extracting filetype information?

sarayourfriend (commented):
Awesome! Just a few questions but otherwise this is looking great :grin:

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -330,12 +336,22 @@ def database_row_to_elasticsearch_doc(row, schema):
             category=row[schema["category"]],
             duration=row[schema["duration"]],
             length=length,
+            filetype=filetype,
```
obulat, line 339: We should probably only have the `extension` here because it is used in search as a filter (`extension=mp3`). When querying, we will only be using the `extension` field, and the `filetype` should already be returned as part of the `source` object when #684 is merged.

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -280,6 +269,19 @@ def get_aspect_ratio(height, width):
...
         return aspect_ratio.lower()
 
+    @staticmethod
+    def get_extension(url):
+        """
+        Get the extension from the last segment of the URL separated by a dot.
+        TODO: Use the `filetype` field once the following issue is completed:
+        https://github.com/WordPress/openverse-catalog/issues/536
+        """
+        extension = url.split(".")[-1].lower()
```
obulat, line 279: I think 
```suggestion
        if filetype:
            return filetype
        extension = url.split(".")[-1].lower()
```

ingestion_server/ingestion_server/elasticsearch_models.py
```diff
@@ -280,6 +269,19 @@ def get_aspect_ratio(height, width):
             aspect_ratio = Image.AspectRatios.SQUARE.name
         return aspect_ratio.lower()
 
+    @staticmethod
+    def get_extension(url):
```
obulat, line 273:
```suggestion
    def get_extension(filetype, url):
```

obulat: We would also need to change line 231 above to `extension = Image.get_extension(row[schema["filetype"]], row[schema["url"]])` (adding the filetype as a parameter).

krysal: Why these changes? I wanted to keep image extension changes minimal on this PR.

ingestion_server/ingestion_server/elasticsearch_models.py
krysal: I tried returning always a list but when there is only a single extension it produces a parsing error.
I'm fine with including the preview's file type in the indexing. If more folks think we should not I'll exclude it.

ingestion_server/ingestion_server/elasticsearch_models.py
obulat: I agree, it makes sense to keep the PR small, @krysal :)

ingestion_server/ingestion_server/es_mapping.py
krysal: The thing is that the extension(s) is still the compilation of filetypes so here, this will still represent the filetype of the preview or the single audio file. The Audio model is keeping the `filetype` field name, are we going to change that too? (question for @obulat and the rest of @WordPress/openverse-maintainers) I believe that deserves its own separate discussion because it would be better to be consistent between the Audio and Image models.

ingestion_server/ingestion_server/elasticsearch_models.py
krysal: It is necessary to have it here now for having the `filetype` appearing in the search results instead of always `null`, it can be changed later in that PR if found unnecessary. Please follow [this comment](https://github.com/WordPress/openverse-api/pull/769#discussion_r906151591) about the filetype/extension discussion.

ingestion_server/ingestion_server/elasticsearch_models.py
sarayourfriend: > I tried returning always a list but when there is only a single extension it produces a parsing error.

That's strange :thinking: I wonder what the difference could be...

> I'm fine with including the preview's file type in the indexing. If more folks think we should not I'll exclude it.

The problem is I think it might be inaccurate to include it unless we can confirm it's actually a downloadable audio track (rather than just a preview audio)... I kind of wish this list was defined in the catalog rather than munged in the API :disappointed: That way it could be per-provider and we could confirm that the list is actually the list of downloadable formats.

ingestion_server/ingestion_server/es_mapping.py
sarayourfriend: I feel like we keep kicking this conversation down the road though and I'm not sure where to have it or how to come to a resolution on it. We have discussed it in a variety of places and come down to this completely quiet issue: https://github.com/WordPress/openverse-catalog/issues/478

I'm concerned that the way we're going now, of putting this information into ES the way we ware or doing it in the API or anything like that without first evaluating the consequences of it and how we want to ideally do it in the future, we'll lock ourselves into a bad situation and have our audio v1 API full of messy, incongruous concepts... I could be being overly cautious though. And I don't want to stall this work which will improve the existing non-ideal implementation.

I'm hesitant to just keep pushing these discussions down the road though and wonder what the best way to prioritize it because it keeps coming up over and over again.

krysal: The way I see it "filetype" and "extension" are pretty much synonymous in our case. So perhaps changing the filter param name (from `extension` to `filetype`) would make the situation clearer? That name is from CC times. The other way seems that is much more work and I'm not sure if it's worth it.

> We have discussed it in a variety of places and come down to this completely quiet issue: https://github.com/WordPress/openverse-catalog/issues/478

That's unfortunate, that issue was off my radar so sadly I don't think it will enter the milestone. My vote would be to ditch that field at this point, we're not using it.

> I'm concerned that the way we're going now, of putting this information into ES the way we ware or doing it in the API or anything like that without first evaluating the consequences of it and how we want to ideally do it in the future, we'll lock ourselves into a bad situation and have our audio v1 API full of messy, incongruous concepts...

I have exactly the same thoughts when it comes to adding new media types. I don't exactly have the answer to this problem either, other than to prioritize at some point, and hope that new types of media are better planned with the lessons learned from audio.

ingestion_server/ingestion_server/elasticsearch_models.py
krysal: I thought I had answered this, in my mind I wrote that I only focused on changes needed for Audio here 😅

ingestion_server/ingestion_server/es_mapping.py
sarayourfriend: No, specifically the discussion related to the issue I linked, that is, metadata extraction from files during ingestion (or during some other stage in our data pipeline, whenever it may be). It is related to the waveform discussion as well as waveform data is a form of metadata and requires us to download the files to produce it.

I'm not advocating for needing to decide and implement this now. I'm just worried that if we don't at least have _some_ idea of how it might happen or indeed if it will happen at all, then we'll do things like mix together filetype and extension. If the project decided it didn't want to do this then at least we could just ignore the question entirely and do whatever is easiest for the moment. Right now, however, it's just hanging over these metadata decisions. Some of these decisions could lock us into APIs we wouldn't want in the future if we do decide to do something different with metadata.

krysal: Actually, I thought for a moment to follow that path of grouping extensions too, as @obulat did something similar for [image extensions in the catalog](https://github.com/WordPress/openverse-catalog/pull/545) and I found that .ogg and .oga can be considered the same filetype but then filetype for images are limited to one media file, it's not the same case for audio.

Honestly, I'm not very familiar with the similarity between many filetypes/container formats, and from what I see there seems to be more variety of containers for audio formats. So I thought it may be actually better to have the finer grain filter, if someone wants the whole group it could simply pass all the extensions and it would be the same result.

I agree that the `bitrate` and the `waveform` will probably follow a similar approach, and I want to investigate it, but I believe `filetype` can be as simple as this, given we already have that data in the catalog (it's [confirmed](https://github.com/WordPress/openverse-catalog/issues/559#issuecomment-1164343209) that all the audio files have this field).

sarayourfriend (approved):
LGTM! Should we add a unit test for the `Audio::get_extensions` method? Or is it already tested inherently through the integration tests?

obulat (approved):
Great to finally get this solved, @krysal !

PR #776, opened 2022-06-29
Add more logging to ES query builder

sarayourfriend:
## Fixes

Fixes #775  by @sarayourfriend 

## Description

Adds request IDs to all logs so that we can trace them easily throughout a request without having to rebuild it ourselves (we can simply query for the request ID and then follow a request's logs in their entirety).

Also begins the process of adding logs to the search_controller around the pagination logic. I'm stumped by this branch though and am unsure how to describe it in the logs:

https://github.com/WordPress/openverse-api/blob/f823fb782f9713a7fa5ac91294cba6fe7ca8d026/api/catalog/api/controllers/search_controller.py#L69-L80

@dhruvkb @krysal @obulat do any of y'all have any ideas for this? Part of the problem is I'm unclear on how the link mask works and in particular this call is confusing the heck out of me:

https://github.com/WordPress/openverse-api/blob/f823fb782f9713a7fa5ac91294cba6fe7ca8d026/api/catalog/api/controllers/search_controller.py#L70

I have no idea what it's meant to do! calling `list(accumulate(integer_list))` basically builds out a factorial-ish kind of list as a result...

```
>>> list(accumulate([1, 2, 3, 4, 5]))
[1, 3, 6, 10, 15]
```

I have absolutely _zero_ clue about what this is meant to be doing for us here. If anyone can help explain this then I will also add some comments to this bit of code (and probably write a doc page about the query masking in general). It's quite cryptic without any context aside from the code!

## Testing Instructions

TBD

## Checklist

- [ ] My pull request has a descriptive title (not a vague title like `Update index.md`).
- [ ] My pull request targets the *default* branch of the repository (`main`) or a parent feature branch.
- [ ] My commit messages follow [best practices][best_practices].
- [ ] My code follows the established code style of the repository.
- [ ] I added or updated tests for the changes I made (if applicable).
- [ ] I added or updated documentation (if applicable).
- [ ] I tried running the project locally and verified that there are no visible errors.

[best_practices]:https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

## Developer Certificate of Origin

<details>
<summary>Developer Certificate of Origin</summary>

```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


[...]

PR #778, opened 2022-06-29
Refactor search controller for consistency and clarity

dhruvkb:
This is a revision of PR #699 formed by reverting commit 7e009fd74a26f9c707fa09d82638a6077a88ffa7 and then rebasing against `main`.

## Original description

> This PR refactors the search controller, splitting it up into smaller, easier to understand chunks with proper documentation and type hinting.
>
> This PR 
> - removes frivolous arguments to search, using the `MediaSearchRequestSerializer` as the single source of truth for search params
> - applies the excluded provider check only when there are non-zero excluded providers
> - updates the `get_sources` function to use DSL instead of raw query to ES
> - updates usage of `Search` to be idiomatic with Elasticsearch DSL conventions
> - moves globally accessible ES connection to `settings.ES`
> - reduces code duplication

## Testing Instructions

This is a refactor, passing CI should be sufficient proof that no functionality is impacted.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+) -> Search:
+    """
+    Parse and apply a filter from the search parameters serializer. The
+    parameter key is assumed to have the same name as the corresponding
+    Elasticsearch property. Each parameter value is assumed to be a comma
+    separated list encoded as a string.
+
+    :param s: the search query to issue to Elasticsearch
+    :param query_serializer: the ``MediaSearchRequestSerializer`` object with the query
+    :param basis: the name of the field in the serializer and Elasticsearch
```
sarayourfriend, line 53: nit/clarification request: where does the name `basis` come from/what does it mean in this context? I understand how the parameter is used (and its necessity) but I do not understand the name basis or how it relates to the purpose.

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+    index: Literal["image", "audio"],
+    ip: int,
+) -> Tuple[List[Hit], int, int]:
+    """
+    Perform a ranked, paginated search based on the query and filters given in the
+    search request.
+
+    :param query_serializer: the ``MediaSearchRequestSerializer`` object with the query
+    :param index: The Elasticsearch index to search (e.g. 'image')
+    :param ip: the users' hashed IP to consistently route to the same ES shard
```
sarayourfriend, line 84: This is interesting, I didn't know we did this. I wonder if circumventing ES's built-in load balancing could be an issue for us.

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -0,0 +1,198 @@
...
+    :return: the list of search results with the page and result count
+    """
+
+    s = Search(using="default", index=index)
+    search_params = query_serializer.data
+
+    rules: dict[Literal["filter", "exclude"], list[Union[str, tuple[str, str]]]] = {
```
sarayourfriend, line 91: nit: With such a specific structure it seems like maybe a group of dataclasses configured in this fashion could be (a) more robust in the face of future refactors and (b) clearer for someone adding a new aspect to this configuration. It'd also give each thing a clearer and individual name (and a docstring explaining why the name):

```py
from dataclasses import dataclass, field


@dataclass
class Basis:
    """https://stackoverflow.com/a/61480946"""
    name: str
    _field: Optional[str] = field(init=False, repr=False, default=None)
    
    @property
    def field(self) -> str:
	return self._field or self.name
    
    @field.setter
    def field(self, value: Optional[str] = None) -> None:
	self._field = value
	

@dataclass
class Rules:
    filter: list[Basis]
    exclude: list[Basis]
    
    def get(self) -> Mapping[Literal['filter', 'exclude'], list[Basis]]:
	return {
	    "filter": self.filter,
	    "exclude": self.exclude
	}


rules = Rules(
    filter=[
	Basis("title"),
	Basis("license", "license.keyword"),
	# ... etc
    ],
    exclude=[
	Basis("excluded_source", "source"),
    ]
)


for behaviour, bases in rules.get():
    for basis in bases:
	s = _apply_filter(s, query_serializer, basis, behaviour)
```

It would give a natural (and central/easy to find) place to explain or cite the origin of the name "basis" as well.

sarayourfriend (approved):
As I was last time, I am still uncomfortable with the size and scope of this refactor. Even though I do agree it seems to make things much more uniform and consistent. I'm not sure that for me it makes anything clearer; perhaps for someone already familiar with the code it does though!

I left one suggestion that might help clarify at least a small part of the code. I think a lot of the code is pretty cryptic to begin with and I hope we can refactor/add tests/add comments to help with that in the future.

If there is absolutely _any_ way to break up this refactor and to deploy it in steps, I think that would be a benefit, but I can appreciate that that is not always possible.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: I don't know enough to comment if it can be a problem, but I can see how redirecting all requests from a particularly active consumer to a single shard can be a bottleneck. I haven't modified it in this PR though, it's been that way from before.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: `basis` was a bad name for the variable. I've replaced that with a much clearer name 'FieldMapping' in 0c86c03.  The new name clearly states that this is a mapping between fields in the `MediaSearchRequestSerializer` and the corresponding field in the ES index.

api/catalog/api/controllers/elasticsearch/search.py
dhruvkb: Using a data-class for `FieldMapping` (`Basis` in your example) was a bit too verbose so I used a regular class for that purpose. Hopefully the new code in 0c86c03 is much easier to grasp.

dhruvkb:
I'm going to merge this so that we can unblock #776. It's very huge as is so any changes to this can be addressed in faster, smaller, easier-to-review PRs.

PR #694, opened 2022-05-11
 Add command to resend verification emails

sarayourfriend:
## Fixes

Fixes #689 by @AetherUnbound  

## Description

Adds a new command to run for resending verification emails. Note: I haven't written unit tests for this but I need to stop working on this for the day and come back to it tomorrow. Just wanted to get a draft up early.

## Testing Instructions

I tried to write comprehensive unit tests. If you notice a missing branch please let me know so I can add additional tests.

You can test this on your local API (though I'm hoping the unit tests are sufficient) by setting the email backend to SMTP in your API settings without filling in the SMTP server information and then making some registration requests with multiple application names. Use a mix of the same email and different emails. This will cause the verifications to be completed, but because we haven't filled in the SMTP server information, it will fail to send the email (which is what was happening in production).

Then switch your settings back to the Console email backend and run the management command. You should see some helpful logs throughout the way, but also you'll see the emails that _would_ be sent pop into the console. It will look like this:

```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Verify your API credentials
From: 
To: user@example.com
Date: Thu, 12 May 2022 09:20:50 -0000
Message-ID: <165234725049.145.11337592980239670284@fa90ab837821>


The Openverse API OAuth2 email verification process has recently been fixed.
We have detected that you attempted to register an application using this email.

To verify your Openverse API credentials, click on the following link:

https://api.openverse.engineering//v1/auth_tokens/verify/<token redacted>

If you believe you received this message in error, please disregard it.

-------------------------------------------------------------------------------
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Verify your API credentials
From: 
To: user@example.net
Date: Thu, 12 May 2022 09:20:50 -0000
Message-ID: <165234725049.145.2282716428617802226@fa90ab837821>


The Openverse API OAuth2 email verification process has recently been fixed.
We have detected that you attempted to register an application using this email.

To verify your Openverse API credentials, click on the following link:

https://api.openverse.engineering//v1/auth_tokens/verify/<token redacted>

If you believe you received this message in error, please disregard it.

-------------------------------------------------------------------------------
The following applications had email verifications sent.                                         

Application name: My amazing project
Cleaned related application count: 2
Cleaned related verification count: 0
Cleaned related registration count: 2


Application name: My amazing project5
Cleaned related application count: 0
[...]

PR #779, opened 2022-06-30
Add back and correct "resendoauthverification" command

### LINKED ISSUE #770: Sent API registration emails contained a typo
We recently used a one-off script to send emails to users who had registered for the Openverse API during a time period in which verification emails were broken. Unfortunately, the emails we sent contained broken link with a url with two slashes, like so:

```
                      error here: ↓
https://api.openverse.engineering//v1/auth_tokens/verify/{mytoken}
```

We should most likely:

1. Reintroduce the one-off script (it was removed in this PR: https://github.com/WordPress/openverse-api/pull/768)
2. Fix the duplicated slash
3. **Bonus**: The emails were shown to be sent from 'noreply'; ideally we could set a name like 'Openverse API Team' or something in the code. Here's how it looked in Gmail:
  CleanShot 2022-06-24 at 12 42 30@2x

Edit: As @sarayourfriend indicated, we can identify the users we emailed in production under the `resendoauthverification:processed` redis key.

sarayourfriend:
Fixes #770 by @zackkrida 

## Description
Corrects the email. I took a shot at the text of the corrected email but please feel free to offer better wording if you can think of any!

## Testing instructions
Follow instructions from the original PR for testing locally: https://github.com/WordPress/openverse-api/pull/694

Note also that CI should pass.

### DIFF
--- a/api/catalog/management/commands/resendoauthverification.py
+++ b/api/catalog/management/commands/resendoauthverification.py
@@ -1,0 +1,225 @@
+import argparse
+from dataclasses import dataclass
+
+from django.conf import settings
+from django.core.mail import send_mail
+from django.db import transaction
+from django.db.models import Q
+from rest_framework.reverse import reverse
+
+from django_redis import get_redis_connection
+from django_tqdm import BaseCommand
+
+from catalog.api.models.oauth import (
+    OAuth2Registration,
+    OAuth2Verification,
+    ThrottledApplication,
+)
+
+
+def get_input(text):
+    """
+    Wrapped ``input`` to allow patching in unittests
+    """
+    return input(text)
+
+
+verification_msg_template = """
+The Openverse API OAuth2 email verification process has recently been fixed.
+We have detected that you attempted to register an application using this email.
+
+To verify your Openverse API credentials, click on the following link:
+
+{link}
+
+If you believe you received this message in error, please disregard it.
+"""
+
+
+@dataclass
+class Result:
+    saved_application_name: str
+    deleted_applications: int
+    deleted_registrations: int
+    deleted_verifications: int
+
+
+class Command(BaseCommand):
+    help = "Resends verification emails for unverified Oauth applications."
+    """
+    This command is meant to be used a single time in production to remediate
+    failed email sending.
+
+    It stores a cache of successfully sent emails in Redis, so running it multiple
+    times (in case of failure) should not be an issue.
+    """
+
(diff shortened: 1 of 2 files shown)


zackkrida (approved):
LGTM! The copy is good, thank you.

zackkrida:
@obulat and @krysal can you review this today or tomorrow so we can resend these emails this week?

krysal (approved):
I ran the `dry` version first and then the one with the `--no-dry_run` flag and I got different outputs. Is this expected?

Running `just dj resendoauthverification`:

```
The following applications had email verifications sent.                                                                                              

Application name: My test project
Cleaned related application count: 3
Cleaned related verification count: 3
Cleaned related registration count: 3


Application name: My test project2
Cleaned related application count: 0
Cleaned related verification count: 0
Cleaned related registration count: 0


Application name: My test project3
Cleaned related application count: 1
Cleaned related verification count: 1
Cleaned related registration count: 1

The above was a dry run and no records were actually affected.
```

Running `just dj resendoauthverification --no-dry_run`:

```
The following applications had email verifications sent.                                                                                              

Application name: My test project
Cleaned related application count: 3
Cleaned related verification count: 0
Cleaned related registration count: 3


Application name: My test project2
Cleaned related application count: 0
Cleaned related verification count: 0
Cleaned related registration count: 0


Application name: My test project3
Cleaned related application count: 1
Cleaned related verification count: 0
Cleaned related registration count: 1
```

I verified the link sent in the email and it works. It would be nice to see a unit test for it since the actual error was there. I'm less concerned as this has been already run in production, so approving to unblock.

sarayourfriend:
I do not know why that discrepancy exists. Reading through the code it doesn't seem like it should, the counts for each thing happen independent of the deletions and deleting one thing or another doesn't make the verifications query work different, at least not in any way that is evident to me by reading the code...

> It would be nice to see a unit test for it since the actual error was there

I started to add a unit test that actually made a request with the URL... but I wasn't sure how to do that without having to spin up an API :thinking: If we use the RequestFactory approach it wouldn't _really_ be all that useful, I don't think. We could assert in a regex or something that there are no double `/`s. Would that work for you?

krysal:
> We could assert in a regex or something that there are no double /s. Would that work for you?

That sounds good!

sarayourfriend:
Added the unit test. If you put back the extra `/` they will fail. Thanks for the suggestion, Krystle!

PR #790, opened 2022-07-06
Add logging and fix all logging contexts

sarayourfriend:
## Fixes

Fixes #776 by @sarayourfriend 

## Description

Add logging to the elasticsearch controller methods.

I also went ahead and updated the logging contexts for all our existing logging in the Django app. To learn more about how this works check out the [`Logget.getChild` docs](https://docs.python.org/3/library/logging.html#logging.Logger.getChild) for more insight into how this works. These changes are necessary to make it easy to discern the context of the logs without having to repeat some context sigil (like the name of the method) in the log messages. This simplifies the code that adds to the log and also ensures a uniform way of filtering logs by context (i.e., method and module name) across the app.

I also added a documentation page describing the approach to logging and the reasoning behind it. Please let me know if any more details should be added here or if the reasons are unclear or otherwise do not make sense.

## Testing Instructions

Checkout this branch. Update your local env to have `LOG_LEVEL` set to `"DEBUG"`. Run the app locally and then run `just logs web`. Test a few endpoints and confirm that the logging works.

Review the added logs and ensure that they are clear and have sufficient context to understand where they're being logged within a method. Remember that the child logger approach ensures that we always know what module and method we're in. We just need to add context (using the code as a map for this context, by the way) enough to be able to tell what's going on from the logs (again, cross referencing with the code). Confirm that I didn't miss any important branches (ideally I've covered all branches except trivial ones).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/controllers/elasticsearch/utils.py
```diff
@@ -55,26 +64,71 @@ def paginate_with_dead_link_mask(
...
+            f"page_size={page_size} "
+            f"page={page} "
+            f"start={start} "
+            f"end={end} "
+        )
+        return start, end
+
+    # TODO: What does this request_size mean vs the one below
+    # that doesn't subtract 1 from the page? Why do we subtract
+    # 1 here and the other time we do not?
```
stacimc, line 90: The refactoring here looks good to me in that it preserves the original logic, but I agree that I am also confused 😄 Excited to have the logs to allow for investigation.

api/catalog/api/controllers/elasticsearch/utils.py
```diff
@@ -129,10 +198,18 @@ def post_process_results(
                 return results
 
             s = s[start:end]
+            logger.info(f"executing additional filler query depth={depth}")
```
stacimc, line 201: I initially thought this was a typo and `filler` was meant to be `filter`, but I realize now this is probably using "fill" to refer to the process of `"continue accruing additional results until the requested page_size of validated results is reached"` described above. Is that correct?

api/catalog/api/controllers/elasticsearch/utils.py
```diff
@@ -148,13 +225,22 @@ def get_result_and_page_count(
...
+    logger = parent_logger.getChild("get_result_and_page_count")
     result_count = response_obj.hits.total.value
     natural_page_count = int(result_count / page_size)
     if natural_page_count % page_size != 0:
+        logger.debug("incrementing natural_page_count")
         natural_page_count += 1
     last_allowed_page = int((5000 + page_size / 2) / page_size)
     page_count = min(natural_page_count, last_allowed_page)
     if len(results) < page_size and page_count == 0:
+        logger.debug("setting result_count to len(results)")
```
stacimc, line 237:
```suggestion
        logger.debug(f"setting result_count to {len(results)}")
```

stacimc (changes requested):
This is going to be so helpful ✨ The inclusion of the new doc is fantastic, and the approach here with the logging context is also a major improvement.

I tested a few endpoints to try to generate and verify as many of the logs as I could, except the audio waveforms. I only encountered the one error when searching by `source`.

api/catalog/api/controllers/elasticsearch/utils.py
sarayourfriend: Yes! I was struggling to find a word to describe the process. "backfill" didn't seem right to me, but maybe it technically is what we're doing: https://www.thefreedictionary.com/backfill

> to refill an excavated trench, esp (in archaeology) at the end of an investigation

More or less what we're doing here lol!

api/catalog/api/controllers/elasticsearch/utils.py
sarayourfriend: Oh actually this was intentionally _not_ formatted, mostly because we print `result_count` below so this information might be redundant (and thus introduce an unnecessary call to `len` for whatever that's worth).

It seems silly to think so hard about this now though, I'll update this with your suggestion (but formatted with the `name=value` way)

api/catalog/api/controllers/elasticsearch/search.py
```diff
@@ -68,13 +80,19 @@ def _apply_filter(
...
     if mapping.serializer_field in search_params:
         filters = []
         for arg in search_params[mapping.serializer_field].split(","):
             filters.append(Q("term", **{mapping.es_field: arg}))
         method = getattr(s, behaviour)  # can be ``s.filter`` or ``s.exclude``
+        logger.debug(
+            "applying filter "
+            f"behaviour={behaviour} "
+            f"mapping={mapping} "
+            f"filters={json.dumps(filters)} "
```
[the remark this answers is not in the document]
sarayourfriend: Fix is up!

stacimc (approved):

zackkrida (approved):

sarayourfriend:
@zackkrida that's actually incomplete. If you change the log level to `DEBUG` you'll get even more information.

zackkrida:
@sarayourfriend ah yep, I set the log level on the ES service instead of the web service. Re-testing now!

zackkrida:
Sorry about that! I checked out a different branch before generating the outputs.

yeah, it's working 😅 

CleanShot 2022-07-07 at 14 29 02@2x

sarayourfriend:
Thanks!

PR #792, opened 2022-07-11
Remove bespoke throttle exemption

### LINKED ISSUE #585: Remove bespoke throttle exemption classes and instead create a separate oauth "unlimited" scope
## Problem

I came up with the exemption class solution out of ignorance of the actual solution, which was to create a new OAuth2 scope to apply to tokens that would use an unlimited throttle.

## Description

Remove the exemption classes scheme and implement the unlimited scope as a new oauth scope and apply the scope to the relevant API keys.

Note: first apply the scope the API keys in production that need it _then_ deploy the change to add the new scope. Because of the existing exemption class approach, the scope is ignored on the key so it can be safely changed without affecting the existing throttling behavior.

## Alternatives

Keep the extra code (but why maintain that when there's an existing solution for it?)

## Implementation

sarayourfriend:
## Fixes

Fixes #585  by @sarayourfriend 

Unblocks https://github.com/WordPress/openverse-frontend/issues/1558

## Description

Removes the throttle exemption classes and switches to just using the scoped OAuth tokens. While that is also bespoke, it already exists and this allows us to delete a substantial amount of hot-path code :slightly_smiling_face:

I wrote some documentation to describe the state of the throttling and authentication world. Please read that, then read this section unless you're already familiar with concepts like rate limit models and throttle scopes (and how they're related).

This PR adds a new rate limit model of "exempt" and ties it to a scope with `None` throttling configuration, which is the same as setting it to infinity. This truly exempts the requester from all rate limiting, just like how the frontend is currently exempting itself via IP address.

In the interest of now having to coordinate a complicated deployment, I've left the IP based throttling in for now.

Before deploying this PR we _MUST_ update the WordPress.com/Jetpack application to have the exempt rate limit model in order to maintain the current behavior of that application.

## Testing Instructions

Note that in addition to below I've also added comprehensive unit tests to the throttling module. If you notice anything missing, please let me know so I can add it.

Please also review the documentation I added and let me know if it needs any clarification.
To test this locally, I recommend using the load testing script from this PR: https://github.com/WordPress/openverse/pull/257

Update settings.py to:
1. Enable throttling
2. Add your docker host address to ALLOWED_HOSTS. This will either be `host.docker.internal` on Docker Desktop or `172.17.0.1` otherwise.

Next create an application and token by following the authentication flow on your local API. Open up Django admin to be able to move the application between the different rate limit models.

I would start by first testing anonymous rate limiting by putting a bogus api key in the load testing script. Confirm that you're quickly rate limited.

Next, use your actual key. Fiddle with the `api.sh` variables of requests and concurrency until you surpass the rate limit. This is testing the "standard" model.

Move the application into the "enhanced" model and do the same. You should be able to overwhelm this throttle by setting the api load testing script to make a little over 200 requests, maybe slightly more due to throttle race conditions. Use 300 and you'll definitely see it rate limit.

Now move the application into the "exempt" model and repeat. You can make the number of requests as high as you want, but you'll never get throttled.

Add your local IP to the `ip-whitelist` in your local redis. I don't know what the best way to find this IP is, it might be different per computer. I just put a breakpoint in a request and check to see what the IP of the request is from inside the API.

Run `just exec cache bash` then `redis-cli`. Add your IP to `ip-whitelist` by running:

```
> sadd "ip-whitelist" "<ip>"
```

Note the quotes are significant here!

Without restarting your API (because the cache is not persistent between restarts on local), make the same anonymous api load testing requests and confirm that you are likewise never throttled.

We will remove this final behavior in a subsequent PR once the frontend has been migrated to use API keys to circumvent the rate limiting.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


justfile
```diff
@@ -53,6 +53,9 @@ recreate:
 logs services="" args=(if IS_CI != "" { "" } else { "-f" }):
     docker-compose {{ DOCKER_FILE }} logs {{ args }} {{ services }}
 
+attach service:
```
dhruvkb, line 56: Would be nice to have a comment that describes the recipe.

api/docs/reference/api/authentication_and_throttling.md
```diff
@@ -0,0 +1,33 @@
...
+
+The `ThrottledApplication` is an extension of the base `AbstractApplication` class from `django-oauth-toolkit` that adds an additional concept of a `rate_limit_model`. Each rate limit model can be thought of as a "tiers" of rate limiting that the Openverse API supports.
+
+## Rate limit models (tiers)
+
+The following tiers exist:
+
+* `standard`: The default rate limit that anyone is able to unlock simply by authenticating. Slightly higher and much more useful limit for an application.
+* `enhanced`: A selectively granted rate limit that allows for significantly higher burst and sustained requests.
+* `exempt`: A rate limit model that exists purely to accommodate internal infrastructural needs. For example, the server that renders the frontend can use tokens of this tier to prevent itself from being rate limited by the API during server side rendering. Likewise, when we need to circumvent the throttling to do load testing, we use a key of this tier.
```
dhruvkb, line 23: These tiers are so well-named.

dhruvkb (approved):
Looks good to me!

obulat:
I'm having trouble initializing the API:
```
ModuleNotFoundError: No module named 'log_request_id'
```

dhruvkb:
@obulat you need to rebuild the `web` image:

```shell
$ docker-compose build web
```

api/catalog/api/utils/oauth2_helper.py
```diff
@@ -24,18 +24,22 @@ def get_token_info(token: str):
...
+    if expired:
+        logger.info(
+            "rejected expired access token "
+            f"application.name={application.name} "
+            f"application.client_id={application.client_id} "
+        )
         return None, None, None
+
+    client_id = str(application.client_id)
+    rate_limit_model = application.rate_limit_model
```
AetherUnbound, line 43: If the assignment of `application` on L29 fails, won't these lines fail as well? Did we want to re-raise after the except on L30 or return early?

api/docs/reference/api/authentication_and_throttling.md
```diff
@@ -0,0 +1,33 @@
+# Authentication and Throttling
+
+This document primarily serves to document our additions to DRF's built-in throttling support. In order to accomplish this, some basic facts about our OAuth flow and configuration must also be explained.
+
+This document deals primarily with the server's perspective of the OAuth flow and assumes you are familiar with the user's perspective. If you are not familiar with the user's perspective, please study the [Register and Authenticate](https://api.openverse.engineering/v1/#section/Register-and-Authenticate) section of the API documentation.
+
+Furthermore, this section assumes at least a surface level understanding of the `django-oauth-toolkit` library and its concepts. In particular, it is necessary to understand the [`Application` concept, the docs for which can be found here].
```
AetherUnbound, line 7: Looks like a link might be missing here.

api/catalog/api/utils/oauth2_helper.py
sarayourfriend: Oops, yes, I meant to return the `None, None, None` tuple.

sarayourfriend: Fixed in https://github.com/WordPress/openverse-api/pull/792/commits/0e93aec3a2522b1f0e57eec1f6c22a449458b08c

AetherUnbound (approved):
I wasn't able to get the testing steps to work locally, even when I set my application to `exempt` I was still getting throttled with that key 😕 I feel like I must be doing something wrong, but I'm not sure where. I'm confident in the code though, and I know this is currently blocking. If someone else was able to get the testing steps working please let me know!

sarayourfriend:
To verify some things, @AetherUnbound, as it concerns me that you weren't able to get it to work.

Did you run migrations (`just dj migrate`)? Did you try recreating + init?

I left out the instructions to run the migration from the PR description, sorry if that is the issue!

AetherUnbound:
I did run the migration, but that was after I made the application 🤔 do you think that would have had an effect? I'll try recreate + init!

sarayourfriend:
I'm not sure! It would be a bad sign if it did cause a problem as that would be one of the situations we'll run into when we migrate some of the existing "exempted" applications :thinking:

sarayourfriend:
Which reminds me... we will have to be careful when we deploy this PR. We'll have to run the migrations, then update the rate_limit_model for currently exempted applications, then deploy the rest of the PR.

Interesting to note that with automated deployments we won't be able to follow this manual process... in those cases we'd have to split this PR into two:

1. First a part for the migration;
2. then, the rest of the PR with the refactor and code removal

We are still months from that scenario but it is something for us to start thinking about now so we're not caught off guard when the time does come.

AetherUnbound:
Ah okay, I got it working! I also had to set the throttled application as "verified". One other thing to note though, the Admin UI required that I set a value for `Client type` before I could change the exception level. Is that an issue?

sarayourfriend:
It's not an issue, we can just set it to public I guess? Or we could fix it in this PR and include a data migration to update existing applications to have public set explicitly.

I'm not sure why it's able to be empty :thinking: The column is configured like this:

```
    client_type = models.CharField(max_length=32, choices=CLIENT_TYPES)
```

https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L106

Kind of strange, no?

obulat:
What is the correct path for running migrations? I couldn't pass the API healthcheck, so I ran `just recreate` (I also deleted all the images and cache to be sure). When I try to run `just init` after that, I get the `relation doesn't exist` error because the migrations weren't applied. So, I tried recreating again, this time running `just dj migrate` before running `just init`. But it still didn't work, the API healthcheck failed.
I did manage to run the migrations using Docker desktop to open the API container shell. Then I got throttled on the healthcheck requests 😆  (because `just init` was still running and querying that endpoint)

sarayourfriend:
@obulat can you run `just logs web` and see if there is an error in the API causing the healthcheck to fail?

obulat:
> @obulat can you run `just logs web` and see if there is an error in the API causing the health check to fail?

I can no longer reproduce the API error; it works well after I ran the migrations inside the web shell. But it was something about missing relations in the database.

AetherUnbound:
> It's not an issue, we can just set it to public I guess? Or we could fix it in this PR and include a data migration to update existing applications to have public set explicitly.
> 
> I'm not sure why it's able to be empty thinking The column is configured like this:
> 
> ```
>     client_type = models.CharField(max_length=32, choices=CLIENT_TYPES)
> ```
> 
> https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L106
> 
> Kind of strange, no?

I checked this again on #808 (which is off of `main`), the exact same thing happened there! So it's nothing related to this branch, and it seems we've been working with that quirk without issue 🙂

PR #801, opened 2022-07-15
Add trailing slashes to URL conf paths

dhruvkb:
## Description

This PR adds trailing slashes to two URLs that did not have those. This makes them more consistent with the rest of the URLs in the codebase and also makes it possible to call them from the frontend (which mandatorily suffixes a slash to API calls).

It will not affect any existing users as `APPEND_SLASH` setting should ideally redirect them to the right address.

## Testing Instructions

0. Check out the branch and run the dev server `just up`.
1. Try visiting the changed URLs (with and without the slash).
2. Observe redirects when you visit the URL without the slash.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/views/oauth2_views.py
+++ b/api/catalog/api/views/oauth2_views.py
@@ -67,7 +67,7 @@ curl \\
   -X POST \\
   -H "Content-Type: application/json" \\
   -d '{"name": "My amazing project", "description": "To access Openverse API", "email": "user@example.com"}' \\
-  https://api.openverse.engineering/v1/auth_tokens/register
+  https://api.openverse.engineering/v1/auth_tokens/register/
 """  # noqa: E501
 
     register_api_oauth2_request = openapi.Schema(
@@ -238,7 +238,7 @@ You can use this endpoint to get information about your API key such as
     key_info_bash = """
 curl \\
   -H "Authorization: Bearer DLBYIcfnKfolaXKcmMC8RIDCavc2hW" \\
-  http://api.openverse.engineering/v1/rate_limit
+  http://api.openverse.engineering/v1/rate_limit/
 """
 
     @swagger_auto_schema(
--- a/api/catalog/urls/__init__.py
+++ b/api/catalog/urls/__init__.py
@@ -35,7 +35,7 @@ discontinuation_message = {
 }
 
 versioned_paths = [
-    path("rate_limit", CheckRates.as_view(), name="key_info"),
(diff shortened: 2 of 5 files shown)


sarayourfriend (approved):
Change looks fine to me but... I have to wonder.... are we being too fiddly with these trailing slashes? What is the reason to be so strict about them? It's caused issues with users trying to use the API as well, we've gotten at least one email that I'm aware of from someone confused by this.

dhruvkb:
I'm not a hundred percent sure of the reason but I've always preferred trailing slashes after URLs because Django will automatically append those if you miss it but will never strip them if you add one accidentally.

Also, the URL config looks much cleaner and nests easily if each path segment ends with a slash, so for consistency, it's generally easier to just add the slash everywhere.

dhruvkb:
I don't think I've seen any linter that can catch missing trailing slashes but I'll definitely go through the URLs manually and ensure we're not missing any.

sarayourfriend:
A unit test might be able to do it though, if not a linter (which I also doubt is able to given the programatic nature of these)?

sarayourfriend:
@obulat suggested that for validation we can use the `generateschema` command and then pass it through a validation. Olga, can you share what tool you were using to validate the generated schema?

sarayourfriend:
@dhruvkb I opened this PR with a proposal of a start to how we might validate it using Olga's approach: https://github.com/WordPress/openverse-api/pull/806

I'm not sure if it's the best way though and it doesn't appear to be currently catching the lack of trailing slashes on some endpoints, so it doesn't even seem to accomplish what we want from it anyway.

dhruvkb:
@sarayourfriend IIUC the specifications that the tool validates only check for a leading slash and not the trailing one.

sarayourfriend:
We could write our own validation against the generated YAML schema to check for the trailing slashed at least... in addition to spec based validation (which seems useful anyway?).

We'd need to extend the `validateopenapischema` command to check the parsed YAML for trailing slashes and raise an error when appropriate. Do you think that could work?

stacimc (approved):
This looks good to me. I'd be fine with merging this and continuing the unit testing work in a separate PR since it's an interesting problem all on its own; what do y'all think?

sarayourfriend:
+1, Staci. Makes sense to me!

PR #806, opened 2022-07-19
Add validate spec command

sarayourfriend:
## Description

Adds a command that we could run in CI to validate the schema. This is just a rough proposal. I see that it is not catching the lack of trailing slash on the `healthcheck` endpoint for example. Maybe there's some configuration we need to do with the `validate_spec`, like maybe pass a specific spec_url to use.

## Testing Instructions

Run `just build` so that you get the new dependencies. Then run `just dj validateopenapischema` and verify that it does not fail with validation errors.

If you want to see it working (i.e., failing) you can undo some of the changes I made like removing the trailing slashes from some URL configurations or adding back that `unique=True` on the auth_token endpoint configuration. The latter yields a very mysterious error that took me quite a while to find a solution for! This webapp gave me a clear, descriptive error that was much more helpful than the validate_spec error: https://editor.swagger.io/

Maybe there's a way to use the validation underlying that in the future as an iterative improvement on this.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound (approved):
I'm glad this is so easy! I don't have any advice for the non-OpenAPI-spec URLs unfortunately, but I would love to add this as part of CI too like you mention 🚀

krysal (commented):
Many packages are being updated here, do you think [these instructions](https://github.com/pypa/pipenv/issues/2665#issuecomment-524919810) could work for installing the new ones? 🤔

api/catalog/management/commands/validateopenapischema.py
```diff
@@ -0,0 +1,45 @@
...
+    def handle(self, *args, **options):
+        file_path = str(
+            (Path(options["output_dir"]).absolute() / "openapi.yaml").absolute()
+        )
+        call_command("generate_swagger", file_path, overwrite=True)
+        spec_dict, spec_url = read_from_filename(file_path)
+        # Do not catch the error here so that we can still write the file
+        # out for debugging.
+        self.handle_validation(spec_dict)
+        os.remove(file_path)
```
krysal, line 45: Using [`Path.unlink()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink) could save us the `os` import and convert the path to a string.

krysal (approved):
This is very helpful, thanks! I left two suggestions but this looks already g2g. Sorry for the late review.

api/catalog/api/views/oauth2_views.py
```diff
@@ -80,7 +80,6 @@ class Register(APIView):
                 type=openapi.TYPE_STRING,
                 min_length=1,
                 max_length=150,
-                unique=True,
                 description="A unique human-readable name for your application "
```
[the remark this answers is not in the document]
sarayourfriend: It does actually still have to be unique on the model side, it's just that the `unique` keyword argument on this function doesn't exist:

https://github.com/WordPress/openverse-api/blob/419082f155372ee05bdc5c4261feea06906eb49f/api/catalog/api/models/oauth.py#L11-L16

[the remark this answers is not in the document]
krysal: Ah, nice! I thought it was a bit weird it had to be unique among **all** apps (uniqueness with the email would make more sense), but I guess that makes our management job easier 🤷🏻‍♀️

q
```diff
@@ -0,0 +1,59 @@
+[1mdiff --cc .github/workflows/ci_cd.yml[m
+[1mindex 4c936852,1f70810a..00000000[m
+[1m--- a/.github/workflows/ci_cd.yml[m
```
krysal, line 3: What is this file? 😳

sarayourfriend: Oh gosh, I must have somehow accidentally piped `git diff` output to a file during the rebase :facepalm: It was a messy rebase :sweat_smile:

PR #822, opened 2022-07-27
Leave a comment and add `migrations` on issues that have migrations reminding the author to rebase before merging

### LINKED ISSUE #811: Prevent divergent migrations from being merged by concurrent, not-rebased PRs
## Problem

See https://github.com/WordPress/openverse-api/pull/810 and the description for the issue.

## Description

We need some way to prevent this.

Here are some suggestions I left in this comment, copied below: https://github.com/WordPress/openverse-api/pull/810#issuecomment-1191503799

> How can we prevent this? Is there any way we can ensure PRs with migrations are always rebased before merging? Could we run an action triggered by pushes that introduce new migrations using https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files and the iterate through all open, non-draft PRs and request changes on them if there is a migration in them?
> 
> We'd also need a similar job in the CI + CD workflow to check for migration conflicts against main any time the PR is pushed to.

Marked as high due to the risk it introduces.

## Implementation

sarayourfriend:
## Fixes

Related to #811 by @sarayourfriend 

## Description

This is just the first step on this issue. I'd like to keep it open for us to make more significant improvements in the future that will actually rebase the branch and run the check for you when possible and then request changes on the PR when the checks fail on the rebased version.

This just requires a whole lot more work on top of the script I already sort of had working.

## Testing Instructions

Check out the comments left on this PR.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend:
Okay. I'm going to let the CI finish running now. The target branch for this PR has been switched to a test branch and I've updated the CI to reference that branch name for the pr target events.

Next I need to introduce a migration in this branch. After that, I will push a migration to the target branch and we can see if the two Django checks run as expected.

sarayourfriend:
We probably might as well run unit tests in these cases as well, eh? :thinking: I can scope it to just the API files generally? Or should we just see how it goes with migrations and add unit testing to this stuff in the future?

sarayourfriend:
I wonder if there's an easy way to skip the build of the ingestion server for the pull request target event. Can we use an expression to define the matrix for that job?

sarayourfriend:
The target branch is updated...

sarayourfriend:
So what I've been trying to do is not possible yet: https://github.com/orgs/community/discussions/7399

The alternative approach we discussed was to have a workflow triggered by pushes to main that run checks against individual PRs with migrations. I believe this can work, it just requires us to use some "hacks" to make it possible. For example, we have no way to make the check required for a PR to be merged, so while the check is running, we would probably need to have the bot user request changes on the PR, adding a note that it is verifying migration safety. Then the bot user would dismiss its own requested changes if the checks pass. Otherwise, it would update the comment to note that the PR needs to be rebased due to a migration conflict.

@dhruvkb do you think that is workable? The script I had in mind would skip all PRs that don't introduce migrations... because it'd be cumbersome to handle multiple PRs in a single script, I think we'd want to dispatch a separate workflow per branch to run the check using the GitHub script action so that we could dispatch several at once.

What do you think? Is it worth trying?

dhruvkb:
Sorry, I accidentally hit the close button, that was not my intention. Still getting used to the trackpad.

dhruvkb:
I think for a first iteration, it might be sufficient to have the following:

1. a label that indicates that the PR must only be merged if up to date
2. a workflow that runs on `push` events on `main` and when `migrations/` is changed:  
   the workflow goes through all open PRs, finds PRs that contain migrations and applies the label mentioned in pt. 1 and adds a comment.
3. a workflow that runs on each PR `synchronize` if that contains a `migration/` change:  
   the workflow applies the label mentioned in pt. 1 and adds a comment.

The responsibility of keeping the PR updated falls on the maintainers but it's a good place to start and have something up quickly. Later, we can add more complex features to the workflows in pt. 2 and pt. 3.

sarayourfriend:
Okay, so basically just add the notification but don't actually do any checking?

sarayourfriend:
I had the script already kind of written, I'm sure it won't work right yet but if getting the checks to run isn't too big a lift, do you think it's worth doing now? Or should we really scope it down to just adding the warning? Maybe we can time box it and if I can't get the workflow dispatch check approach to work then I'll do the label and comment version you suggested and we can open another ticket later to iterate?

sarayourfriend:
Status update: I will get back to this PR next week once I am through with the mini support rotation. I will do as Dhruv suggested and just add a comment notification on the PR for now and open a follow up issue to do the more involved tested guarding.

.github/workflows/migration_safety_warning.yml
```diff
@@ -0,0 +1,41 @@
...
+              owner: 'WordPress',
+              repo: 'openverse-api',
+              comment_id: ${{ steps.existing-warning.outputs.comment-id }},
+            })
+
+      - uses: peter-evans/create-or-update-comment@v2
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            This PR has migrations. Please rebase it before merging to ensure that conflicting migrations are not introduced.
```
AetherUnbound, line 37: If I'm understanding this correctly, will this issue a comment on every PR that has a change to migrations, even if it's already rebased onto main?

AetherUnbound (approved):
Makes sense! It would be cool if we could have some logic to detect whether or not the branch is already rebased (which may be the case already if I'm misinterpreting things), but this is GTG 🚀 ✅

.github/workflows/migration_safety_warning.yml
sarayourfriend: Yes. I've tried to make much more complicated versions of this but I'd really like to just get this basic version in to mitigate the issue for now. There are much more compelling versions of this that would just run the check for you and only comment/request changes when it determines that the PR does indeed have migration conflicts (instead of just giving a gentle nudge).

The gentle nudge is good enough for now. Kudos to @dhruvkb for encouraging a good scope on this ticket.

stacimc (approved):
I haven't done much work with GH actions, but this looks good to me. Agreed that this seems like a very sensible first iteration. The added logic to remove an existing warning before nudging again is nice 👍

PR #830, opened 2022-07-31
Fix broken load script and port to Python

### LINKED ISSUE #828: Loop code is always running ingestion for images thrice
## Description

The code to rerun ingestion for images if it fails, added in #802 is broken. It's running the image ingestion three times in what seems to be a Bash code bug.

## Log

```
curl -X POST -H 'Content-Type: application/json' -d '{"model": "image", "action": "INGEST_UPSTREAM", "index_suffix": "init"}' 'http://localhost:8001/task'
{"message": "Successfully scheduled task", "task_id": "338b9145cc5b4049b6e7de6a4400a422", "status_check": "http://localhost:8001/task/338b9145cc5b4049b6e7de6a4400a422"}Waiting for index image-init to be ready...
./load_sample_data.sh: line 116: [0: command not found
curl -X POST -H 'Content-Type: application/json' -d '{"model": "image", "action": "INGEST_UPSTREAM", "index_suffix": "init"}' 'http://localhost:8001/task'
{"message": "Successfully scheduled task", "task_id": "043e83ff31d6480cadb106671a8175de", "status_check": "http://localhost:8001/task/043e83ff31d6480cadb106671a8175de"}./load_sample_data.sh: line 116: [0: command not found
curl -X POST -H 'Content-Type: application/json' -d '{"model": "image", "action": "INGEST_UPSTREAM", "index_suffix": "init"}' 'http://localhost:8001/task'
{"message": "Successfully scheduled task", "task_id": "e7874b5ca5d74e52bd6b9fb306b9dc88", "status_check": "http://localhost:8001/task/e7874b5ca5d74e52bd6b9fb306b9dc88"}./load_sample_data.sh: line 116: [0: command not found
```

## Additional context

It might be a good idea to replace the Bash script with Python as it's generally a much friendly language to write scripts in. It might help reduce the occurrence of bugs.

## Resolution

dhruvkb:
## Fixes

Fixes #828 by @dhruvkb 

## Description

This PR 
1. fixes the issue in the `load_sample_data.sh` script that caused the loop to run 3 times without ever breaking (commit fde670f)
2. ports the script to Python (making the above point redundant), thereby adding many new features and capabilities

The Python script has the following advantages:
- **idempotency**: This allows you to reset everything by running `just init` as many times as you want. There is more scope to improve but it will never crash on re-runs.
- **clarity**: The output of Bash scripts can be messy and out of control (e.g. `set -x`). With Python logging levels, you can tweak how much of a mess you want in the output.
- **DX friendliness**: Python is leaps and bounds ahead of Bash in terms of ease of use (subjective). The Python script is well documented, modular and easy to update.
- **testable**: It might be overkill but you can write unit tests for the script's functions. 

Oh, and you can drop into a shell and import and run the functions individually.

## Side-effects

To have a Python file in the root of the repo, we need to have a Python workspace at the root of the repo. In a separate PR we can delete `pre-commit` and `pycodestyle` from the inner packages.

In the future, we can consider moving the documentation tooling (Sphinx & friends) out of the API into the root as well.

## Testing Instructions

The passing CI should be sufficient. It means the load script is doing it's job. As for idempotency, follow these steps:
0. Check out the PR.
1. Repeatedly run `just init`.
2. Observe that there are no errors.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


dhruvkb:
Converting to draft to improve the output and logging of the script.

load_sample_data.py
```diff
@@ -0,0 +1,384 @@
...
+    with the same name can be deleted before copying and the table can be renamed after
+    copying.
+
+    :param name: the name of the source table to copy
+    :param target_name: the name to assign to the copied table
+    :param delete_if_exists: whether to delete any existing tables with the target name
+    """
+
+    if target_name is None:
+        target_name = name
```
obulat, line 113:
```suggestion
    target_name = target_name or name
```

obulat: Would this work, or is there a case where valid target_name can be falsy?

dhruvkb: That would definitely work. I just went with `is None` because that seemed to more clearly convey the intent.

justfile
```diff
@@ -73,8 +73,8 @@ env:
     cp ingestion_server/env.template ingestion_server/.env
 
 # Load sample data into the Docker Compose services
-init: up wait-for-es wait-for-ing wait-for-web
-    ./load_sample_data.sh
+init log_level="INFO": up wait-for-es wait-for-ing wait-for-web
+    env LOG_LEVEL={{ log_level }} pipenv run python load_sample_data.py
```
obulat, line 77: There needs to be an install step before `load_sample_data`, because currently it breaks because `decouple` is not installed:
```
    from decouple import config
ModuleNotFoundError: No module named 'decouple'
```

sarayourfriend (commented):
Are there any drawbacks to moving the load_sample_data.py into a new directory, maybe `utils` or `bin` or something like that, and keeping the Pipfile* in there? I'm worried about the potential confusion of having nested Pipenv files. It also might be nice to avoid setting a precedence of putting additional files at the root level of the repository when it's avoidable. Additionally it would more nicely allow us to break the script up into a couple different modules so that it's easier to comprehend everything that's going on in there without scrolling through a 460+ line file.

sarayourfriend (commented):
I actually kind of wonder if some of this would be better off (more manageable, intelligible, smaller, simpler in some cases) as a management command instead of inlined python scripts passed to docker exec. We'd just want to make sure it can only be run locally, I guess...

Why was the sample images CSV renamed, btw?

sarayourfriend:
This works for me fine, locally, btw, so that's awesome! The improved logging is also super welcome, thanks for that!

dhruvkb:
@sarayourfriend I agree with you that it initially seemed confusing to have a `Pipfile` in the root but I found no problems when using it. It felt very much akin to working in a monorepo. Having the root Python environment allows us to use it for things that are common to the codebase such as `pre-commit` (and in the future, also Sphinx). Breaking the script into files and making it a module does seem useful though.

Since most of the actions are happening in services other than the API, creating management commands does not seem a very appealing idea to me. Also this script is only for setting up the dev environment and will never really be used in production, so I like that this code is not mixed with any of the actual services'. Happy to explore any alternatives compared to passing inline-scripts to Docker.

The file was renamed because the media types are `'image'` and `'audio'`. The new name makes it easier to locate the file using f-strings like `f'./sample_data/sample_{media_type}.csv'`.

obulat (approved):
One initial thought I had was that the script has become too nested: just script is calling a python script which is calling a just script that will run the python function.

The most clear example of it is the `create_users` function :) I think it could be extracted to a separate Python file where we would call it using `django.core.management.call_command` ([docs](https://docs.djangoproject.com/en/4.0/ref/django-admin/#running-management-commands-from-your-code)).
But since most of the python code in `load_sample_data.py` is used to create the SQL command, I'm not sure how much improvement that would bring.

I guess we can and will improve this script in the future, but it's already a big improvement and unblocks work, so ✅

dhruvkb:
@obulat, you're right, there is a lot of scope to simplify and un-nest these context but hopefully we can do that in subsequent improvement PRs instead of dragging this one out for longer.

dhruvkb:
@sarayourfriend @krysal could either of you quickly review this PR? It's already proven that it works fine and fixes a bug. We can iteratively improve it in the future.

sarayourfriend:
I will review this on Monday if Krystle doesn't get it it before me.

load_sample_data.py
```diff
@@ -0,0 +1,469 @@
...
+                user = User.objects.create_user(
+                    username, f'{{username}}@example.com', 'deploy'
+                )
+                user.save()
+        EOF"""
+    compose_exec(WEB_SERVICE_NAME, bash_input)
+    logging.info("Users created.")
+
+
+def backup_table(media_type: MediaType):
```
krysal, line 225: This step is new to me, do we need to backup test data? 🤔

krysal (commented):
It is impressive that you were able to port the `load_sample_data.sh` script so quickly to Python @dhruvkb, but I'm not sure how to feel about this change. The current script is 129 lines and is easier to grok than its 469 python lines counterpart here, or maybe I'm used to it. As Olga mentioned, its mostly doing call to `just` commands and using bash anyway so it's more tangled.

I see good info here in the comments, so I probably may have to look at it again on Monday but would prefer if only the problematic section (ingestion) is ported to python and fixed here.

load_sample_data.py
dhruvkb: This doesn't back up the test data, it only keeps a copy of the `image` and `audio` tables so that we can run the ingestion script again idempotently.

During ingestion, the ingestion server modifies the tables in the downstream DB, which means that if we run the `load_sample_data` script again, it will fail. This step allows us to use the initial tables (as created by Django) as the starting point every time we run this script.

dhruvkb:
The Bash script is definitely smaller but would mostly be because of 
- no functions (everything runs in a single block, as Bash scripts generally are)
- no docstrings (except for a few lines describing what's happening)
- direct calls to `just` `docker-compose` (which Python cannot do, unfortunately)

I find Bash's syntax confusing and easier to break accidentally (subjective, from personal experience) so I prefer having the script be pure Python but if the majority opinion is that it feels like a detrimental change, we can revert to fde670f and just merge the fix.

load_sample_data.py
dhruvkb: 1. Django migration creates `image` in downstream.
2. We copy the `image` table to `image-template`.
3. `image-template` is used to build the `image-view` table in upstream.
4. Data is loaded into `image-view` from the CSV.
5. Data refresh overwrites downstream `image` with upstream `image-view`.

Since `image` has changed in step 5 we cannot use it to create `image-view` again. By having it backed up in step 2 we ensure that step 3-5 can be repeated again and again, allowing us to run `just init` on an existing init-ed setup.

sarayourfriend:
It's worth noting that Bash has functions which can be documented. Redirecting printf allows for error and debug logging. It has different ergonomics but for executing shell commands (just and docker) it is a more flexible and direct tool.

I welcome all of these improvements, but I am also personally in favor of smaller PRs that do not push the scope of the issue into refactors that are not strictly necessary for accomplishing the issue. This allows us to better estimate the amount of time an issue will take (because we are sticking to it's definition instead of interpreting them expansively) and makes review easier because we are only reviewing the scope of the issue, instead of in effect the issue itself as well as a refactor. It's also worth noting that "refactoring to fix" obscures the fix when the PR is squashed and merged to main which can make understanding the history of an area and why it changed that much more difficult (in fact, requires the GitHub UI to do so).

I don't think the refactor is itself bad. However, I would want it to be iterated upon and improved (factored into smaller pieces than it is now), which is another dedication of time that I assume we'd probably rather spend on other things than this utility script.

We can also see the other issue with refactors like this one: merge conflicts. We've run into this several times with big refactors that were not strictly necessary.

I do wonder sometimes where these refactors come from. Are they indications that the code base is really that difficult to work in that we feel so motivated to spend significant amounts of time refactoring and supporting those refactors (longer, more in-depth and careful reviews, riskier deployments, etc) vs fixing any of the many bugs that exist in the application? It's  a general question the project should come to an understanding of. The broader WordPress project has made its position on refactors clear: https://make.wordpress.org/core/2011/03/23/code-refactoring/. I do strongly agree with the sentiments there. Even performance may be the very thing that is biting us in the butt with the largest recent refactor around the search controller area. But we don't know, because the refactor is not measured in that way.

Anyway, I do not feel strongly about this PR. I don't feel the refactor is necessary or brings significant enough value to have justified the time spent on it, but it was neither my time nor my call to make on that front. People are welcome to spend their time how they see fit. It is however everyone's responsibility to understand and take into consideration the effect of a refactor on the reviewers and folks who will go on to maintain the new code. Especially a refactor that adds more code than existed before or completely changes the approach to a problem. We all know the adage: the best code to maintain is code that doesn't exist. Of course that's not a hard and fast rule, but it is very true in my experience.

Personally I do find it frustrating to open a PR to review that is meant to fix something to discover that it is adding a large number of new lines of code and completely changing how something is done when the fix does not _necessarily_ call for it. I will of course still review those PRs when they are assigned to me. But I will also continue to advocate for aggressively cutting scope in PRs and preventing scope creep outside the original issue.

And just to be clear @dhruvkb this isn't directed at you specifically. This is something I see us doing as a project in general (and something I have done as well) and is also everyone working on the project's responsibility. It's something we all should keep in mind and I think that we can improve on. It is also something that we should hold each other accountable for, if indeed it is something we agree on. We must also not believe in the sunk cost of already written code that we are deleting without it ever getting committed to main. At worst, we all learn something when trying these things and at best we avoid making unnecessary changes that have unknown downstream effects even after it has been reviewed and merged.

cc @WordPress/openverse because this discussion is about the project in general.

Also: this isn't a big statement about this particular PR. I am just explaining my position. I reiterate that I do not feel strongly one way or another about this PR beyond what I have shared above, mostly because it is a non-critical utility script that can break without too much issue. I will not block this PR from merging in its current state (after rebasing), even if I will remain sceptical of the broader changes contained within it.

zackkrida:
Could we keep this PR, get the change from https://github.com/WordPress/openverse-api/commit/fde670fd2c575bc126ac42687a7fd54c881c1e6f into a standalone PR, and continue to discuss the merits of switching to Python? I don't want to block that small, single-character-diff bugfix.

dhruvkb:
Opened #869 to fix the bug. I'll close this as it's way out of scope and unimportant right now. Maybe we can circle back to it someday but this is not the time for such changes.

PR #844, opened 2022-08-03
Remap ports to the non-conflicting range

dhruvkb:
## Fixes

Addresses WordPress/openverse#200

## Description

This PR
- removes binding for ports that are only used for internal communication
- changes port mappings to ports in the 502xx range to prevent any conflicts with services running on the host
- updates documentation and tests to keep up with the changes
- fixes some small side issues along the way
  - removing unused settings
  - using env vars in the `run.sh` entrypoint code and aligning it with the `justfile`
  - using the `_api-up` recipe in more places
  - converting the NGINX conf into [a template to use env vars](https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration-new-in-119)

## Testing Instructions

0. Check out the PR and run the dev server.
1. Visit some endpoints that can help gauge the communication between processes e.g.
   - the stats endpoint, to test ES and cache
   - the thumbnails endpoint, to test the thumbnails service

Apart from this, a passing CI is pretty good proof of inter-process communication working normally.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -143,8 +143,7 @@ jobs:
         run: just init
 
       - name: Smoke test ReDoc site
-        run: |
-          curl --fail 'http://localhost:8000/v1/?format=openapi'
+        run: just api-doctest
 
       - name: Run API tests
         run: just api-test
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -142,9 +142,6 @@ jobs:
       - name: Start API, ingest and index test data
         run: just init
 
-      - name: Smoke test ReDoc site
-        run: just api-doctest
-
       - name: Run API tests
         run: just api-test
 
@@ -160,6 +157,31 @@ jobs:
         with:
           name: api_logs
           path: api_logs
+
+  test-redoc:
+    name: Check for API consumer docs
+    runs-on: ubuntu-latest
+    needs:
+      - build-images
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup just
+        uses: extractions/setup-just@v1
+
+      - name: Download all images
+        uses: actions/download-artifact@v2
+        with:
+          path: /tmp
+
+      - name: Load all images
+        run: |
+          docker load --input /tmp/api/api.tar
+          docker load --input /tmp/ingestion_server/ingestion_server.tar
+
+      - name: Test ReDoc site
+        run: just api-doctest
 
   django-check:
     name: Run Django check
--- a/api/docs/guides/test.md
+++ b/api/docs/guides/test.md
@@ -12,12 +12,3 @@
    ```bash
    just api-test
    ```
-
-## On the host
-
-1. Start the Django dev server. See the {doc}`"Run" <./run>` document for instructions.
-
-2. Run the tests in a terminal on the host.
-   ```bash
-   just api-testlocal
-   ```
--- a/api/run.sh
+++ b/api/run.sh
@@ -2,10 +2,9 @@
 
 set -e
 
-while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' http://es:9200/)" != "200" ]]
-do
-  echo "Waiting for Elasticsearch connection..."
-  sleep 2
+while [ "$(curl -s -o /dev/null -w '%{http_code}' "http://${ELASTICSEARCH_URL:-es}:${ELASTICSEARCH_PORT:-9200}/_cluster/health")" != "200" ]; do
+  echo "Waiting for Elasticsearch connection..." && sleep 5;
 done
+echo "Elasticsearch connection established!"
 
 exec "$@"
--- a/api/run.sh
+++ b/api/run.sh
@@ -2,7 +2,8 @@
 
 set -e
 
-while [ "$(curl -s -o /dev/null -w '%{http_code}' "http://${ELASTICSEARCH_URL:-es}:${ELASTICSEARCH_PORT:-9200}/_cluster/health")" != "200" ]; do
+es_url="http://${ELASTICSEARCH_URL:-es}:${ELASTICSEARCH_PORT:-9200}/_cluster/health"
+while [ "$(curl -s -o /dev/null -w '%{http_code}' "$es_url")" != "200" ]; do
   echo "Waiting for Elasticsearch connection..." && sleep 5;
 done
 echo "Elasticsearch connection established!"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,8 +72,8 @@ services:
     volumes:
       - ./api:/api
     ports:
-      - "50280:8000"
-      - "50230:3000"
+      - "50280:8000" # Django
+      - "50230:3000" # Sphinx (unused by default; see `sphinx-live` recipe)
     depends_on:
       - db
       - es
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,8 +2,8 @@ version: "2.4"
 services:
   db:
     image: postgres:13.2-alpine
-    expose:
-      - "5432"
+    ports:
+      - "50254:5432"
     env_file:
       - postgres/env.docker
     healthcheck:
@@ -34,8 +34,8 @@ services:
 
   cache:
     image: redis:4.0.10
-    expose:
-      - "6379"
+    ports:
+      - "50263:6379"
 
   es:
     image: docker.elastic.co/elasticsearch/elasticsearch:7.12.0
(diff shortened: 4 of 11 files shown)


api/run.sh
```diff
@@ -2,10 +2,9 @@
 
 set -e
 
-while [[ "$(curl --insecure -s -o /dev/null -w '%{http_code}' http://es:9200/)" != "200" ]]
-do
-  echo "Waiting for Elasticsearch connection..."
-  sleep 2
+while [ "$(curl -s -o /dev/null -w '%{http_code}' "http://${ELASTICSEARCH_URL:-es}:${ELASTICSEARCH_PORT:-9200}/_cluster/health")" != "200" ]; do
```
sarayourfriend, line 5: Can this line be split up?

```suggestion
es_url = "http://${ELASTICSEARCH_URL:-es}:${ELASTICSEARCH_PORT:-9200}/_cluster/health"
while [ "$(curl -s -o /dev/null -w '%{http_code}' "$es_url")" != "200" ]; do
```

Would this work?

justfile
```diff
@@ -222,13 +226,13 @@ ipython:
...
 # Compile Sphinx documentation into HTML output
-sphinx-make: up wait-for-es wait-for-ing wait-for-web
+sphinx-make: _api-up
     just exec web sphinx-build -M html docs/ build/
 
 # Serve Sphinx documentation via a live-reload server
-sphinx-live port="3000": up wait-for-es wait-for-ing wait-for-web
-    just exec web sphinx-autobuild --host 0.0.0.0 --port {{ port }} docs/ build/html/
+sphinx-live: _api-up
+    just exec web sphinx-autobuild --host 0.0.0.0 --port 3000 docs/ build/html/
```
sarayourfriend, line 234: Would we want to change this port to something in 502xx as well?

dhruvkb: The 3000 here refers to the port of the container (running inside `just exec`). It is correctly set to 3000.

api/run.sh
dhruvkb: It works, except for the spaces around `=` (shell scripts, right?). Fixed in 6df39aa.

.github/workflows/ci_cd.yml
```diff
@@ -162,6 +158,31 @@ jobs:
...
+
+      - name: Download all images
+        uses: actions/download-artifact@v2
+        with:
+          path: /tmp
+
+      - name: Load all images
+        run: |
+          docker load --input /tmp/api/api.tar
+          docker load --input /tmp/ingestion_server/ingestion_server.tar
```
sarayourfriend, line 181: It doesn't look like the ingestion server is needed for the api doctest recipe.
```suggestion

```

api/docs/guides/test.md
```diff
@@ -12,12 +12,3 @@
...
    ```
-
-## On the host
-
-1. Start the Django dev server. See the {doc}`"Run" <./run>` document for instructions.
-
-2. Run the tests in a terminal on the host.
-   ```bash
-   just api-testlocal
-   ```
```
sarayourfriend, line 23: Why is this removed?

docker-compose.yml
```diff
@@ -2,8 +2,8 @@ version: "2.4"
 services:
   db:
     image: postgres:13.2-alpine
-    ports:
-      - "5432:5432"
+    expose:
+      - "5432"
```
sarayourfriend, line 6: I think some folks on occasion do connect to the local DB using psql. It'd be nice to leave this one mapped to the host as well.

docker-compose.yml
```diff
@@ -34,13 +34,13 @@ services:
 
   cache:
     image: redis:4.0.10
-    ports:
-      - "6379:6379"
+    expose:
+      - "6379"
```
sarayourfriend, line 38: Ditto with Redis. It can be useful to be able to debug with the Redis CLI locally.

justfile
```diff
@@ -73,8 +73,12 @@ env:
     cp api/env.template api/.env
     cp ingestion_server/env.template ingestion_server/.env
 
+# Ensure all services are up and running
+@_api-up: up wait-for-ing wait-for-web
+    exit 0
```
sarayourfriend, line 78: This seems kind of like a weird way to write this recipe. Wouldn't this be a little more expected/natural?

```suggestion
# Ensure all services are up and running
@_api-up:
    just up
    just wait-for-ing
    just wait-for-web
```

docker-compose.yml
```diff
@@ -72,8 +72,8 @@ services:
     volumes:
       - ./api:/api
     ports:
-      - "8000:8000"
-      - "3000:3000"
+      - "50280:8000"
+      - "50230:3000"
```
sarayourfriend, line 76: Just a small suggestion because this always confuses me when I check the port mappings in the docker compose. I never remember which to use for the Django API.

```suggestion
      # django
      - "50280:8000"
      # sphinx
      - "50230:3000"
```

sarayourfriend (changes requested):
LGTM! Tested fine locally. Only one of the suggestions I'm strong-ish on is mapping the Postgres and Redis ports. Otherwise this is great.

api/docs/guides/test.md
dhruvkb: Since I hadn't mapped the DBs and Redis to the host, testing locally was not possible. Now that you and @AetherUnbound have expressed support for mapping these services locally, we can restore this functionality.

justfile
dhruvkb: I don't know who wrote the recipe like that but it's been fixed in 670dd8a.

docker-compose.yml
dhruvkb: Added comments in faa97ae.

docker-compose.yml
dhruvkb: Exposed port in d714b04.

docker-compose.yml
dhruvkb: I have never done this but that shouldn't prevent anyone else from being able to do so. Exposed port in d714b04.

api/docs/guides/test.md
sarayourfriend: I typically use `api-testlocal` when running unit tests locally.

dhruvkb: Can we fix that in a separate PR? Our tests are fundamentally flawed in that there is no separation between unit tests which should run in the Django environment and E2E/integration tests which should run outside. This means that when ports are different  inside and outside the container they cannot use the same env vars.

Updating tests would make this PR quite big and hard to review. So I'm hoping we can open a high-priority issue for that and fix it in post.

As a workaround, you can speed up test execution by running specific tests by passing a filename after `just api-test <file_name>`.

sarayourfriend: I always run specific tests when writing new ones, the issue is the container startup time and difficulties with attaching debuggers.

I'm fine solving it in a follow up PR but please open an issue describing the problem and the solution. I'm not entirely convinced the integration tests need to run outside the container (see Playwright tests in the frontend, for example) but we can discuss the trade offs in another issue as you suggested.

sarayourfriend (approved):
LGTM!

dhruvkb:
@obulat could you please review the PR when you have some time? Thanks!

krysal:
@dhruvkb After Olga's review, be sure to merge this PR after the v2.5.10 is cut, please. This is looking like a non-trivial change.

sarayourfriend:
Good point @krysal. Should just hold off on merging this at all until the ES situation is resolved?

dhruvkb:
We did agree to no merge anything unrelated to ES but my interpretation was to not touch the ES-related bits. I can definitely hold off on merging this but this only affects the dev setup, there is no change to the code itself.

I'll wait for the v2.5.0 to be cut before merging and block it till then.

dhruvkb:
Merging since 2.5.10 has been cut and deployed to production.

.github/workflows/ci_cd.yml
AetherUnbound, line 181: I don't know if this is possible with GHA, but it would be really cool to have this as a set of steps that we could just reference as a precursor to specific tasks! Since this does appear to be reused elsewhere in this file.

api/docs/guides/test.md
AetherUnbound: FWIW the catalog has a just recipe for setting up a "test session" `docker run`, which I use frequently for rapidly iterating on tests: 
https://github.com/WordPress/openverse-catalog/blob/40b4306b161bb4328f4b2e77bc9790e34332eb41/justfile#L65-L67

Perhaps it could be used here!

AetherUnbound (approved):
Looks and works great!

.github/workflows/ci_cd.yml
sarayourfriend: It is possible using YAML anchors but `actionlint` does not support them yet: https://github.com/rhysd/actionlint/issues/133

PR #859, opened 2022-08-05
Limit page depth of searches

### LINKED ISSUE #857: Limit page depth of searches
## Problem

As a temporary fix for some performance issues with the Openverse API, we should cap the max page depth to `20`. This means that if a search had 40 pages of results users would still be limited to accessing pages 1-20.  

## Description

This prevents the scenario described here https://github.com/cc-archive/cccatalog-api/issues/565, which has become an increasing problem with the number of users scraping Openverse's dataset, while we make improvements to pagination and dead link elimination.

## Implementation

AetherUnbound:
## Fixes

Fixes #857 by @zackkrida

## Description

This PR adds a cap to the number of pages returned in API results to 20 (defined by `MAX_TOTAL_PAGE_COUNT` in `pagination.py`). It does so by:

1. Preventing the pagination class from setting a page size that's higher (e.g. when `page` is in the query parameters)
2. Limiting the number returned when determining "result and page count"

The former is pretty straightforward and I've added tests for that case. The latter makes way less sense to me. I've capped the `page_count` within that function to `MAX_TOTAL_PAGE_COUNT` as well, but the logic within it is confusing. I wrote tests along with it with some of my own notes. I hesitate to change _more_ functionality because the current calculations are "working", I only altered what the possible cap was.

## Testing Instructions

- `just api-test`
- Try http://localhost:8000/v1/images/ before and after checking this branch out and see that even though there are at least 250 pages worth of results, only 20 pages are shown (and going to page 21 and above returns an error)

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Pipfile
+++ b/api/Pipfile
@@ -7,6 +7,7 @@ verify_ssl = true
 ipython = "*"
 pycodestyle = "*"
 pytest-django = ">=3.5"
+pytest-raises = "*"
 remote-pdb = "*"
 pre-commit = "*"
 locust = "*"
--- a/api/catalog/api/utils/pagination.py
+++ b/api/catalog/api/utils/pagination.py
@@ -4,6 +4,9 @@ from rest_framework.response import Response
 from catalog.api.utils.exceptions import get_api_exception
 
 
+MAX_TOTAL_PAGE_COUNT = 20
+
+
 class StandardPagination(PageNumberPagination):
     page_size_query_param = "page_size"
     page_query_param = "page"
@@ -42,6 +45,8 @@ class StandardPagination(PageNumberPagination):
         value = int(value)  # convert str params to int
         if value <= 0:
             raise get_api_exception("Page must be greater than 0.", 400)
+        elif value > 20:
+            raise get_api_exception("Searches are limited to 20 pages.", 400)
         self._page = value
 
     def get_paginated_response(self, data):
--- a/api/test/unit/controllers/test_search_controller.py
+++ b/api/test/unit/controllers/test_search_controller.py
@@ -1,0 +1,51 @@
+from unittest import mock
+
+import pytest
+
+from catalog.api.controllers import search_controller
+from catalog.api.utils.pagination import MAX_TOTAL_PAGE_COUNT
+
+
+@pytest.mark.parametrize(
+    "total_hits, real_result_count, page_size, expected",
+    [
+        # No results
+        (0, 0, 10, (0, 0)),
+        # Setting page size to 0 raises an exception
+        pytest.param(
+            0, 0, 0, (0, 0), marks=pytest.mark.raises(exception=ZeroDivisionError)
+        ),
+        # Fewer results than page size leads to max of result total
+        (5, 5, 10, (5, 0)),
+        # Even if no real results exist, total result count and page count are returned
+        # (seems like an impossible case IRL)
+        (100, 0, 10, (100, 10)),
+        # If there are real results and ES reports no hits, nothing is expected
+        # (seems like an impossible case IRL)
+        (0, 100, 10, (0, 0)),
+        # Evenly divisible number of pages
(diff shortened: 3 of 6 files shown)


api/catalog/api/utils/pagination.py
```diff
@@ -42,6 +45,8 @@ def page(self, value):
         value = int(value)  # convert str params to int
         if value <= 0:
             raise get_api_exception("Page must be greater than 0.", 400)
+        elif value > 20:
+            raise get_api_exception("Searches are limited to 20 pages.", 400)
```
sarayourfriend, line 49: This validation should be removed from here and the new validation should be added to the serializer: https://github.com/WordPress/openverse-api/blob/d26df92f04120e7382d042ae0b439fb6ea9209be/api/catalog/api/serializers/media_serializers.py#L167-L174

```suggestion
```

It was erroneously added back to this area during the revert.

api/Pipfile
```diff
@@ -7,6 +7,7 @@ verify_ssl = true
 ipython = "*"
 pycodestyle = "*"
 pytest-django = ">=3.5"
+pytest-raises = "*"
```
sarayourfriend, line 10: Is this new dependency being used at all?

sarayourfriend (changes requested):
LGTM but we should keep the validation in the serializer and remove the code added back in the revert.

api/test/unit/controllers/test_search_controller.py
```diff
@@ -0,0 +1,51 @@
...
+from catalog.api.utils.pagination import MAX_TOTAL_PAGE_COUNT
+
+
+@pytest.mark.parametrize(
+    "total_hits, real_result_count, page_size, expected",
+    [
+        # No results
+        (0, 0, 10, (0, 0)),
+        # Setting page size to 0 raises an exception
+        pytest.param(
```
obulat, line 15: Does it make sense to add a negative page size, too?

sarayourfriend:
Assigning this to myself as Madison is the MSR runner for this week and I will be taking this off her hands.

api/catalog/api/utils/pagination.py
sarayourfriend: Actually, this is inaccurate. The validation needs to be moved to the serializer, but I think it should be done in a different PR dedicated to a small refactor of the pagination parameter validation. We're under-utilising serializer validation at the moment and we can delete a significant amount code. But it should be done in a different PR.

sarayourfriend (approved):
LGTM.

api/test/unit/controllers/test_search_controller.py
sarayourfriend: Upstream validation catches negative values here.

api/catalog/api/utils/pagination.py
zackkrida, line 49:
```suggestion
            raise get_api_exception("Searches are limited to the first 20 pages of results while we mitigate API performance issues. We are working on a solution for bulk exports of Openverse's data.", 400)
```

Might want to be super verbose here? Not sure what else we could say 🤷 CC @WordPress/openverse-maintainers if anyone else has ideas. Would we, for example, want to tell them to email us or anything? Not sure.

AetherUnbound: Maybe we could just say `"Searches are currently limited to 20 pages"`, which implies that might change in the future?

api/Pipfile
AetherUnbound: It is used here: https://github.com/WordPress/openverse-api/pull/859/files#diff-3f72d9dc992edb46dcea0b7136c9ecaa0f888653a2c4c6261ba0116de468eef8R16

PR #862, opened 2022-08-08
Define user to be used to run container process

### LINKED ISSUE #820: Fix docker generated file permissions
## Description

When docker runs as root and creates files on a mounted volume, those files will have `root:root` permissions and will not be able to be modified or deleted by a regular user on the computer. See https://github.com/WordPress/openverse-api/runs/7526184802?check_suite_focus=true for example.

We can fix the permissions by running as a non-root user. See how this is done for the frontend: https://github.com/WordPress/openverse-frontend/blob/c4a117a11ddda2d0085506c88284db43e2784223/docker-compose.playwright.yml#L11

## Reproduction

Run the API via docker and then try to delete the files it creates as a normal user on your computer. It should fail similarly to in the GitHub action linked above.

## Resolution

ritesh-pandey:
## Fixes

Fixes #820  by @sarayourfriend 

## Description

Configures user to be used to run container commands.

This will fix file permission issue. By default files created by Docker are owned by `root`. This creates problem when a dev tries to manipulate those files.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -82,6 +82,7 @@ services:
       - api/env.docker
     stdin_open: true
     tty: true
+    user: $(USER_ID)
 
   ingestion_server:
     build: ./ingestion_server/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -82,7 +82,7 @@ services:
       - api/env.docker
     stdin_open: true
     tty: true
-    user: $(USER_ID)
+    user: ${DOCKER_USER_ID}:${DOCKER_GROUP_ID}
 
   ingestion_server:
     build: ./ingestion_server/
--- a/justfile
+++ b/justfile
@@ -32,6 +32,9 @@ DOCKER_FILE := "-f " + (
     else { "docker-compose.yml" }
 )
 
+export DOCKER_USER_ID := `id -u`
+export DOCKER_GROUP_ID := `id -g`
+
 # Build all (or specified) services
 build *args:
     docker-compose {{ DOCKER_FILE }} build {{ args }}


sarayourfriend:
Thanks for the PR @ritesh-pandey. Docker compose uses curly braces for interpolation: `${USER_ID}`. However, `USER_ID` isn't defined in CI.

Where available, we use the [`id`](https://www.man7.org/linux/man-pages/man1/id.1.html) command to get the user ID and the group ID. You can see us doing so in the frontend repository here:

https://github.com/WordPress/openverse-frontend/blob/c4a117a11ddda2d0085506c88284db43e2784223/bin/playwright.sh#L7

It will be necessary to define the `USER_ID` variable in the environment before running the docker commands in the `justfile`.

ritesh-pandey:
I saw the failed build and realised that something is wrong. 
Let me mark this PR as draft.

ritesh-pandey:
Thank you @sarayourfriend.

I've added two new variables - `DOCKER_USER_ID` and `DOCKER_GROUP_ID` in `justfile`. They are assigned values using `id`. These variables are then exported as environment variables so that `docker-compose` file can access them.

I've tested this by running `api-test` and looking at the file permissions of cache files. Earlier they were owned by `root`. Now the current user owns them.

sarayourfriend (approved):
Awesome! Thank you so much, Ritesh. :rocket:

krysal:
@sarayourfriend how were you able to confirm this fixes the issue? I tried generating the docs and inspected files permissions in the container and I see them still owned by `root`, with two new warnings when I connect.

```sh
# Recreate
docker-compose run web bash

# Build the Sphinx docs
just sphinx-make

# then I connect to the container and list the generated doc files
docker-compose run web bash

WARNING: The DOCKER_USER_ID variable is not set. Defaulting to a blank string.
WARNING: The DOCKER_GROUP_ID variable is not set. Defaulting to a blank string.
Creating openverse-api_web_run ... done
root@056b6c57b3f9:/api# ls -la build/html/
total 356
drwxr-xr-x 15 root root    480 Aug  9 16:51 .
drwxr-xr-x  4 root root    128 Aug  9 16:51 ..
-rw-r--r--  1 root root    230 Aug  9 16:51 .buildinfo
drwxr-xr-x  4 root root    128 Aug  9 16:51 _images
drwxr-xr-x  6 root root    192 Aug  9 16:51 _sources
drwxr-xr-x 23 root root    736 Aug  9 16:51 _static
-rw-r--r--  1 root root 175363 Aug  9 16:51 genindex.html
drwxr-xr-x 11 root root    352 Aug  9 16:51 guides
-rw-r--r--  1 root root  19125 Aug  9 16:51 index.html
-rw-r--r--  1 root root   5815 Aug  9 16:51 objects.inv
-rw-r--r--  1 root root  30973 Aug  9 16:51 py-modindex.html
drwxr-xr-x  5 root root    160 Aug  9 16:51 reference
-rw-r--r--  1 root root  18013 Aug  9 16:51 search.html
-rw-r--r--  1 root root  72992 Aug  9 16:51 searchindex.js
-rw-r--r--  1 root root  24651 Aug  9 16:51 terms_of_service.html
```

sarayourfriend:
@krysal if you run `docker-compose` directly without using `just` the variables will not exist. They're only defined in the `justfile`. In that case it will fall back to the default behaviour that matches what we have now (owned by root) with the warning. If you do `just up` etc instead, does it work?

We could suppress the warning by adding defaults in the docker-compose file though.

krysal:
I did run the sphinx command with `just`, so the docs files should be owned by my non-root user. Though it's a good point that we would need a just command to enter the container, something like `just bash <service>` to use the current user, but that it's not the main issue here. I wonder what could be the reason 🤔 I can't be sure it will fix the CI pipeline issue but perhaps it works for @dhruvkb.

sarayourfriend:
It's interesting, it doesn't matter what I do it always works for me as long as I use just. If I run `just sphinx-make`, as long as I've deleted the root-created `api/build` directory then it completes successfully and the files are all owned by my user.

AetherUnbound:
This isn't super helpful now, but I _do_ remember there being some difference between how users behave with Docker on Mac vs. on Linux. I suspect that may be the discrepancy we're seeing, I know that I tried to address this in a previous project. I'll check in on that today and see if there's anything helpful I can share from that.

krysal:
Interestingly it works the same: the files in the container are owned by `root` but in the host (macOS) are owned by my user. Probably it's just an issue of Docker in non-macOS hosts then.

sarayourfriend:
As long as they work on the host and the processes run inside the container as expected then I think we're good here. The issue is to get them easily accessible by the host without breaking the container runs.

docker-compose.yml
```diff
@@ -82,6 +82,7 @@ services:
       - api/env.docker
     stdin_open: true
     tty: true
+    user: ${DOCKER_USER_ID}:${DOCKER_GROUP_ID}
```
dhruvkb, line 85: I recommend setting defaults for this to avoid these warnings. It seems inevitable  to me that we will occasionally invoke `docker compose` outside of a `just` recipe. 

Screenshot 2022-08-17 at 5 52 40 PM

dhruvkb (approved):
I confirmed that on macOS, as long as you start the containers (with `just up`) and then build the docs (with `just sphinx-make`), the generated files are owned by the host user. This should fix our problems 🎉! 

Side-note: Is this line no longer necessary in the workflow?
https://github.com/WordPress/openverse-api/blob/db6f6ef24961cb0f821058d94e70b3dcc37ed70e/.github/workflows/ci_cd.yml#L367

docker-compose.yml
sarayourfriend: > It seems inevitable to me that we will occasionally invoke docker compose outside of a just recipe

Under what circumstances would one do this? If there aren't concrete use cases then I don't think we should worry about supporting them. In fact, supporting them could get us into trouble as we would have to continue to monitor those workflows.

Given how trivial it is to add new recipes to the justfile I think we should just expect people to always go through `just`... we have enough specific configuration stuff as it is in there that it's non-trivial to run a `docker compose` command that will actually work outside very specific circumstances.

dhruvkb: I regularly run `docker-compose ps` to see the status of services, `docker-compose restart` to restart some services etc. Going through just for every use case of `docker-compose` is not feasible.

sarayourfriend: We can fairly easily create a `dc` recipe for this:

```make
dc *args:
    docker-compose {{ DOCKER_FILE }} *args
```

We don't have to re-write every command. It would simplify our justfile as well. All the `docker-compose {{ DOCKER_FILE }}` stuff can be replaced with `just dc`.

sarayourfriend: PR for the above suggestion: https://github.com/WordPress/openverse-api/pull/886

PR #864, opened 2022-08-08
Add throttling classes for thumbnail endpoint

### LINKED ISSUE #863: Adjust throttling of thumbnail endpoints
## Problem

Currently thumbnail requests are throttled at a rate of one thousand per minute:

https://github.com/WordPress/openverse-api/blob/e9cfd3cc705c573e3f6ab2b632737e94bb82aee0/api/catalog/api/views/image_views.py#L96

When used by unauthenticated users making requests rapidly, as is common in scraping workflows, this can potentially send a lot of traffic to our API partners which may lead to rate limiting or denylisting of our API from their endpoints.

## Description

We should reduce the unauthenticated thumbnail throttling to something like `150/min`. This seems a reasonable amount for anyone building a user interface that displays thumbnails on-demand, which is more in-line with our intended use cases for the API. 

The throttling for _authenticated_ accounts should be higher, perhaps `500/min`?

## Additional context

We have also discussed implementing CSRF tokens on the front-end to solve this problem, which would be used in conjunction with this approach but with the unauthenticated throttling even more aggressive.

## Implementation

krysal:
## Fixes

Fixes #863 by @zackkrida

## Description

As title says, adds the `AnonThumbnailRateThrottle` and `OAuth2IdThumbnailRateThrottle` classes.

## Testing Instructions

1. Set the following variables in your `end.docker` file
```
DISABLE_GLOBAL_THROTTLING=False
THROTTLE_ANON_THUMBS=2/minute
# You may also need to add "localhost" to the ALLOWED_HOSTS
```
2. `just up`
3. Try viewing 3 thumbnails (or just one refreshing more than twice)
4. You should get throttled in this endpoint
5. Repeat for the authenticated (**verified**) version 

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/utils/throttle.py
+++ b/api/catalog/api/utils/throttle.py
@@ -61,6 +61,10 @@ class OneThousandPerMinute(AbstractAnonRateThrottle):
     rate = "1000/min"
 
 
+class OneHundredFiftyPerMinute(AbstractAnonRateThrottle):
+    rate = "150/minute"
+
+
 class OnePerSecond(AbstractAnonRateThrottle):
     rate = "1/second"
 
@@ -88,6 +92,11 @@ class AbstractOAuth2IdRateThrottle(SimpleRateThrottle, metaclass=abc.ABCMeta):
             return None
 
         return self.cache_format % {"scope": self.scope, "ident": ident}
+
+
+class OAuth2IdThumbsRateThrottle(AbstractOAuth2IdRateThrottle):
+    applies_to_rate_limit_model = "standard"
+    scope = "oauth2_client_credentials_thumbs"
 
 
 class OAuth2IdSustainedRateThrottle(AbstractOAuth2IdRateThrottle):
--- a/api/catalog/api/utils/throttle.py
+++ b/api/catalog/api/utils/throttle.py
@@ -53,16 +53,16 @@ class SustainedRateThrottle(AbstractAnonRateThrottle):
     scope = "anon_sustained"
 
 
+class AnonThumbnailRateThrottle(AbstractAnonRateThrottle):
+    scope = "anon_thumbnail"
+
+
 class TenPerDay(AbstractAnonRateThrottle):
     rate = "10/day"
 
 
 class OneThousandPerMinute(AbstractAnonRateThrottle):
     rate = "1000/min"
-
-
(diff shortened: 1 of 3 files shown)


api/catalog/api/utils/throttle.py
```diff
@@ -61,6 +61,10 @@ class OneThousandPerMinute(AbstractAnonRateThrottle):
     rate = "1000/min"
 
 
+class OneHundredFiftyPerMinute(AbstractAnonRateThrottle):
+    rate = "150/minute"
```
sarayourfriend, line 65: It _mayyyy_ be a good idea (given we're kind of guessing at a good rate limit for this) to make this an environment variable and change the class have a specific name for thumbs:

```suggestion
class AnonThumbnailThrottle(AbstractAnonRateThrottle):
    rate = config("ANON_THUMBNAIL_THROTTLE_RATE", "150/minute")
```

What do you think? This way if we find that it needs to be adjusted we don't have to make a code change, cut a new release, and then redeploy, it'd just be a redeploy.

sarayourfriend: Okay, I think I have my final suggested change here to make it configurable via the environment:

1. Rename the throttle to AnonThumbnailThrottle and add a scope
2. Add the scope to the `DEFAULT_THROTTLE_RATES` in `settings.py` and set the value to an environment variable similar to how the anon throttle rates are set

It's a minor change in the context of this PR but it puts the configuration added here more in line with the rest of our throttles.

krysal: I applied the suggested changes but doesn't seem to work well for authenticated requests, so I will convert it to draft.

krysal:
Not sure what I was doing before but I tested it again and it works, lol. The existing unit tests seem to cover the newly classes added, as they're part of `AbstractOAuth2IdRateThrottle` & `AbstractAnonRateThrottle` subclasses, robust tests! :)

dhruvkb (approved):
Good change, should we remove the very numeric throttle classes (`OneThousandPerMinute`/`OnePerSecond` etc.) now that they're unused?

krysal:
@dhruvkb Fair point, I agree on removing them in another PR for clean-up.

PR #865, opened 2022-08-09
Only backfill dead links if at least one on first page was not dead

### LINKED ISSUE #855: Only fill in pages if at least one link from first page is not dead
## Problem

If we make a request to ES for 20 items and find that any of them are dead, we attempt to "fill the page in" with subsequent "deeper page" queries. If _all_ the links for a query are dead, this means we will slam ES with tons of fast queries to try to fill in the page.

## Description

Kudos @obulat for the idea. We should not try to fill pages if all of the links on the first page are dead.

## Additional context
We are also working on excluding providers that account for a large number of dead links.

## Implementation

sarayourfriend:
## Fixes

Fixes #855 by @sarayourfriend

## Description

If after validating dead links, there are zero results, assume any subsequent deeper paginations of the same query will not yield better returns. The lines changed is a lie: it's like 97% changes in the unit tests. The actual change required to make this work is two SLOC with some explanatory comments.

## Testing Instructions

Check out the unit tests. To test this locally the best thing to do would be to use OpenSnitch or some other firewall to block outbound requests to Flickr. Then observe the logs and ensure that only a single query is sent to Elasticsearch. Note: this will be hard to do if on Linux with Docker as root (you'll need to mess with a bunch of iptables configurations). Instead, you can just modify the code to set the status code on the responses manually using this patch:

```diff
diff --git a/api/catalog/api/utils/validate_images.py b/api/catalog/api/utils/validate_images.py
index 6ad23f6f..c4dcbb42 100644
--- a/api/catalog/api/utils/validate_images.py
+++ b/api/catalog/api/utils/validate_images.py
@@ -58,7 +58,7 @@ def validate_images(query_hash, start_slice, results, image_urls):
         # Response didn't arrive in time. Try again later.
         else:
             status = -1
-        to_cache[cache_key] = status
+        to_cache[cache_key] = 404
 
     thirty_minutes = 60 * 30
     twenty_four_hours_seconds = 60 * 60 * 24
@@ -83,7 +83,7 @@ def validate_images(query_hash, start_slice, results, image_urls):
     for idx, url in enumerate(to_verify):
         cache_idx = to_verify[url]
         if verified[idx] is not None:
-            cached_statuses[cache_idx] = verified[idx].status_code
+            cached_statuses[cache_idx] = 404
         else:
             cached_statuses[cache_idx] = -1
```

Then you can make any request and all links will get cached as dead. You should get responses back that have empty results now from queries that previously gave back results with the local data (like `source=flickr`).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/utils/validate_images.py
+++ b/api/catalog/api/utils/validate_images.py
@@ -10,6 +10,14 @@ from catalog.api.utils.dead_link_mask import get_query_mask, save_query_mask
 parent_logger = logging.getLogger(__name__)
 
 
+CACHE_PREFIX = "valid:"
+
+
+def _get_cached_statuses(redis, image_urls):
+    cached_statuses = redis.mget([CACHE_PREFIX + url for url in image_urls])
+    return [int(b.decode("utf-8")) if b is not None else None for b in cached_statuses]
+
+
 def validate_images(query_hash, start_slice, results, image_urls):
     """
     Make sure images exist before we display them. Treat redirects as broken
@@ -28,11 +36,7 @@ def validate_images(query_hash, start_slice, results, image_urls):
     start_time = time.time()
     # Pull matching images from the cache.
     redis = django_redis.get_redis_connection("default")
-    cache_prefix = "valid:"
-    cached_statuses = redis.mget([cache_prefix + url for url in image_urls])
-    cached_statuses = [
-        int(b.decode("utf-8")) if b is not None else None for b in cached_statuses
-    ]
+    cached_statuses = _get_cached_statuses(redis, image_urls)
     logger.debug(f"len(cached_statuses)={len(cached_statuses)}")
     # Anything that isn't in the cache needs to be validated via HEAD request.
     to_verify = {}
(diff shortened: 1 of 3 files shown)


api/catalog/api/utils/validate_images.py
```diff
@@ -10,6 +10,14 @@
 parent_logger = logging.getLogger(__name__)
 
 
+CACHE_PREFIX = "valid:"
+
+
+def _get_cached_statuses(redis, image_urls):
+    cached_statuses = redis.mget([CACHE_PREFIX + url for url in image_urls])
+    return [int(b.decode("utf-8")) if b is not None else None for b in cached_statuses]
```
sarayourfriend, line 18: This change is to facilitate mocking the cached statuses in the test. Mostly this is to avoid the status cache being prepopulated based on other tests request/response cycle.

AetherUnbound:
@krysal do you mind reviewing this in my stead while I run MSR this week?

api/catalog/api/utils/validate_images.py
krysal: It's also welcome modularization!

krysal (approved):
Tests look good to me 👍 We could also try with some of the problematic providers (e.g. EOL).

zackkrida (approved):
I believe I'm seeing the correct behavior! I ran this query:

```
http http://localhost:8000/v1/images/\?source\=flickr\&page\=5
```

before and after applying the patch to `validate_images.py`, and the later resulted in a single ES query followed by many `Deleting broken image from results id=... status=404` calls.

Issue #843, opened 2022-08-02
Remove `get_api_exception` and replace with built in DRF exceptions


sarayourfriend:
## Description

The bespoke `get_api_exception` function recreates built in DRF functionality and makes searching for particular exceptions more difficult. Instead of searching for `raise NotFound` or `raise <built-in-DRF-exception>`, the typical way to find code that explicitly raises exceptions, you'd have to know about the `get_api_exception` function and search for usages with the specific HTTP status code (which has to be hand written by the user each time, meaning they have to remember what the status code is).

It also obscures stack traces because all errors created by `get_api_exception` have the same name (only solvable by further meta programming using the `type` constructor instead of a closed class): https://sentry.io/share/issue/abf49297eb2749e889eff15e4d37e151/

In short: we can eliminate code by removing this, be more in line with default DRF functionality and the community.

1. Replace all usages of `get_api_exception` function with the appropriate built-in DRF exception from `rest_framework.exceptions`. For any that are following the `try: queryset.get(); except Model.DoesNotExist: ...` pattern, replace it with [`get_object_or_404`](https://github.com/django/django/blob/3b79dab19a2300a4884a3d81baa6c7c1f2dee059/django/shortcuts.py#L64).
1. Remove the `get_api_exception` function

AbhiYHub:
Hello @sarayourfriend, I am new to open source and I would like to work on this.

sarayourfriend:
Hi @AbhiYHub! That's great, thank you for your willingness to contribute :slightly_smiling_face: I've assigned the issue to you. If you have any questions here or run into issues during implementation, please feel free to ping me or `@WordPress/openverse-maintainers` for help.

abhi2296:
Hello @sarayourfriend !! I am looking to start my journey in Open Source projects. Could you please assign this issue to me?

AbhiYHub:
Hello @sarayourfriend and @dhruvkb, Apologies for the delay. I have solved the issue and also ran the test cases. All the test cases have passed successfully. I would request if you can reassign me the task so that I could submit the solution.

dhruvkb:
@AbhiYHub please push your solution as a PR. Thanks!

PR #868, opened 2022-08-09
Move page and page_size query param validation into serializer

sarayourfriend:
## Fixes

Related to #857 and #843 

## Description

In this PR I move scattered query parameter validation into the query serializer, where it probably belongs. This has two benefits:

1. It removes code in favour of using built-in Django Rest Framework serializer validation
2. It removes usages of the bespoke `get_api_exception`
3. It reduces instances of a loose pattern of scattered validation where parameter validation happens multiple times and deep into the runtime of a search query, rather than once at the top (where it is expected to happen and easier to find)

None of these changes change the API surface.

The lines changed is also deceptive on this PR as something like 85 of the lines added or changed were in the tests.

## Testing Instructions

Ensure CI passes. Checkout the branch, set up the local API following the README instructions and make some smoke test requests to ensure everything works as expected.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/controllers/search_controller.py
```diff
@@ -425,10 +424,9 @@ def _get_result_and_page_count(
     :return: Result and page count.
     """
     result_count = response_obj.hits.total.value
-    natural_page_count = int(result_count / page_size)
-    if natural_page_count % page_size != 0:
-        natural_page_count += 1
-    page_count = min(natural_page_count, MAX_TOTAL_PAGE_COUNT)
```
sarayourfriend, line 431: This clamping is moved up into the paginator and out of the search logic.

sarayourfriend:
Keeping this PR as a draft as it should not be merged until we've fixed the production ES CPU usage.

dhruvkb:
@sarayourfriend since the ES CPU usage is under control now, should we push to update, undraft and get this PR merged?

dhruvkb:
Diff before updating to `main`, preserved for self-reference.

image

api/catalog/api/serializers/media_serializers.py
```diff
@@ -161,10 +175,30 @@ def validate_title(self, value):
...
+            if is_anonymous
+            else settings.MAX_AUTHED_PAGE_SIZE
+        )
+
+        validator = MaxValueValidator(
+            max_value,
+            message=serializers.IntegerField.default_error_messages["max_value"].format(
+                max_value=max_value
+            ),
+        )
```
dhruvkb, line 189: Can be extracted into a separate validator, and applied to the field using the [`validators` argument](https://www.django-rest-framework.org/api-guide/fields/#validators). That removes the need for the `try`-`except` handling below.

api/test/unit/controllers/test_search_controller.py
```diff
@@ -41,8 +40,7 @@
         (20, 5, 5, (20, 5)),
         # Fewer hits than page size, but result list somehow differs, use that for count
         (48, 20, 50, (20, 0)),
-        # Page count gets truncated always
-        (5000, 10, 10, (5000, MAX_TOTAL_PAGE_COUNT)),
+        (5000, 10, 10, (5000, 5000 / 10)),
```
dhruvkb, line 43: Need a comment here, explaining what this test case is testing for.

dhruvkb (approved):

dhruvkb (approved):
This can be compared against an earlier (approved) implementation in #778, so this PR should be too!

api/test/unit/controllers/test_search_controller.py
krysal: I believe we can remove this case, this function does not truncate the `page_count` anymore as that logic relies on the serializer now.

api/catalog/api/serializers/media_serializers.py
krysal: Great! We can do that on another refactor to provide a more specific error. Currently, it says `"Ensure this value is less than or equal to 20."` when before it was hinted that that was just the limit for unauthed requests: `"Page size must be between 1 & 20 for unauthenticated requests."`

api/test/unit/controllers/test_search_controller.py
krysal, line 43:
```suggestion
```

zackkrida (approved):
Looks great using the test instructions from the prior PR.

api/catalog/api/controllers/search_controller.py
krysal: So glad to see this function simplified even more!

krysal (approved):
Thank you both @sarayourfriend & @dhruvkb for working on this! Looks great 💯

PR #875, opened 2022-08-11
Do not 500 when upstream provider is unavailable

### LINKED ISSUE #847: Thumbnail endpoint returns 500 if the upstream image URL 404s
## Description

I was looking through production logs to see where some 500 errors were coming from and I found a handful of logs that look like this:

Field | Value
-- | --
`@ingestionTime` | 1659625225674
`@log` | 140733703834:/prod/api/nginx
`@logStream` | i-05a7072edd44e97d9
`@timestamp` | 1659625221900
body_bytes_sent | 66
host_header | api.openverse.engineering
http_referrer | https://search.openverse.engineering/
http_user_agent | Mozilla/5.0  (Linux; Android 6.0.1; Nexus 5X Build/MMB29P) AppleWebKit/537.36  (KHTML, like Gecko) Chrome/103.0.5060.134 Mobile Safari/537.36  (compatible; Googlebot/2.1; +http://www.google.com/bot.html)
http_x_forwarded_for | [redacted]
remote_addr | [redacted]
request | GET /v1/images/81954166-168e-41a6-bff5-eb966cb940ec/thumb/ HTTP/1.1
request_time | 1.175
status | 500
time_local | 04/Aug/2022:15:00:21 +0000
upstream_response_time | 1.175

Indeed, if you visit `https://api.openverse.engineering/v1/images/81954166-168e-41a6-bff5-eb966cb940ec/thumb/` you will get a 500 back with a message saying that the upstream request resulted in a 404. I think these should just 404 all the way, instead of resulting in a 500. 

I wonder if these should also be marked as dead links? It should already be anyway because the upstream url 404s and this is how we detect dead links during the search request, but clearly some still get through :thinking: 

## Reproduction

1.  Visit https://api.openverse.engineering/v1/images/81954166-168e-41a6-bff5-eb966cb940ec/thumb/
2.  See the 500 response code with the 404 body details.

## Resolution

sarayourfriend:
## Fixes

Fixes #847 by @sarayourfriend 

## Description

Fixes the issue by returning a 424: https://httpstatus.in/424/

> The method could not be performed on the resource because the requested action depended on another action and that action failed.

This makes sense for this error because we're not encountering a problem ourselves, rather there is an issue with the dependency (the upstream provider).

Still capture the exception in Sentry so that it can be addressed and tracked.

## Testing Instructions

Checkout this branch and `just down && just up` to get the fresh code running. Block a request to a provider using a firewall and see that 424 is returned. 

Check out the unit test. I tried to add it to the existing media integration suite but that wasn't possible because it didn't allow for mocking the `urlopen` function as that whole suite relies on `requests` instead of the django test client. It makes actual http requests to the running process which cannot be patched (as far as I can tell).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/test/unit/views/media_views_test.py
+++ b/api/test/unit/views/media_views_test.py
@@ -1,0 +1,40 @@
+from test.factory.models.image import ImageFactory
+from unittest import mock
+from urllib.error import HTTPError
+
+from rest_framework.test import APIClient
+
+import pytest
+
+from catalog.api.models.image import Image
+
+
+@pytest.fixture
+def api_client() -> APIClient:
+    return APIClient()
+
+
+@pytest.fixture
+def image() -> Image:
+    return ImageFactory.create()
+
+
+@pytest.mark.django_db
+def test_thumb_error(api_client, image):
+    error = None
+
+    def urlopen_503_response(url, **kwargs):
+        nonlocal error
+        error = HTTPError(url, 503, "Bad error upstream whoops", {}, None)
+        raise error
+
+    with mock.patch(
+        "catalog.api.views.media_views.urlopen"
+    ) as urlopen_mock, mock.patch(
+        "catalog.api.views.media_views.capture_exception", autospec=True
+    ) as mock_capture_exception:
+        urlopen_mock.side_effect = urlopen_503_response
+        response = api_client.get(f"/v1/images/{image.identifier}/thumb/")
+
+    assert response.status_code == 424
+    mock_capture_exception.assert_called_once_with(error)
--- a/api/catalog/api/views/media_views.py
+++ b/api/catalog/api/views/media_views.py
@@ -9,8 +9,11 @@ from django.conf import settings
 from django.http.response import HttpResponse
 from rest_framework import status
 from rest_framework.decorators import action
+from rest_framework.exceptions import APIException
 from rest_framework.response import Response
 from rest_framework.viewsets import ReadOnlyModelViewSet
+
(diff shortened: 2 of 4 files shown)


api/test/unit/views/media_views_test.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@pytest.mark.django_db
+def test_thumb_error(api_client, image):
+    error = None
+
+    def urlopen_503_response(url, **kwargs):
+        nonlocal error
+        error = HTTPError(url, 503, "Bad error upstream whoops", {}, None)
+        raise error
```
AetherUnbound, line 29: Why does this need to be set to `None` initially, then set to `HTTPError` within the function? Is that to test that the mock's side effect does indeed get called? I imagine we could just declare the `error` at the beginning of the test and use it throughout.

AetherUnbound (approved):
Looks good, TIL about 424! Just one question about the test.

api/test/unit/views/media_views_test.py
sarayourfriend: The error needs the `url` that is passed to the `urlopen` function, so it can't be initialized outside of this handler.

AetherUnbound: Ah! Got it, thanks 🙂

dhruvkb (approved):

PR #878, opened 2022-08-11
Extend valid link cache time to 30 days and make cache expiration times minimally configurable

### LINKED ISSUE #867: Increase the dead links cache expiration timeout
## Problem

Currently, every front-end search results in requesting the provider image URL (the direct URL) twice: once for the API dead links filtering and once for the thumbnail generation.
The thumbnails are cached for a period of around a year (to be confirmed), and the link validation cache expires after 24 hours for active links and 120 days for broken links. This means we're re-requesting popular valid images every 24 hours. We could increase the cache expiration timeout to reduce the number of network requests to the providers (which have often been 429-ed lately).

## Description

We could set the cache expiration timeout for active links based on an environment variable to make it easily tweakable.
I think increasing the active link cache TTL to **one week** could be a good balance between keeping the link status up-to-date and reducing the number of requests. What do you think about the timeout period, @openverse-team?

## Alternatives

Another option for reducing the number of requests to providers could be to request the thumbnail endpoint first, and if the request headers show that the response header comes from the Cloudflare cache, consider the link not dead. If, however, the response is coming from the proxy and does not have a 200 status, then we can consider the link dead.

## Implementation

sarayourfriend:
## Fixes

Fixes #867 by @obulat 

## Description

Makes the link validation cache expiration times minimally configurable and increases the valid link cache time to 30 days.

I have another patch that introduces more flexible configuration by status code, in case that is something we want to explore.

## Testing Instructions

Checkout this branch and `just down && just up` to ensure you get the latest code. Smoke test the app around link validation and confirm that things are getting cached. You can do so by observing the log output. For example, a request to `/v1/images/` will output a bunch of warning logs about making insecure requests to staticflickr the first time but should not output any on the second request.

If you want to confirm the cache times more explicitly, you'll need to upgrade redis locally to version 7 so that you can use the `EXPIRETIME` command. To do this, update the `cache` image tag from `4.x.x` to `7`:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index beed92cc..11ebea67 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,7 @@ services:
       test: "pg_isready -U deploy -d openledger"
 
   cache:
-    image: redis:4.0.10
+    image: redis:7
     ports:
       - "6379:6379"
```

Then run `redis-cli` locally and try the following:

```
# copy one of these keys
> keys "valid*"
# find the status code cached for that link
> get "<copied key>"
# see the expire time set for the key
> expiretime "<copied key>"
```

expiretime returns the Unix timestamp from origin so you'll have to do some math to confirm that it matches your expectation based on when the key was inserted. To clear your local Redis, run `redis-cli flushall`. This is useful for trying this process a couple times to confirm it's working as expected.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/utils/validate_images.py
+++ b/api/catalog/api/utils/validate_images.py
@@ -3,6 +3,7 @@ import time
 
 import django_redis
 import grequests
+from decouple import config
 
 from catalog.api.utils.dead_link_mask import get_query_mask, save_query_mask
 
@@ -16,6 +17,10 @@ CACHE_PREFIX = "valid:"
 def _get_cached_statuses(redis, image_urls):
     cached_statuses = redis.mget([CACHE_PREFIX + url for url in image_urls])
     return [int(b.decode("utf-8")) if b is not None else None for b in cached_statuses]
+
+
+def _get_expiry(status, default):
+    return config(f"LINK_VALIDATION_CACHE_EXPIRY__{status}", default=default, cast=int)
 
 
 def validate_images(query_hash, start_slice, results, image_urls):
@@ -69,14 +74,17 @@ def validate_images(query_hash, start_slice, results, image_urls):
         # Cache successful links for a day, and broken links for 120 days.
         if status == 200:
             logger.debug("healthy link " f"key={key} ")
-            pipe.expire(key, twenty_four_hours_seconds)
+            expiry = _get_expiry(200, twenty_four_hours_seconds)
         elif status == -1:
             logger.debug("no response from provider " f"key={key}")
             # Content provider failed to respond; try again in a short interval
-            pipe.expire(key, thirty_minutes)
+            expiry = _get_expiry(-1, thirty_minutes)
         else:
             logger.debug("broken link " f"key={key} ")
-            pipe.expire(key, twenty_four_hours_seconds * 120)
+            expiry = _get_expiry("DEFAULT", twenty_four_hours_seconds * 120)
+
+        pipe.expire(key, expiry)
+
     pipe.execute()
 
     # Merge newly verified results with cached statuses
--- a/api/catalog/api/utils/validate_images.py
+++ b/api/catalog/api/utils/validate_images.py
@@ -74,11 +74,11 @@ def validate_images(query_hash, start_slice, results, image_urls):
         # Cache successful links for a day, and broken links for 120 days.
         if status == 200:
             logger.debug("healthy link " f"key={key} ")
-            expiry = _get_expiry(200, twenty_four_hours_seconds)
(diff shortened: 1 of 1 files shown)


api/catalog/api/utils/validate_images.py
```diff
@@ -18,6 +19,10 @@ def _get_cached_statuses(redis, image_urls):
     return [int(b.decode("utf-8")) if b is not None else None for b in cached_statuses]
 
 
+def _get_expiry(status, default):
+    return config(f"LINK_VALIDATION_CACHE_EXPIRY__{status}", default=default, cast=int)
```
dhruvkb, line 23: This might be bikeshedding but the `__` in the name looks weird.

api/catalog/api/utils/validate_images.py
```diff
@@ -69,14 +74,17 @@ def validate_images(query_hash, start_slice, results, image_urls):
         # Cache successful links for a day, and broken links for 120 days.
         if status == 200:
             logger.debug("healthy link " f"key={key} ")
-            pipe.expire(key, twenty_four_hours_seconds)
+            expiry = _get_expiry(200, twenty_four_hours_seconds * 30)
```
dhruvkb, line 77: I think a variable called `thirty_days_seconds` might make this clearer.

dhruvkb (approved):
This is great! All I have are bikeshedding comments.

api/catalog/api/utils/validate_images.py
sarayourfriend: It's a not-uncommon separator for environment variables: https://github.com/WordPress/openverse-catalog/blob/f5de3e2a5f9c9da62f7e47e9c86b566e3e345123/env.template#L11

dhruvkb: It's new to the API. But it's cool 👍. Could you move this to `settings` (so that almost all env reading happens in one place) and also add it to the `env.template` file?

dhruvkb (approved):

api/catalog/api/utils/validate_images.py
sarayourfriend: Move the `_get_expiry` function?

dhruvkb: Oh, my bad. I didn't realise that the `default=` argument to `config` was set by the caller to `_get_expiry`. Please ignore my suggestion.

Issue #866, opened 2022-08-09
Thumbnail tests make actual upstream provider requests instead of using mocked image responses


sarayourfriend:
## Description

If the upstream provider (in the case of our current test configuration, Flickr) returns an error, then our tests will fail regardless of whether anything actually changed to make them fail. Right now Flickr is returning 503 and this causes the tests to fail.

## Reproduction

1.  Mock the `url` value of an image that the thumbnail is requested with `https://httpbin.org/status/503` and see the test fail. You could also use a firewall like OpenSnitch or LittleSnitch to block the request on a computer level and see the requests happen during the test and fail them.

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

sarayourfriend:
Re-prioritizing this to work on it now because it is causing CI to fail due to inconsistencies in the upstream provider behaviours (which we should not depend on anyway).

sarayourfriend:
@WordPress/openverse-maintainers I'd like to propose the following solution for this:

1. Add minio to the local docker-compose and add a couple sample images and audio files from it
2. Change all the sample data `url`s to point to the local minio version

Problem with this is that it will make running the app locally outside of docker even more difficult. I think this is fine though, as it's not an approach we recommend or support anyway.

dhruvkb:
@sarayourfriend I support this solution. Minio is great and while we do somewhat support running the API outside Docker, we still need the scaffolding services to be provisioned by Docker Compose so adding additional services should be OK.

AetherUnbound:
I love the idea of employing minio here! And agreed, I think requiring that the API run inside of Docker is fine in my mind (we're already doing that for the Catalog anyway).

sarayourfriend:
I am actually sceptical of the current approach and would like to investigate it more. I can see the value in the integration style tests, but given the infrastructural difficulties and the absolute need to define hard boundaries in some areas (like preventing all outbound network requests, maybe through something like HTTPretty or the like) I wonder if it's sufficient to move our sample data into a Python script that uses the same factories that the tests use. This has two benefits in my book:

1. It forces the factories to be up to date and valid and reflect the actual models correctly
2. It makes the test data very clear and easy to understand and maintain

[...]

PR #884, opened 2022-08-15
Prevent tests from making network requests for link validation and fix potentially non-unique factory props

### LINKED ISSUE #883: Prevent unique field clashes in unit test model factories
## Problem

https://github.com/WordPress/openverse-api/runs/7790427928?check_suite_focus=true

This test failed because the `url` property on a factory-generated model was not unique.

## Description

We can use https://factoryboy.readthedocs.io/en/stable/recipes.html#forcing-the-sequence-counter to add an incrementing value to the fields that need to be unique on the factories.

## Implementation

sarayourfriend:
## Fixes

Fixes #883 by @sarayourfriend 
Related to #866 by @sarayourfriend

## Description

It turns out the issue with the flaky tests is two fold. We do indeed have the chance for a run to have non-unique urls that clash with database constraints. That is fixed in this PR through adding a new Faker provider `globally_unique_url` that appends a UUID to the URL that Faker uses. Faker's approach does include randomness but it has a small search space, so clashes are easy to come by (especially when generating hundreds of fixtures like we do in some of our tests):

```py
    def url(self, schemes: Optional[List[str]] = None) -> str:
        """
        :param schemes: a list of strings to use as schemes, one will chosen randomly.
            If None, it will generate http and https urls.
            Passing an empty list will result in schemeless url generation like "://domain.com".
        :return: a random url string.

        """
        if schemes is None:
            schemes = ["http", "https"]

        pattern: str = f'{self.random_element(schemes) if schemes else ""}://{self.random_element(self.url_formats)}'

        return self.generator.parse(pattern)
```
 
There is an additional error that the audio fixture has `results_count` at 0 which means the fixture data has all dead links on the first page. As of #855 we no longer back fill dead links if the first page is empty from link validation. That means we need to force the first page to have all valid links, from the perspective of the `validate_images` function. So I'm forcing link statuses into redis before the test and then removing them after the test.

## Testing Instructions

Check the CI test runs and run `just api-test` locally to verify the fix.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/test/audio_integration_test.py
+++ b/api/test/audio_integration_test.py
@@ -33,25 +33,49 @@ from catalog.api.utils.validate_images import CACHE_PREFIX
 
 
 @pytest.fixture
-def force_first_page_audio_validity():
-    first_page = requests.get(
-        f"{API_URL}/v1/audio?filter_dead=False", verify=False
-    ).json()
-    statuses = {f"{CACHE_PREFIX}{item['url']}": 200 for item in first_page["results"]}
-    with get_redis_connection() as redis:
-        redis.mset(statuses)
+def force_result_validity():
+    statuses = {}
 
-    yield
+    def force_validity(query_response):
+        nonlocal statuses
+        new_statuses = {
+            f"{CACHE_PREFIX}{item['url']}": 200 for item in query_response["results"]
+        }
+        statuses |= new_statuses
+        with get_redis_connection() as redis:
+            redis.mset(new_statuses)
+
+    yield force_validity
 
     with get_redis_connection() as redis:
         redis.delete(*list(statuses.keys()))
 
 
 @pytest.fixture
-def audio_fixture(force_first_page_audio_validity):
-    res = requests.get(f"{API_URL}/v1/audio", verify=False)
+def audio_fixture(force_result_validity):
+    res = requests.get(
+        f"{API_URL}/v1/audio/", data={"filter_dead": False}, verify=False
+    )
+    parsed = res.json()
+    force_result_validity(parsed)
     assert res.status_code == 200
-    parsed = json.loads(res.text)
+    return parsed
+
+
+@pytest.fixture
+def jamendo_audio_fixture(force_result_validity):
+    """
+    Thumbnail tests must use Jamendo results because the Wikimedia
+    sample audio results do not have thumbnails.
+    """
+    res = requests.get(
+        f"{API_URL}/v1/audio/",
+        data={"source": "jamendo", "filter_dead": False},
+        verify=False,
+    )
+    parsed = res.json()
+    force_result_validity(parsed)
+    assert res.status_code == 200
     return parsed
 
 
@@ -98,8 +122,8 @@ def test_audio_stats():
     stats("audio")
 
 
-def test_audio_thumb(audio_fixture):
-    thumb(audio_fixture)
+def test_audio_thumb(jamendo_audio_fixture):
+    thumb(jamendo_audio_fixture)
 
 
 def test_audio_detail_without_thumb():
@@ -117,16 +141,16 @@ def test_audio_search_without_thumb():
     assert parsed["results"][0]["thumbnail"] is None
 
 
-def test_audio_thumb_compression(audio_fixture):
-    thumb_compression(audio_fixture)
+def test_audio_thumb_compression(jamendo_audio_fixture):
+    thumb_compression(jamendo_audio_fixture)
 
 
-def test_audio_thumb_webp(audio_fixture):
-    thumb_webp(audio_fixture)
+def test_audio_thumb_webp(jamendo_audio_fixture):
+    thumb_webp(jamendo_audio_fixture)
 
 
-def test_audio_thumb_full_size(audio_fixture):
-    thumb_full_size(audio_fixture)
+def test_audio_thumb_full_size(jamendo_audio_fixture):
+    thumb_full_size(jamendo_audio_fixture)
 
 
 def test_audio_report(audio_fixture):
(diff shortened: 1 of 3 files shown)


api/test/audio_integration_test.py
```diff
@@ -27,11 +27,29 @@
...
+    yield
+
+    with get_redis_connection() as redis:
+        redis.delete(*list(statuses.keys()))
 
 
 @pytest.fixture
-def audio_fixture():
-    res = requests.get(f"{API_URL}/v1/audio?q=", verify=False)
+def audio_fixture(force_first_page_audio_validity):
```
AetherUnbound, line 51: I'm looking into this thumbnail issue a bit more - the invocation of this fixture indeed causes the thumbnails to be `null`. I switched from `main` to this branch and queried the API a few times and saw that the thumbnails still appeared. As soon as I ran `j api-test -k test_audio_thumb` the thumbnails all became `null`.

AetherUnbound: Okay now I'm not so sure 🙈 I ran `flushall` on the redis cache and the thumbnails are still `null`. I'll continue to look.

sarayourfriend: The link validation cache _shouldn't_ have anything to do with it.

Curiously, this problem doesn't appear to affect the CI test run as it's passing fine here :thinking:

sarayourfriend: Also, if you look at the sample data, the thumbnails are indeed set to null for all audio, so it's not a mistake there. The curious thing is why it ever passed (and why is it passing in CI? :woozy_face:)

AetherUnbound: Okay I'm seeing the following behavior on `main` after I run `just recreate`:
1. First hit to http://localhost:8000/v1/audio/ - no thumbnails
2. Second hit to http://localhost:8000/v1/audio/ - no _results_
3. Third hit to http://localhost:8000/v1/audio/ - results again with thumbnails now

I've been able to reproduce this a few times now. Judging from the logs, it looks like it's clearing out the Wikimedia results (which might not have thumbnails?), and the end result is only Jamendo results.

AetherUnbound: Oh dear, it actually looks like `validate_images` is improperly named:
```
openverse-api-web-1  | [2022-08-15 22:26:20,217 - catalog.api.utils.validate_images.validate_images - 105][WARNING] [7d18c2d30cc64f309ed1dbc4cc5db9b4] Image validation failed due to rate limiting or blocking. url=https://upload.wikimedia.org/wikipedia/commons/8/88/LL-Q1860_%28eng%29-Vealhurl-walker.wav status=403 
```

It seems that it should probably be `validate_media`, I'll make an issue for this.

AetherUnbound: Okay this feels really inadvisable, but with this diff I was able to get all the tests to pass:

```diff
diff --git a/api/test/audio_integration_test.py b/api/test/audio_integration_test.py
index c732ceb5..248e6cc9 100644
--- a/api/test/audio_integration_test.py
+++ b/api/test/audio_integration_test.py
@@ -35,7 +35,7 @@ from catalog.api.utils.validate_images import CACHE_PREFIX
 @pytest.fixture
 def force_first_page_audio_validity():
     first_page = requests.get(
-        f"{API_URL}/v1/audio?filter_dead=False", verify=False
+        f"{API_URL}/v1/audio?filter_dead=False&source=jamendo", verify=False
     ).json()
     statuses = {f"{CACHE_PREFIX}{item['url']}": 200 for item in first_page["results"]}
     with get_redis_connection() as redis:
@@ -49,7 +49,7 @@ def force_first_page_audio_validity():
 
 @pytest.fixture
 def audio_fixture(force_first_page_audio_validity):
-    res = requests.get(f"{API_URL}/v1/audio", verify=False)
+    res = requests.get(f"{API_URL}/v1/audio?source=jamendo", verify=False)
     assert res.status_code == 200
     parsed = json.loads(res.text)
     return parsed
@@ -60,7 +60,7 @@ def test_search(audio_fixture):
 
 
 def test_search_category_filtering(audio_fixture):
-    search_by_category("audio", "music", audio_fixture)
+    search_by_category("audio", "news", audio_fixture)
     search_by_category("audio", "pronunciation", audio_fixture)
```

I have to wonder if the reason it's passing on CI is because Wikimedia isn't rate limiting whatever machine the tests are running on, but it _is_ rate limiting all of us 😮

AetherUnbound: We keep getting 403s from Wikimedia - that's supposed to be addressed by #877, perhaps that might be necessary in order to merge this too?

AetherUnbound: I tried adding this commit onto #877 and I'm still getting errors 😞 I think we may just need to be specific about which results we're querying (e.g. `source=jamendo`), at least until we're able to mock all of these requests. As far as I can tell, this all comes back to the dead link filtering. What do you think?

sarayourfriend: I fine with just filtering for a specific source for these tests in particular. I don't think these tests are set up in any way that makes sense to me. I'd rather use fixture/factory data that we actually encode into the tests themselves instead of sample data that is (a) difficult to maintain and (b) difficult to have insights into.

AetherUnbound (approved):
LGTM

obulat (approved):
Thank you for adding the docs about why we need Jamendo for thumbnails. Works well and CI is green!

PR #888, opened 2022-08-18
Ensure that license URL is not blank

dhruvkb:
## Fixes

Fixes WordPress/openverse-frontend#1656 by @sarayourfriend 

## Description

This PR ensures that for `Hits` without a `license_url`, the `license_url` field is not `None` but rather generated using the `license` and `license_version` information.

## Testing Instructions

The PR adds a unit test that covers this so passing CI should be adequate.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


obulat (changes requested):
When do the hits have no `license_url`? Will a full re-indexing add `license_url` to all of the media?

dhruvkb:
It should be fixed by a data refresh _if_ the catalog has done a backfilling of the missing field in ES.

dhruvkb:
The API does a solid job of ensuring that the license URL information is always sent for `Media` instances. But in the search results page, the instances are of the `Hit` class which does not have the `license_url` property. For `Hit` instances, the field either exists in ES (if added during the data refresh from the catalog) and is sent in the response or doesn't exist and is sent as `null`. This PR aims to fix that and ensure it's computed for the `Hit` instances too.

Ideally a full refresh from the catalog would populate the field on existing ES documents, and this code will never be executed. But for these situations to not repeat, the catalog needs to make some guarantees about what data will always exist (so we can remove the workarounds/computations for them) and what data might not exist (so we can handle it in the API).

sarayourfriend:
We do pass the Hits through a serializer that could pretty easily handle this kind of data transformation, for example in `to_representation`: https://github.com/WordPress/openverse-api/blob/db6f6ef24961cb0f821058d94e70b3dcc37ed70e/api/catalog/api/views/media_views.py#L96

api/test/unit/serializers/audio_serializers_test.py
```diff
@@ -0,0 +1,38 @@
+import uuid
+from unittest.mock import MagicMock
+
+from rest_framework.request import Request
+from rest_framework.test import APIRequestFactory
+
+import pytest
+
+from catalog.api.serializers.audio_serializers import AudioSerializer
```
sarayourfriend, line 9: Why is this only tested against the AudioSerializer?

Could we write parameterised tests for this that automatically test this behaviour against all subclasses?

api/test/unit/serializers/audio_serializers_test.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.fixture
+def req():
+    factory = APIRequestFactory()
+    request = factory.get("/")
+    request = Request(request)
+    return request
+
+
+@pytest.fixture
+@pytest.mark.django_db
```
sarayourfriend, line 21: Is this mark necessary?

```suggestion
```

sarayourfriend (changes requested):
Change looks good to me, but if we can write the tests against all subclasses of the MediaSerializer that would be ideal.

api/test/unit/serializers/audio_serializers_test.py
dhruvkb: That's a good idea. Implemented in 9e5a882.

api/test/unit/serializers/audio_serializers_test.py
dhruvkb: I added this when the test used a Django instances and forgot to remove it. Removed in 8c4fd4b.

api/test/unit/serializers/media_serializers_test.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    return hit
+
+
+@pytest.mark.parametrize(
+    "serializer_class",
+    [
+        AudioSerializer,
+        ImageSerializer,
+    ],
```
sarayourfriend, line 36: FWIW it's possible to use `__subclasses__` for this so it automatically updates itself when we add new serializers in the future.

api/test/unit/serializers/media_serializers_test.py
```diff
@@ -0,0 +1,44 @@
...
+)
+def test_audio_serializer_adds_license_url_if_missing(req, hit, serializer_class):
```
sarayourfriend, line 38:
```suggestion
def test_media_serializer_adds_license_url_if_missing(req, hit, serializer_class):
```

api/test/unit/serializers/media_serializers_test.py
dhruvkb: Subclasses are only registered when the serializers are imported. So as I see it, you'll still need to import the serializer files for the side-effect of registering subclasses.

Screenshot 2022-08-19 at 8 50 59 AM

obulat (approved):
Nice fix! When do we think this will be deployed?
I think we can drop the frontend change because API is where it should be fixed.

api/test/unit/serializers/media_serializers_test.py
sarayourfriend: Ahhh, they're not in the same modules. Rats!

PR #901, opened 2022-08-27
updates to new redoc link

### LINKED ISSUE #900: Broken ReDoc Link
## Description
qucikstart.md has a broken link to Api Redoc page. 

## Resolution

If the ReDoc documentation [is this one](https://api.openverse.engineering/v1/), i'll create a pr to update it. Please lmk if this is an old/outdated link as i couldn't find anything newer on the internet/slack

ramadanomar:
## Fixes
#900 by @ramadanomar . If i find any more dead/expired links , i'll try to update them here to prevent from filling out the pr/issue list with small changes like this.

### DIFF
--- a/api/docs/guides/quickstart.md
+++ b/api/docs/guides/quickstart.md
@@ -33,7 +33,7 @@ Ensure that you have installed `mkcert` (and the corresponding NSS tools). You c
    ```
 
 5. Point your browser to `http://localhost:50280`. You should be able to see the API documentation.
-   ![API ReDoc](/_static/api_redoc.png)
+   ![API ReDoc](https://api.openverse.engineering/v1/)
 
 6. Load the sample data. This could take a couple of minutes.
    ```bash


AetherUnbound (changes requested):
This actually points to an image which is referenced on the produced documentation page here: https://wordpress.github.io/openverse-api/guides/quickstart.html#steps

The image itself can be found here: https://github.com/WordPress/openverse-api/blob/main/api/docs/_static/api_redoc.png

While reviewing this though, I did notice that the API's README references the _frontend_ dev docs rather than the API! 

https://github.com/WordPress/openverse-api/blob/5fae4eb89c33f849fd41d2a890057858ec789278/api/README.md?plain=1#L5

This PR could adjust that link instead 😄

krysal:
@ramadanomar Could you keep working on this PR? The linked issue is fixed but it would be nice to correct the image that @AetherUnbound mentioned. It's okay if you are not available at the moment, we understand, just (preferably) let us know :)

PR #932, opened 2022-09-20
Catch struct errors in `watermark/` endpoint

### LINKED ISSUE #849: `piexif.dump` errors are not safely handled
## Sentry link

https://sentry.io/share/issue/a80d52de7f89436586ed0250cd0a32d2/

## Description

The call to `piexif.dump` should be wrapped in a `try/except` to prevent these errors in the watermark endpoint.

## Reproduction

Visit https://api-dev.openverse.engineering/v1/images/a913fde1-d524-4059-bd4f-9bd687578cc3/watermark/ to see an example of this failure.

krysal:
## Fixes

Fixes #849 by @sarayourfriend 

## Description

This PR:
- catches struct errors from `piexif.load` function. The last release of this library is from July 2019, so it will need a replacement sooner than later
- sets environment variables to enable the watermark endpoint locally and name the environment properly

I was writing a test but it wasn't failing even before the fix and it's taking too much time to mock. I haven't found a way to reproduce the error locally so hoping just to catch the exception to solve the Sentry alerts.

<details>

<summary>Non failing test</summary>

```python 
# Adding to file: api/test/unit/utils/watermark_test.py

import struct
from io import BytesIO
from PIL import Image
from unittest import mock
from catalog.api.utils.watermark import _open_image

...

def test_catch_struct_errors_from_piexif(requests):
    with mock.patch("PIL.Image.open") as open_mock, mock.patch("piexif.load") as load_mock:
        img_mock = Image.open(BytesIO(_MOCK_IMAGE_BYTES))
        img_mock.info["exif"] = "bad_info"
        open_mock.return_value = img_mock
        load_mock.side_effect = struct.error('unpack requires a buffer of 2 bytes')

        img, exif = _open_image("http://example.com/")

    assert img is not None
    assert exif is None
```

</details>

## Testing Instructions

Visit the `watermark/` endpoint of any image, eg. http://localhost:50280/v1/images/dba4f725-9370-4442-8db1-522e788d6c55/watermark/

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/utils/watermark.py
+++ b/api/catalog/api/utils/watermark.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import struct
 from enum import Flag, auto
 from io import BytesIO
 from textwrap import wrap
@@ -158,16 +159,17 @@ def _open_image(url):
         response = requests.get(url, headers=HEADERS)
         img_bytes = BytesIO(response.content)
         img = Image.open(img_bytes)
-        # Preserve EXIF metadata
-        if "exif" in img.info:
-            exif = piexif.load(img.info["exif"])
-        else:
-            exif = None
-        return img, exif
     except requests.exceptions.RequestException as e:
         capture_exception(e)
         logger.error(f"Error loading image data: {e}")
         return None, None
+
+    try:
+        # Preserve EXIF metadata
+        exif = piexif.load(img.info["exif"]) if "exif" in img.info else None
+        return img, exif
+    except struct.error:
+        return img, None
 
 
 def _print_attribution_on_image(img, image_info):
--- a/api/env.docker
+++ b/api/env.docker
@@ -4,6 +4,7 @@ DJANGO_SETTINGS_MODULE=catalog.settings
 DJANGO_SECRET_KEY="ny#b__$$f6ry4wy8oxre97&-68u_0lk3gw(z=d40_dxey3zw0v1"
 DJANGO_DEBUG_ENABLED=True
 
+ENVIRONMENT=development
 ALLOWED_HOSTS=api.openverse.engineering,api-dev.openverse.engineering,host.docker.internal
 
 REDIS_HOST=cache
@@ -18,3 +19,5 @@ UPSTREAM_DATABASE_PORT=5432
 SEMANTIC_VERSION=1.0.0
 
 ELASTICSEARCH_URL=es
+
+WATERMARK_ENABLED=True
--- a/api/test/unit/utils/watermark_test.py
+++ b/api/test/unit/utils/watermark_test.py
@@ -1,12 +1,16 @@
 import json
+import struct
 from dataclasses import dataclass
+from io import BytesIO
 from pathlib import Path
 from typing import Callable
+from unittest import mock
 
 import pytest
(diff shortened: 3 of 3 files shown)


sarayourfriend:
> The last release of this library is from July 2019, so it will need a replacement sooner than later

We could switch to Pillow for this, I would think. We already use it elsewhere, and it is a massively popular library. (For another issue, of course.)

sarayourfriend (changes requested):
Can you add a unit test to exercise this change? Otherwise it LGTM.

krysal:
@sarayourfriend 

> Can you add a unit test to exercise this change? Otherwise it LGTM.

As I mentioned at the beginning, I tried but couldn't trigger the same issue we are getting from Sentry 😅 Do you see anything that needs to change in the test I tried to include? It never fails so it's basically useless in that way. I used the "private" `_open_image` function because the public `watermark` was requiring more mocking, and the problematic code is in on the first one anyway.

sarayourfriend:
@krysal I totally glossed over the collapsed test code in the description, sorry.

I suspect the `Image.open` mock is incorrect. Because the module imports it as `from PIL import Image` you'd need to mock Image in the watermark module itself, I think? Python import mocks always take a bit of fiddling for me to figure out too :sweat: 

I think it's fine to test against the private method on this case, FWIW, given it's a weird edge case error. It'd be annoying if we refactored the method but we'll want to be careful with this one anyway.

sarayourfriend:
Taking a look at this now locally. I'm wrong about the mocks, you've got them correct. It's just doing the thing that I cannot figure out how to stop MagicMock from doing (which I never seem to remember it doing before), which is that instead of actually returning the `.return_value` set on `open` it's just returning a new MagicMock instance :angry: I ran into this before with something else. I think it has something to do with `auto_spec`. I'll see if I can figure it out.

sarayourfriend:
Okay, actually everything was fine _except_ a very subtle thing. Here's the test that works (I confirmed with `breakpoint` that it goes to the error handling, I wish there was a better way to verify the test was working as expected):

```py

def test_catch_struct_errors_from_piexif(requests):
    img_mock = Image.open(BytesIO(_MOCK_IMAGE_BYTES))
    img_mock.info["exif"] = "bad_info"

    with mock.patch("PIL.Image.open") as open_mock, mock.patch("piexif.load") as load_mock:
        open_mock.return_value = img_mock
        load_mock.side_effect = struct.error('unpack requires a buffer of 2 bytes')

        img, exif = _open_image("http://example.com/")

    assert img is not None
    assert exif is None
```

I guess the issue was that because we were mocking `PIL.Image.open`, the mocked version of the method was also being used to create the `img_mock`. Moving it outside the context manager of the `mock.patch` for that method fixes the test so that it actually exercises the error condition.

AetherUnbound (approved):
LGTM

sarayourfriend (approved):
:rocket:

stacimc (approved):
LGTM! Testing discussion is also very interesting and helpful :)

Issue #417, opened 2021-12-17
Ingestion server hangs after cleaning


AetherUnbound:
## Description

I've seen this issue a bunch as I've been running tests locally. It seems like the ingestion server hangs immediately after finishing the cleaning:

```
integration_ingestion_server_1  | 2021-12-17 23:49:59,173 INFO cleanup.py:306 - Cleaned all records in 0.4112739562988281 seconds
```

No other logs occur after this and the integration tests fail due to timeout. This seems super transient and is frequently resolved in future runs.

## Reproduction

1. Run `just recreate && just init` or `just ing-testlocal`
2. Get lucky

## Resolution

- [ ] 🙋 I would be interested in resolving this bug.

AetherUnbound:
I attempted to tackle this in #412 - I'm going to close this but we can re-open it if it reappears

PR #936, opened 2022-09-22
Register audio and image models consistently in admin panel

### LINKED ISSUE #919: Expose Django Admin view for mature audio
## Problem

Currently to flag an audio track as mature we have to manipulate the DB records using the Django ORM or directly through a psql/pgcli session. We can ease this job by adding a friendly view in the Django Admin.

## Description

Analogous to the "matureimage" views, add the list and creation form for `MatureAudio` objects.

## Implementation

dhruvkb:
## Fixes

Fixes #919 by @krysal 

## Description

This PR
- ensures that images and audio have all reporting models registered in the admin panel
- adds checks to ensure non-existent identifiers cannot be reported
- distributes concerns from the central `_Report` models to the `Deleted_` and `Mature_` models
- ensures that the changes to mature status are synced to Elasticsearch
- surface `mature` field in the details endpoint for WordPress/openverse-frontend#1822
- limits admin functionality to encourage correct usage of reporting flows
- adds documentation where it was lacking
- implements general refactors and makes code quality improvements

## Testing Instructions

0. Check out the PR and run all services.
1. Visit the Django admin panel.
2. See new models present.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


krysal (dismissed):
So nice how clean it looks the generalized code ✨

However, I can register an image uuid as a MatureAudio and vice-versa, can we prevent that from happening? Strangely the image/audio still shows as not mature after registration but that can be a bug for another issue.

api/catalog/api/models/media.py
```diff
@@ -219,34 +206,95 @@ def save(self, *args, **kwargs):
...
+    """the name of the ES index from ``settings.MEDIA_INDEX_MAPPING``"""
+
     identifier = models.UUIDField(
         unique=True, primary_key=True, help_text="The identifier of the deleted media."
     )
 
     class Meta:
         abstract = True
 
+    def _update_es(self, raise_errors: bool) -> models.Model:
```
sarayourfriend, line 233: Nice idea!

api/catalog/api/models/audio.py
```diff
@@ -175,6 +176,10 @@ class Audio(AudioFileMixin, AbstractMedia):
         help_text="JSON describing alternative files for this audio.",
     )
 
+    @property
+    def mature(self) -> bool:
+        return MatureAudio.objects.filter(identifier=self.identifier).exists()
+
```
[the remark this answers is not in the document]
sarayourfriend: ~Also occurred to me, could this be moved into the `AbstractMedia` class, given it'll probably need to be implemented for every media type? A `mature_class` property on the implementing class would make this nicely generic.~

Update: Never mind, I see that this is probably not easily or ergonomically possible now due to `Mature_` itself requiring the media model class defined on it. The references would be circular.

sarayourfriend (commented):
LGTM so far. I'm checking out the code now to run it locally. I just have one suggestion about the use of `@property` and I/O. I deeply appreciate the unit tests, super helpful :pray:

sarayourfriend:
Hmmm, I'm not able to get my local API to spin up. It hangs waiting for the image index to be ready. I'll debug this a bit more later today so I can finish this review.

AetherUnbound:
> Hmmm, I'm not able to get my local API to spin up. It hangs waiting for the image index to be ready. I'll debug this a bit more later today so I can finish this review.

FWIW a `just recreate` worked for me! This might be an issue with the ingestion server, RE #417

api/catalog/api/models/image.py
```diff
@@ -48,41 +49,43 @@ class Image(ImageFileMixin, AbstractMedia):
     class Meta(AbstractMedia.Meta):
         db_table = "image"
 
+    @property
+    def mature(self) -> bool:
+        return MatureImage.objects.filter(identifier=self.identifier).exists()
```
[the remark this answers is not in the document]
dhruvkb: Based on my understanding, the `@cached_property` decorator wouldn't not do much here, since that it's lifetime is bound to the lifetime of the model instance?

Django has a decorator with the same name, and their [docs](https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.functional.cached_property) state that the cache only persists as long as the instance is in memory.

api/catalog/api/models/audio.py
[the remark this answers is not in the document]
dhruvkb: Based on my understanding, the `@cached_property` decorator wouldn't not do much here, since that it's lifetime is bound to the lifetime of the model instance?

Django has a decorator with the same name, and their [docs](https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.functional.cached_property) state that the cache only persists as long as the instance is in memory.

api/test/unit/models/media_report_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+pytestmark = pytest.mark.django_db
+
+MediaType = Union[Literal["audio"], Literal["image"]]
+
+
+@pytest.fixture
+def media_obj():
+    def _get_media_obj(type: MediaType):
+        klass = Audio if type == "audio" else Image
```
AetherUnbound, line 28: This could potentially be a `dict::get` call, which might make it easier to add future media types:
```suggestion
        klass = {
            "audio": Audio,
            "image": Image,
        }.get(type)
```

api/test/unit/models/media_report_test.py
```diff
@@ -0,0 +1,138 @@
...
+        obj.save()
+        return obj
+
+    return _get_media_obj
+
+
+@pytest.mark.parametrize(
+    "media_type, report_class", [("image", ImageReport), ("audio", AudioReport)]
+)
+@pytest.mark.parametrize("reason", [DMCA, MATURE])
```
AetherUnbound, line 40: Should we add `OTHER` here as a possible reason too?

AetherUnbound, line 40: This is only used in two tests, but we could potentially set this to a variable for use in multiple tests:

```python
reason_params = pytest.mark.parametrize("reason", [DMCA, MATURE, OTHER])

...

@reason_params
def test_foo()
```

AetherUnbound (approved):
This is fantastic! The tests are excellent 👩🏼‍🍳 💯 I was able to test adding reports in the Django Admin and it all behaved as expected!

api/test/unit/models/media_report_test.py
dhruvkb: Implemented in fafcdf8.

dhruvkb: Implemented in fafcdf8.

sarayourfriend:
> FWIW a just recreate worked for me! This might be an issue with the ingestion server, RE https://github.com/WordPress/openverse-api/issues/417

Nah, for me it's because my disk is too full and ES won't let you add or modify indexes if your disk is above 95% full :sweat_smile: Working on moving some things around to fix this issue. I have been successfully ignoring it for a _long_ time.

Should be able to finish reviewing this soon.

dhruvkb:
@sarayourfriend if you do `just down -v` that will delete the ES volume and free up space.

sarayourfriend:
> @sarayourfriend if you do just down -v that will delete the ES volume and free up space.

Yeah, that didn't work. Based on the logs I was seeing ES was measuring the available disk space on the host device, somehow. My main drive is filled to 97% so I'm moving some things off to external drives now to see if it'll fix the situation.

This is entirely a problem of my own creation and nothing to do with this project specifically. Sorry, just gotta bear with me a moment on this one.

sarayourfriend (changes requested):
Noticed a couple things. Some are just nice to haves and could be pushed off into separate issues:

1. The plural form of the audio models could be changed to use `audio` instead of `audios`
2. The order of the items in the Django admin could be rearranged to group audio and image related stuff together. Or maybe just moving the reports to the top of the list so that they're easier to find.
3. I am concerned that we actually delete media objects when we deindex them after review. Should we explore "soft deletion" instead? I noticed this because I tried to create a mature audio report on an audio that I had just deleted.
4. Deleted_ and Mature_ admins allow modifying the identifier. I think this should be a read-only field on these models? Changing the identifier on them and saving them has a strange behaviour: it does not modify the existing record but actually creates a new one. We should prevent this entirely by making these models readonly from the admin. (This is why I am requesting changes on this review).

Otherwise this LGTM :grin:

dhruvkb:
1. Fixed in 1f5aee8. I had not initially done it because Django requires a migration when you change the plural name so the tradeoff didn't seem worth it.
2. Rearranging is possible but it is significantly complicated and will increase the scope of the PR by a lot. Will create a separate issue and PR to address this.
3. Deleted images are not reindexed because during the data refresh we exclude all identifiers for which we have a `DeletedImage` or `DeletedAudio` model instance.
4. Since the identifier UUID is a primary field changing it from the admin makes Django think you're working with a new instance and doesn't affect the old one. Fixed in 41a5339.

sarayourfriend (approved):
Wonderful! Great work, Dhruv :rocket:

PR #958, opened 2022-09-27
Do not unfurl links or media in Slack notifications

### LINKED ISSUE #954: Do not unfurl links and media by default in Slack notifications
## Problem

Recent provider DAG errors have caused notifications containing images to be sent to Slack. For example, a recent data refresh error notification embedded the image that was being processed when the error was encountered. @sarayourfriend pointed out that while these messages have historically been harmless, it's possible that this could happen with NSFW content.

## Description

We have a PR to at least help this situation in the Catalog by [preventing links and media from unfurling](https://github.com/WordPress/openverse-catalog/pull/743) in Slack notifications. We should add the same functionality to the Slack utility in the ingestion server.

We should be able to do this the same way as it is done in the catalog, by using the `unfurl_links` and `unfurl_media` options in the payload [here](https://github.com/WordPress/openverse-api/blob/main/ingestion_server/ingestion_server/slack.py#L48). For reference, [this is where it is done in the Catalog](https://github.com/WordPress/openverse-catalog/blob/main/openverse_catalog/dags/common/slack.py#L97). 

## Additional context

In the Catalog we expose `unfurl_links` and `unfurl_media` as arguments in the Slack utility, so it is possible to set them to `True/False` as needed for an individual message. This _might_ be nice to have, but I don't believe it is currently necessary.

## Implementation

stacimc:
## Fixes

Fixes #954 by @stacimc

## Description

Prevents links and media from unfurling in Slack notifications in the ingestion server.

I've hard-coded this for all Slack notifications rather than exposing this as an option, as is done in the Catalog. For the time  being I don't see any scenarios where we would _want_ to unfurl for the ingestion server, so I do not think it is worth the effort to implement and test. If a situation arises, it should be a quick addition.

## Testing Instructions

This change is pretty straightforward and a nearly identical change has been tested in the Catalog. However, I did test this locally by simulating a recent production error that displayed an image. This error happens during reindexing when duplicate keys are found. To reproduce it I did the following:

1. Updated the [sample_date/sample_images.csv](https://github.com/WordPress/openverse-api/blob/main/sample_data/sample_images.csv) to include a duplicate. To do this, I picked two arbitrary rows and replaced the `url` column on one row with _the url of another row, minus the protocol._ The exact diff I used is available in [this patch.](https://gist.github.com/stacimc/770f347ae3f39abc4d1072068842c252)
2. Set the SLACK_WEBHOOK and LOG_LEVEL environment variables.
3. Run `just recreate` in the `openverse-api` root directory. This will attempt to reindex ES and it should encounter the duplicate key error and notify Slack.

Here's a comparison of an old Slack notification from production, next to the notification from my local test:
| Production | Local test |
| --- | --- |
| Screen Shot 2022-09-27 at 4 20 19 PM | Screen Shot 2022-09-27 at 4 20 32 PM |

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/slack.py
+++ b/ingestion_server/ingestion_server/slack.py
@@ -50,6 +50,8 @@ def _message(text: str, summary: str = None, level: Level = Level.INFO) -> None:
         "text": summary,
         "username": f"Data Refresh Notification | {environment.upper()}",
         "icon_emoji": "arrows_counterclockwise",
+        "unfurl_links": False,
+        "unfurl_media": False,
     }
     try:
         requests.post(webhook, json=data)


sarayourfriend (approved):
Lovely!

zackkrida (approved):

AetherUnbound (approved):
LGTM!

PR #966, opened 2022-10-04
Update of the old CC BASE_URL of media reports

### LINKED ISSUE #965: Replace the old CC `BASE_URL` of media reports
## Description

Currently, the media reports appear in the Django Admin with a URL from Creative Commons, meaning that when you click on the link it will count as a referrer from the old CC Search site, which is not true. The `BASE_URL` should be `https://wordpress.org/openverse/`.

This can be fixed by editing the following line.

https://github.com/WordPress/openverse-api/blob/21405cc4a4ea2f8dca41559d9e5e29763b188e86/api/catalog/api/models/media.py#L145

And potentially and preferably we can centralize this constant in the `settings.py` file for the example responses as well.

https://github.com/WordPress/openverse-api/blob/21405cc4a4ea2f8dca41559d9e5e29763b188e86/api/catalog/api/examples/audio_responses.py#L4

https://github.com/WordPress/openverse-api/blob/21405cc4a4ea2f8dca41559d9e5e29763b188e86/api/catalog/api/examples/image_responses.py#L4

## Reproduction

1. Run `just up` to bring up the API
2. Pick any id from http://localhost:50280/v1/audio/ (or from the image list)
3. Replace `<uuid>` with the id previously selected and create a media report with:
```sh
curl -X POST \
	-H 'Content-Type: application/json' \
	-d '{
		"mediaType":"audio",
		"identifier":"<uuid>",
		"reason":"other",
		"description":"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse blandit id felis sed imperdiet. Integer tristique nisi et metus sollicitudin."
	}'\
	http://localhost:50280/v1/audio/<uuid>/report/
```
4. Go to http://localhost:50280/admin/api/audioreport/ The user & pass for the local API is `deploy`
5. See the wrong URL under the "Audio url" column.

## Screenshots

CleanShot 2022-10-03 at 12 41 28@2x

swapnalshahil:
## Fixes

Fixes #965  by @krysal 

## Description

Currently, the media reports appear in the Django Admin with a URL from Creative Commons, meaning that when you click on the link it will count as a referrer from the old CC Search site, which is not true. The `BASE_URL` should be `https://wordpress.org/openverse/`.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


zackkrida:
Hi @swapnalshahil this looks excellent! Do you think you could also implement this suggestion from @krysal's issue description?

> preferably we can centralize this constant in the `settings.py` file for the example responses as well.

We are working on changing Openverse's domain again in the future, so having this constant centralized in a settings file would make that easier in the future.

api/catalog/settings.py
```diff
@@ -376,3 +376,5 @@
 MAX_ANONYMOUS_PAGE_SIZE = 20
 MAX_AUTHED_PAGE_SIZE = 500
 MAX_PAGINATION_DEPTH = 20
+
+BASE_URL = "https://wordpress.org/openverse/"
```
krysal, line 380: It would be nice to be able to configure this variable from the environment:
```suggestion
BASE_URL = config("BASE_URL", default="https://wordpress.org/openverse/")

```

api/catalog/api/models/media.py
```diff
@@ -142,7 +142,7 @@ class AbstractMediaReport(models.Model):
     deleted_class: type[models.Model] = None
     """the class storing deleted media e.g. ``DeletedImage`` or ``DeletedAudio``"""
 
-    BASE_URL = "https://search.creativecommons.org/"
+    BASE_URL = "https://wordpress.org/openverse/"
```
krysal, line 145: Here you can use the new variable from the settings like this:
```suggestion
    BASE_URL = settings.BASE_URL
```

krysal (changes requested):
Really close! Only needs to use the now centralized variable. I left the exact suggestions in the code comments.

api/catalog/settings.py
swapnalshahil: Thank you @krysal for the suggestions!

krysal (approved):
Thanks @swapnalshahil, and congrats on your first contribution to Openverse! 🎉

dhruvkb (approved):

Issue #921, opened 2022-09-15
`ReadTimeout` raised from general exception catching


krysal:
## Sentry link

https://sentry.io/share/issue/89863f561e9146fba39bb084e6310a68/

## Description

Origin in source code:
https://github.com/WordPress/openverse-api/blob/9f8b15773df511b0900e2a0adf7785382abee450/api/catalog/api/views/media_views.py#L206

## Reproduction

Could not reproduce.

dhruvkb:
This error is now up to 27k in the last 30 days and 1.9k in the last 24 hours.

The current timeout of 10s might be insufficient for some fairly large images to load. Increasing it might reduce the number of error events but bringing it down to zero seems infeasible unless we reliably have low-size alternatives for all large images in the catalog.

AetherUnbound:
It would be really nice to have some data on this to see which providers are contributing the most to these timeouts. Because the URLs stored in the sentry event are our thumbnail links, I'm not sure that's something we'd be able to extract from Sentry. Do you think we could add some logic to store a mapping of TLD -> count in redis? We could increment it each time the `ReadTimeout` is hit, then take a look and see which providers are timing out the most. We might be able to address them upstream in a similar manner to SMK.

dhruvkb:
Unassigning myself because #982 has been merged and we don't have much to do but wait and see the cache for which domains are causing the most timeouts.

PR #982, opened 2022-10-18
Count thumbnail timeouts per domain

dhruvkb:
## Fixes

Related to #921 by @krysal

## Description

This PR caches the number of times `ReadTimeout` occurs for a given domain, which can help us identify the providers that are the slowest for thumbnail generation. Then we can find ways to speed them up or work around their limitations.

## Testing Instructions

Unit tests are included in the PR. They should be sufficient as there is no external change.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/views/media_views.py
```diff
@@ -211,6 +212,19 @@ def _thumbnail_proxy_comm(
             )
 
             return upstream_response, res_status, content_type
+        except requests.ReadTimeout as exc:
+            # Count the incident so that we can identify providers with most timeouts.
+            domain = urlparse(params["url"]).netloc
+            key = f"{settings.THUMBNAIL_TIMEOUT_PREFIX}__{domain}"
```
sarayourfriend, line 218: We currently use `:` to separate prefixes rather than `__`. What is the reason for the difference here?

https://github.com/WordPress/openverse-api/blob/9f8b15773df511b0900e2a0adf7785382abee450/api/catalog/api/utils/validate_images.py#L16

api/test/unit/views/media_views_test.py
```diff
@@ -241,3 +243,19 @@ def test_thumb_full_size(api_client, media_type, media_factory, requests):
         assert (
             entry in requests.sent_requests[1].request.url
         ), f"{entry} not found in prepared request url: {requests.sent_requests[0].request.url}"
+
+
+@pytest.mark.parametrize("count", [1, 3])
+def test_thumb_timeout(count: int):
+    cache.clear()  # to prevent effect of other runs
```
sarayourfriend, line 250: This will clear the cache for every other test... which in theory should be fine but also could not be (in particular with the "integration" tests). Can we use `RedisMock` here like we do in the other unit tests, rather than relying on the real elasticsearch? That way we get a fresh redis to play with each test run. See this test fixture in `validate_images_test` for an example:

https://github.com/WordPress/openverse-api/blob/f0c44d571249a2631eef5a13ea0d56702210daf3/api/test/unit/utils/validate_images_test.py#L12-L22

sarayourfriend (commented):
LGTM! Just a couple questions about separator consistency and the usage of the live redis rather than RedisMock in the tests.

api/catalog/api/views/media_views.py
dhruvkb: I didn't remember we used `:` prefix. Thanks for pointing that out. Fixed in 3b9eca4.

api/test/unit/views/media_views_test.py
dhruvkb: Added a fake Redis fixture in 3b9eca4.

api/test/unit/views/media_views_test.py
```diff
@@ -241,3 +243,30 @@ def test_thumb_full_size(api_client, media_type, media_factory, requests):
...
+def test_thumb_timeout(redis, count: int):
+    with patch("requests.get", side_effect=ReadTimeout()):
+        for idx in range(count):
+            try:
+                MediaViewSet._thumbnail_proxy_comm(
+                    "test", params={"url": f"https://example.com/image/{idx}"}
+                )
+            except UpstreamThumbnailException:
+                # Intentionally raised exception, do nothing.
+                pass
```
sarayourfriend, line 268: Should the test use `pytest.raises` I wonder? If we expect the exception to be raised then it'd be nice if the unit test encapsulated that aspect.

api/env.template
```diff
@@ -24,6 +24,8 @@ ALLOWED_HOSTS=172.17.0.1,host.docker.internal
 #THUMBNAIL_JPG_QUALITY=80
 #THUMBNAIL_PNG_COMPRESSION=6
 
+#THUMBNAIL_TIMEOUT_PREFIX=thumbnail_timeout
```
sarayourfriend, line 27: This value leaves off the separator, but the default in the `settings.py` has it. Should we be consistent here? Do we need to include the separator in the prefix or is that something we could inline in the template string used to generate the key: https://github.com/WordPress/openverse-api/pull/982/files#diff-be5fb4e5665130b04ebed93a52ff1730d80ffb734606e9f5c41dfb066c39bbcbR218

I don't have a strong opinion here in either direction, just wanting to be consistent with the settings.

sarayourfriend (approved):
LGTM. Just a couple comments, nothing critical.

It might be good to document some assumptions about how this information will be used. For example, some providers have multiple domains they provide thumbnails from. The dead link tally script had some hacky code to try to resolve this, for example by stripping the left-most subdomains when any subdomains existed. I don't know if that's necessary here. What if, for example, a provider has several thumbnail servers and a particular one of those times out for us regularly. Maybe the provider splits things geographically based on where the uploader's account is held, or something like that. It's nice in this case that we would have the entire domain in those cases.

Is there anything else we'd want to keep in mind when using the information generated by this enhancement?

api/env.template
dhruvkb: Resolved in 63dda22.

api/test/unit/views/media_views_test.py
dhruvkb: Resolved in 6b24482.

api/catalog/api/views/media_views.py
```diff
@@ -211,6 +212,19 @@ def _thumbnail_proxy_comm(
...
+        except requests.ReadTimeout as exc:
+            # Count the incident so that we can identify providers with most timeouts.
+            domain = urlparse(params["url"]).netloc
+            key = f"{settings.THUMBNAIL_TIMEOUT_PREFIX}{domain}"
+            try:
+                cache.incr(key)
+            except ValueError:  # Key does not exist.
+                cache.set(key, 1)
+
+            capture_exception(exc)
```
AetherUnbound, line 224: Did we want to wrap this whole thing in a `try`/`except` to ensure that we handle any errors that might occur as part of `urlparse` or any other of these steps?

api/test/unit/views/media_views_test.py
```diff
@@ -241,3 +243,27 @@ def test_thumb_full_size(api_client, media_type, media_factory, requests):
...
+def test_thumb_timeout(redis, count: int):
+    with patch("requests.get", side_effect=ReadTimeout()):
+        for idx in range(count):
+            with pytest.raises(UpstreamThumbnailException):
+                MediaViewSet._thumbnail_proxy_comm(
+                    "test", params={"url": f"https://example.com/image/{idx}"}
+                )
+
+    # Conversion to ``int`` required because of bug:
+    # https://github.com/cunla/fakeredis-py/issues/58
```
AetherUnbound, line 268: Woah! It looks like this is closed upstream now, do we still need to do that conversion or does it need to be brought to a different repo?

AetherUnbound (approved):
Looks great, thanks for doing this! Sara's points about the domains are good, I think _not_ trying to do that munging will yield us more interesting results for the described reasons. I have some questions here but nothing to block a merge 🙂

api/test/unit/views/media_views_test.py
dhruvkb: This was closed as "won't fix" (intended behaviour) so we need to do this conversion.

api/catalog/api/views/media_views.py
dhruvkb: very unlikely for URL parse to raise an exception. If we're making a request that times out, the URL would have to be valid and if so, `urlparse` should be able to get a `netloc` out of it. I tried it with some weird URLs I could come up with and `urlparse` can handle anything 😂.

dhruvkb:
Merging this so that we can initiate a proper fix for #921.

api/catalog/api/views/media_views.py
AetherUnbound: Fair point! Just being overly precautious 😄

PR #990, opened 2022-10-28
Add Nginx target to API Dockerfile

### LINKED ISSUE #821: Add nginx to the API multi-stage Dockerfile
## Problem

For the ECS deployment of the frontend, we need to be able to deploy a Django and an Nginx image. To reduce the time it takes for the Django image to go from running to ready and to reduce the complexity of the deployment, we want to front load as much work as possible into the build.

Currently, the API deployment is complicated by the following:

1. `collectstatic` run in Django container with output shared to Nginx via compose volumes
2. Nginx configuration configured entirely during the init script of our EC2 instances
3. Version information passed as environment variables rather than baked into the images themselves at build time (meaning you could run image 1 and pass it version 2 and it would report the wrong version to you :scream:)

## Description

All three of those complexities can be moved into a multi-stage Docker build. Something roughly like this:

```dockerfile
FROM python:10-slim as builder

# ... the existing builder stuff

RUN collectstatic

FROM python:10-slim as django

# ... the existing stuff for this

FROM nginx as nginx

# VERSION is the git ref the image was built from
ARG VERSION
RUN echo "{\"version\":\"$VERSION\"}"
COPY --from=builder collectstatic-output
COPY nginx.conf .
```

Obviously this is missing everything that exists in the current Dockerfile which should be kept.

## Additional context

Part of ongoing work to get the API deployed in ECS.

## Implementation

sarayourfriend:
## Fixes

Fixes #821 by @sarayourfriend 

## Description

This PR adds a new `nginx` target to the Dockerfile that allows us to build an `nginx` image with the API's static files built into it. As described in the linked issue, this allows us to get around the need to distribute these files inside the Django container and without needing to share mounts across images in the ECS service.

The reason we cannot just store these in the Django image and share across volumes anyway is that `collectstatic` runs the entire Django app's settings, and therefore requires connections to Elasticsearch to work. I tried working around this so that it didn't require running the entire Docker compose stack, but there's a lot of assumptions built into how our Django app settings are configured that just assume that ES, Redis, and a database will be available. That's not a bad thing or something we need to change, but it would require a lot more work than just splitting the `nginx` target out to a separate build step in the CI. It was much easier to solve this way than to pull apart the assumptions built into the Django app (which would be riskier to change anyway).

Thanks @AetherUnbound for helping debug the networking of the running image.

## Testing Instructions

Build and run the container as described above. Make test requests to ensure it works, against the following running container:

```shell
cd api
just collectstatic
docker build --target nginx . -t openverse-api-nginx:latest
docker run --rm -p 8080:8080 -e DJANGO_UPSTREAM_URL='api.openverse.engineering' -it openverse-api-nginx:latest
```

You won't get a successful response back because the image isn't able to make a valid request to `api.openverse.engineering` that Cloudflare will accept.

I can't think of a way to get this to successfully run inside of docker-compose as the image requires the compose stack to be running beforehand to even build due to needing `collectstatic` to populate the local filesystem's `./api/static` directory.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


AetherUnbound:
Okay admittedly I'm not sure I understand all the pieces in play here, but this does work locally for me! The following command ran:

```bash
$ docker run --rm -p 8080:8080 -e DJANGO_UPSTREAM_URL='api.openverse.engineering' -it openverse-api-nginx:latest
```

I ran into issues when I tried to hit http://localhost:8080, but the response I got back was `400 Bad Request (cloudflare)`. I take it to mean it's forwarding to api.openverse.engineering correctly! 

Visiting a URL like http://localhost:8080/static/admin/img/search.svg returned the static file as expected too.

I tried some variations of `DJANGO_UPSTREAM_URL`, namely `localhost:50820` and `172.17.0.1:50820` but I wasn't able to get either to work.

AetherUnbound:
Got it working, accidentally swapped port numbers! It should be `docker run --rm -p 8080:8080 -e DJANGO_UPSTREAM_URL='172.17.0.1:50280' -it openverse-api-nginx:latest` (50280 not 50820). Now I'm getting this error, but at least it's coming from Django!
```
<h1>DisallowedHost
       at /</h1>
  <pre class="exception_value">Invalid HTTP_HOST header: '\\localhost:8080'. The domain name provided is not valid according to RFC 1034/1035.</pre>
  

Request Method: | GET
-- | --
http://\localhost:8080/
4.1.2
DisallowedHost
Invalid HTTP_HOST header: '\\localhost:8080'. The domain name provided is not valid according to RFC 1034/1035.
/venv/lib/python3.10/site-packages/django/http/request.py, line 148, in get_host
django.views.generic.base.RedirectView
/venv/bin/python
3.10.8
['/api',  '/usr/local/lib/python310.zip',  '/usr/local/lib/python3.10',  '/usr/local/lib/python3.10/lib-dynload',  '/venv/lib/python3.10/site-packages']
Fri, 28 Oct 2022 20:43:30 +0000
```

sarayourfriend:
Thanks for taking a look, Madison :slightly_smiling_face: I wonder what I was doing wrong on my end that it wasn't working :thinking:

Glad it's working in any case. When I can return to this next week it'll be fun to undraft it soon!

.github/workflows/ci_cd.yml
```diff
@@ -74,6 +74,56 @@ jobs:
...
+          path: /tmp
+
+      - name: Load API and ingestion server images
+        run: |
+          docker load --input /tmp/api/api.tar
+          docker load --input /tmp/ingestion_server/ingestion_server.tar
+
+      - name: Start API, ingest and index test data
+        # Is this necessary for collectstatic?
+        run: just up
```
krysal, line 100: I... have the same question, and will the production instance include test data? 🤔

krysal (commented):
This is very interesting, I didn't think about the possibility of a scoped stage build, good to know that is an option! 💡 I'm not grabbing yet why is not possible to separate the API and Nginx images since it seems we can share files between actions and state the order of the image builds 🤔

Also, currently, this command doesn't work for me:

```docker build --target nginx . -t openverse-api-nginx:latest```

This is the output I get:

```
[+] Building 0.0s (1/2)                                                                                                                               
 => [internal] load build definition from Dockerfile                                                                                             0.0s
 => => transferring dockerfile: 2B                                                                                                               0.0s
failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount3456991816/Dockerfile: no such file or directory
```

I hope is not something with Docker at macOS only. I tried with `docker-compose` as well and it didn't work: `unknown flag: --target`

File: .github/workflows/ci_cd.yml [hunk 2]
```diff
@@ -74,6 +74,56 @@ jobs:
...
+
+      - name: collectstatic
+        run: just collectstatic
+
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
krysal, line 106: Can we include updating this action on this file?
```suggestion
        uses: docker/setup-buildx-action@v2
```

api/Dockerfile
```diff
@@ -34,6 +34,22 @@ COPY Pipfile Pipfile.lock ./
...
+
+# This target assumes that the build host has run `manage.py collectstatic`
+# `just collectstatic` is provided as a convenient alias for running it the expected way
+
+FROM nginx:1.23.2-alpine as nginx
+
+WORKDIR /app
+
+COPY nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
+COPY /static /app/static
```
krysal, line 49: The `static` folder needs to exist for the first time as well, or the build for the API fails. We can add a `.gitkeep` file inside to track it.

krysal (changes requested):
I tried again changing to the `api` directory before the docker commands and it works! With a little addition.

api/Dockerfile
sarayourfriend: I should have added in the testing instructions that you need to run `just collectstatic`. The build is meaningless without the static files, I think...

.github/workflows/ci_cd.yml
sarayourfriend: No, I'll remove the `test data` comment, it is irrelevant for this.

sarayourfriend:
> I'm not grabbing yet why is not possible to separate the API and Nginx images since it seems we can share files between actions and state the order of the image builds

Can you clarify what you meant here? I'm not understanding what this is in reference to as the goal of this PR is to separate the API and Nginx images. Is it something with the GitHub actions?

krysal (approved):
Thank you for adding the steps to test it!

What I mean is the way in which the Nginx image is built under the api's `Dockerfile` is new to me (really cool though as I said!) Before this PR I was thinking the separation of Nginx and the API server would require another service entry in the **Docker Compose file**, as is currently being setup in the EC2 instance. The building from the `Dockerfile` doesn't tweak the original Nginx image, it only puts configuration and static files in and seems kind of odd being there, it's not affecting the API image built. I understand there is complexity with the Django's static files though, and this solution looks easier (now that it's already done!)

I imagine the alternative would be to make Terraform produce the static files in the API and then share them between the API and Nginx images (?) Anyway, that can be researched later.

Thanks for this amazing work!I just let some suggestion to use the last versions of GitHub actions and a question.

PD: Can we remove the "proxy" service from the `docker-compose.yml` after this? Not asking to include it in this PR but we're not using it.

api/nginx.conf.template
```diff
@@ -0,0 +1,85 @@
+error_log /var/log/nginx/error.log;
```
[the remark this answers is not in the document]
sarayourfriend: Those are all redundant with the default Nginx configuration. If you add them here Nginx will error on initialisation due to multiple declarations.

sarayourfriend:
@krysal I've updated the actions :+1: 

Also, I realised that the obvious test for this I didn't mention or think about until yesterday, is to try to retrieve a static file. If you hit http://localhost:8080/static/admin/css/base.css locally after running the image with port 8080 exposed, you should get a CSS file. This should match this file from the production API: https://api.openverse.engineering/static/admin/css/base.css

sarayourfriend:
I'm reverting the updates to the action versions to see if it fixes the ingestion server test failures. If not I'll reapply them and then ping for help on the test failures.

AetherUnbound (approved):
I was able to get this working and tested out the new instructions you provided with success (@zackkrida I also had a local port conflict with a proxy I was running, if that's what you were also running into).

This looks great!

PR #1012, opened 2022-11-21
Make quoted queries behave as described in the API documentation (return exact matches only)

### LINKED ISSUE #933: Exact match does not function as described in the documentation
## Description

It would appear that using a quoted search (e.g. `"this is my search"` vs `this is my search`) does not behave as described in the documentation:

> ```
> # Example 2: Search for audio that is an exact match of Giacomo Puccini
> curl \
>   -H "Authorization: Bearer DLBYIcfnKfolaXKcmMC8RIDCavc2hW" \
>   "https://api.openverse.engineering/v1/audio/?q="Giacomo Puccini""
> ```

Firstly, this documentation doesn't escape the quotes correctly, so `Puccini` is read in as a bash command.

Additionally, the quoted search appears to do nothing to the results. From the search controller code, it looks like we're boosting exact matches in all cases, but ignoring quotes otherwise: 

https://href.li/?https://github.com/WordPress/openverse-api/blob/a44f8d19cc3b2612469bbf20a3aac54c44086712/api/catalog/api/controllers/search_controller.py#L340-L356

It seems that we will need to change this logic so that if a term is quoted, _only_ an exact match search is made. Exact matches could still be boosted in regular searches, but non-exact matches should not be returned in quoted searches.

## Reproduction

1. `just recreate`
2. Visit http://localhost:50280/v1/images/?q=food
3. Visit http://localhost:50280/v1/images/?q=%22food%22 and observe that the list remains unchanged, even though some posts (those on page 6) are matched on the tag food

## Resolution

sarayourfriend:
## Fixes

Fixes #933 by @AetherUnbound 

## Description

Conditionally applies the `quote_field_suffix` query feature as suggested in the Elasticsearch documentation that @AetherUnbound shared in the linked issue: https://www.elastic.co/guide/en/elasticsearch/reference/current/mixing-exact-search-with-stemming.html

We can continue to "boost" exact matches against the title as we originally were, but I've added the `.exact` modifier there as well to ensure that the boost only applies to actual exact matches.

The latter is hard to test locally, I haven't found any suitable queries for it in the test data and I do not know how to efficiently add usable test data.

The former, however, is pretty straightforward to test locally. Find a query that returns some results that you could narrow using an exact match. I found one for images and one for audio that are used in the integration tests:

Audio: `http://localhost:50280/v1/audio/?q=water%20running` vs `http://localhost:50280/v1/audio/?q=%22water%20running%22`

Images: `http://localhost:50280/v1/images/?q=bird%20perched` vs `http://localhost:50280/v1/images/?q=%22bird%20perched%22`

You can try these locally to see how the results differ and that the way they differ makes sense. The integration test only confirms that they do indeed differ. I'm not  sure how to reasonably test that the exact match is "working" in a way that doesn't test implementation details. We could intercept the query to ES and see that `quote_field_suffix` is applied. Alternatively we could test that the exact match is present in the match fields of the response... but both feel like either testing Elasticsearch (which we do not have to do) or testing an abstract implementation detail that doesn't approximate the actual expectation. Any advice here or creative solutions would be appreciated.

Finally, I also fixed the incorrectly quoted example queries for images and audio exact match queries.

## Testing Instructions

Check out the testing instructions in the previous section. If you can find any other queries that demonstrate this working with the test data, please share them here.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -357,7 +357,7 @@ def search(
         quotes_stripped = query.replace('"', "")
         exact_match_boost = Q(
             "simple_query_string",
-            fields=["title.exact"],
+            fields=["title"],
             query=f"{quotes_stripped}",
             boost=10000,
         )
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -340,18 +340,25 @@ def search(
     search_fields = ["tags.name", "title", "description"]
     if "q" in search_params.data:
         query = _quote_escape(search_params.data["q"])
+        base_query_kwargs = {
+            "query": query,
+            "fields": search_fields,
+            "default_operator": "AND",
+        }
+
+        if '"' in query:
+            base_query_kwargs["quote_field_suffix"] = ".exact"
+
         s = s.query(
             "simple_query_string",
-            query=query,
-            fields=search_fields,
-            default_operator="AND",
+            **base_query_kwargs,
         )
-        # Boost exact matches
+        # Boost exact matches on the title
         quotes_stripped = query.replace('"', "")
         exact_match_boost = Q(
             "simple_query_string",
-            fields=["title"],
-            query=f'"{quotes_stripped}"',
+            fields=["title.exact"],
+            query=f"{quotes_stripped}",
             boost=10000,
         )
         s = search_client.query(Q("bool", must=s.query, should=exact_match_boost))
--- a/api/test/image_integration_test.py
+++ b/api/test/image_integration_test.py
@@ -15,6 +15,7 @@ from test.media_integration import (
     search_all_excluded,
     search_consistency,
     search_quotes,
+    search_quotes_exact,
     search_source_and_excluded,
     search_special_chars,
     stats,
@@ -51,6 +52,11 @@ def test_search_source_and_excluded():
 
 def test_search_quotes():
     search_quotes("images", "dog")
+
+
+def test_search_quotes_exact():
+    # ``bird perched`` returns different results when quoted vs unquoted
+    search_quotes_exact("images", "bird perched")
 
 
 def test_search_with_special_characters():
--- a/api/catalog/api/examples/audio_requests.py
+++ b/api/catalog/api/examples/audio_requests.py
@@ -10,7 +10,7 @@ identifier = "8624ba61-57f1-4f98-8a85-ece206c319cf"
 syntax_examples = {
     "using single query parameter": "test",
     "using multiple query parameters": "test&license=pdm,by&categories=illustration&page_size=1&page=1",  # noqa: E501
-    "that is an exact match of Giacomo Puccini": '"Giacomo Puccini"',
+    "that is an exact match of Giacomo Puccini": r"%22Giacomo%20Puccini%22",
     "related to both dog and cat": "dog+cat",
     "related to dog or cat, but not necessarily both": "dog|cat",
     "related to dog but won't include results related to 'pug'": "dog -pug",
--- a/api/catalog/api/examples/image_requests.py
+++ b/api/catalog/api/examples/image_requests.py
@@ -10,7 +10,7 @@ identifier = "4bc43a04-ef46-4544-a0c1-63c63f56e276"
 syntax_examples = {
     "using single query parameter": "test",
     "using multiple query parameters": "test&license=pdm,by&categories=illustration&page_size=1&page=1",  # noqa: E501
-    "that are an exact match of Claude Monet": '"Claude Monet"',
+    "that are an exact match of Claude Monet": "%22Claude%20Monet%22",
     "related to both dog and cat": "dog+cat",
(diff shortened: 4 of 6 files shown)


api/catalog/api/controllers/search_controller.py
```diff
@@ -340,18 +340,25 @@ def search(
...
+            **base_query_kwargs,
         )
-        # Boost exact matches
+        # Boost exact matches on the title
         quotes_stripped = query.replace('"', "")
         exact_match_boost = Q(
             "simple_query_string",
-            fields=["title"],
-            query=f'"{quotes_stripped}"',
+            fields=["title.exact"],
```
stacimc, line 360: I think this might actually be breaking the exact match boost, oddly. When searching without quotes (http://localhost:50280/v1/images/?q=bird%20perched), I noticed that some results that did **not** have exact title match were included in the first page of results. (For example, there's a `Black Bird Photo` and `Bird Nature Photo` mixed in.)

I tried removing the `.exact` added here and then the boost appeared to work again -- only photos titled `Bird Perched Photo` appeared in the first page of results.

stacimc (dismissed):
Thank you for the testing instructions and finding some good sample queries!

Tests went well! For example, when testing with the `bird perched` query, I received **45** results when not using quotes and **26** when quoted. The exact match when searching with quotes worked perfectly. I think the _boost_ may be broken however -- see comment below.

As for the testing, I agree with everything you said. Although it'd be nice to test further, the best I can come up with is testing that the appropriate parameters were passed, which feels like testing Elasticsearch itself.

api/catalog/api/controllers/search_controller.py
sarayourfriend: Interesting! Thank you for looking into and testing this. I'll make the change (and do a little reading to understand why that would be the case :thinking:)

stacimc: It is very weird! I tried removing the `.exact` totally on a whim to compare, definitely expected your syntax here to be the correct one 🤔

sarayourfriend: Staci, I did a bit of reading in the ES documentation because I was confused why it didn't work the way we expected. I think it comes down to us using the `bool` query in a particular way that I have to admit, I do not fully understand. I tried to re-write the boost (just out of curiosity, not for this PR) to use the field-boosting described in the simple query string DSL documentation, but I could not get the results to budge _at all_. I am pretty curious to read more about Elasticsearch to understand better how these things are meant to work and how score boosting is best approached. It'd be nice, in any case, to document the current approach, why it works and why (if it indeed is) it is the best and correct approach for our use case.

Anyway, I removed the `.exact` on this one and things are back in working order. Thanks again for looking into this!

stacimc: Thanks for looking into it! Super strange.

I re-tested and everything looks good to me except for the test failure. Not sure what's going on there, as. when I test locally it's definitely the case that we get more unquoted results than quoted ones.

AetherUnbound (approved):
This looks great! Your suggested queries do return different results as expected.

I took a look into why the tests are failing, through some debug printing it seems that both the quoted & unquoted searches are returning 20 results:

```
test/image_integration_test.py::test_search_quotes_exact len(quoted_results)=20
len(unquoted_results)=20
FAILED
```

I suspect this is because both queries actually return _more_ than 20 but are being capped at the default page size (since both 45 and 26 are above that first page length). When I change the test to use the search term "fireworks celebration" instead (which returns 7 and 3 results unquoted and quoted respectively) the tests pass fine!

api/test/image_integration_test.py
```diff
@@ -53,6 +54,11 @@ def test_search_quotes():
     search_quotes("images", "dog")
 
 
+def test_search_quotes_exact():
+    # ``bird perched`` returns different results when quoted vs unquoted
+    search_quotes_exact("images", "bird perched")
```
AetherUnbound, line 59:
```suggestion
    search_quotes_exact("images", "fireworks celebration")
```

This should make the tests pass!

sarayourfriend:
> I took a look into why the tests are failing, through some debug printing it seems that both the quoted & unquoted searches are returning 20 results:

I hadn't had a chance to debug this yet today but this strikes me as an "oh gosh, of course"! Thank you for pointing it out. We could change the query or, more robustly, we could compare the "results" count that is returned for each search. I will probably do that in case we add more results that would be returned by any of the test queries in the future.

AetherUnbound:
Ooo, result count is a MUCH better thing to compare, good call!

api/test/image_integration_test.py
sarayourfriend: Just to document why this wasn't committed, taking a slightly different approach described here: https://github.com/WordPress/openverse-api/pull/1012#issuecomment-1330019285

stacimc (approved):
🎉

PR #1013, opened 2022-11-22
Add db connection commands and non-root user for `web` container

krysal:
## Fixes

Related to #952 by @zackkrida in the sense that, if I remember correctly, it was driven by a difficulty presented to @stacimc to connect to the API DB.

## Description

This PR:
- installs `pgcli` as dev dependency
- adds a `just dbshell` _and_ `just pgcli` commands, taking inspiration from [the one in the catalog repository](https://github.com/WordPress/openverse-catalog/blob/main/justfile#L115-L117) and using the provided[ django command](https://docs.djangoproject.com/en/4.1/ref/django-admin/#dbshell) as well
- adds a non-root `opener` user to the `web` container, removing the tricky pass of the host user, it wasn't working well in macOS, so I believe creating a proper user is the best route we can follow
- adds `GITHUB_TOKEN` to steps so setup just, it's necessary to prevent rate limiting errors (as [said in docs](https://github.com/extractions/setup-just#usage))
- bumps action dependencies to prevent warnings on deprecated commands

## Testing Instructions

1. `just up` & `just db-shell` 
2. Enter the development password: `deploy`
3. You should be able to make queries in the pgcli console

If you try these commands at the commit `2c1c7246dbf019c2cf1747211273923fa2e50826`, without the user created internally, you will notice it fails (at least on macOS).

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -79,7 +79,7 @@ COPY --from=awf /usr/local/bin/audiowaveform /usr/local/bin
 #   - libpq-dev: required by `psycopg2`
 # - Create directory for dumping API logs
 RUN apt-get update \
-      && apt-get install -y curl libpq-dev libexempi8 \
+      && apt-get install -y curl libpq-dev libexempi8 postgresql-client \
       && rm -rf /var/lib/apt/lists/* \
     && mkdir -p /var/log/openverse_api/openverse_api.log
 
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -45,12 +45,8 @@ FROM nginx:1.23.2-alpine as nginx
 
 WORKDIR /app
 
-# Create a non-root user
-RUN useradd --create-home opener
-USER opener
-
-COPY --chown=opener nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
-COPY --chown=opener /static /app/static
+COPY nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
+COPY /static /app/static
 
 # Only environment variables with this prefix will be available in the template
 ENV NGINX_ENVSUBST_FILTER="DJANGO_NGINX_"
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -45,8 +45,12 @@ FROM nginx:1.23.2-alpine as nginx
 
 WORKDIR /app
 
-COPY nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
-COPY /static /app/static
+# Create a non-root user
+RUN useradd --create-home opener
+USER opener
+
+COPY --chown=opener nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
+COPY --chown=opener /static /app/static
 
 # Only environment variables with this prefix will be available in the template
 ENV NGINX_ENVSUBST_FILTER="DJANGO_NGINX_"
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -83,8 +83,12 @@ RUN apt-get update \
       && rm -rf /var/lib/apt/lists/* \
     && mkdir -p /var/log/openverse_api/openverse_api.log
 
+# Create a non-root user
+RUN useradd --create-home opener
+USER opener
+
 # Copy code into the final image
-COPY . /api/
+COPY --chown=opener . /api/
 
 # Exposes
 # - 8000: Dev server for API Django app
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,7 +87,6 @@ services:
       MEDIA_ROOT: ${MEDIA_ROOT:-}
     stdin_open: true
     tty: true
-    user: ${DOCKER_USER_ID}:${DOCKER_GROUP_ID}
 
   ingestion_server:
     build: ./ingestion_server/
--- a/justfile
+++ b/justfile
@@ -262,9 +262,13 @@ nginx upstream_url='api.openverse.engineering': collectstatic
       -e DJANGO_NGINX_GIT_REVISION="$(git rev-parse HEAD)" \
       openverse-api-nginx:latest
 
-# Launch a pgcli shell on the web container
-db-shell db_host="db":
-    docker-compose {{ DOCKER_FILE }} exec web pgcli -h {{ db_host }} openledger deploy
+# Launch a psql shell in the web container
+@dbshell:
+    just exec web python manage.py dbshell
+
+# Launch a pgcli shell in the web container (require typing credentials)
+@pgcli db_host="db":
+    just exec web pgcli -h {{ db_host }} openledger deploy
 
 
 ##########
--- a/justfile
+++ b/justfile
@@ -262,8 +262,8 @@ nginx upstream_url='api.openverse.engineering': collectstatic
       openverse-api-nginx:latest
 
 # Launch a pgcli shell on the web container
-db-shell: _api-up
-    docker-compose {{ DOCKER_FILE }} exec web pgcli openledger deploy
+db-shell db_host="db":
+    docker-compose {{ DOCKER_FILE }} exec web pgcli -h {{ db_host }} openledger deploy
 
 
 ##########
--- a/justfile
+++ b/justfile
@@ -247,6 +247,9 @@ ipython:
     # The resulting output will be at `api/static` and is git ignored for convenience.
     @STATIC_ROOT="./static" just dj collectstatic --noinput
 
+@collectstatic-as-admin: _api-up
+    just exec -u root web python manage.py collectstatic --noinput
+
 # Run the nginx image locally
 nginx upstream_url='api.openverse.engineering': collectstatic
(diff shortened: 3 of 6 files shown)


justfile
```diff
@@ -237,6 +237,10 @@ ipython:
     # The resulting output will be at `api/static` and is git ignored for convenience.
     @STATIC_ROOT="./static" just dj collectstatic --noinput
 
+# Launch a pgcli shell on the web container
+db-shell db_host="db":
+    docker-compose {{ DOCKER_FILE }} exec web pgcli -h {{ db_host }} openledger deploy
```
dhruvkb, line 242: We should use the [`dbshell` management command](https://docs.djangoproject.com/en/4.1/ref/django-admin/#dbshell). It will automatically set the host, username and password for us (from the `DATABASES` setting) and drop us directly in the SQL prompt.

```suggestion
    just dj dbshell
```

It uses the `psql` command-line client so not 100% sure if it'll work inside the Docker container without adding a step to install it first.

docker-compose.yml
```diff
@@ -87,7 +87,6 @@ services:
       MEDIA_ROOT: ${MEDIA_ROOT:-}
     stdin_open: true
     tty: true
-    user: ${DOCKER_USER_ID}:${DOCKER_GROUP_ID}
```
dhruvkb, line 90: Among other services, having to set `user` on just this one was weird so I appreciate that it's gone.

justfile
krysal: It also needs an extra package installed on the `web` container, `postgresql-client`, or it throws an error:

```sh
CommandError: You appear not to have the 'psql' program installed or on your path.
error: Recipe `exec` failed on line 72 with exit code 1
error: Recipe `dj` failed on line 223 with exit code 1
```

I guess we can add it as well to avoid having to type the password, could be the `just pqsql` command, but [`pgcli`](https://github.com/dbcli/pgcli) has some nice extra features in comparison to the basic `psql`, like autocompletion and syntax highlighting 📝

I wonder if the rest of folks find valuable the extra size for this feature? @obulat @AetherUnbound

api/Dockerfile
```diff
@@ -81,8 +81,12 @@ RUN apt-get update \
       && rm -rf /var/lib/apt/lists/* \
     && mkdir -p /var/log/openverse_api/openverse_api.log
 
+# Create a non-root user
+RUN useradd --create-home opener
+USER opener
+
 # Copy code into the final image
-COPY . /api/
+COPY --chown=opener . /api/
```
AetherUnbound, line 89: I will mention (in dealing with this for a local project) that this is a *non-trivial change* and may have adverse and unexpected effects on any volumes we mount. I suspect for the API this won't be as much of an issue, but it's been an issue for some personal work where I've been trying to use the official Airflow image (which uses the `airflow` user rather than `root`).

justfile
AetherUnbound: I'm definitely a big fan of `pgcli`, and I was about to mention that we'd probably need to use `psql` (with the code as it stands now) since we're not [installing it in our postgres image like we do for the catalog](https://github.com/WordPress/openverse-catalog/blob/main/docker/local_postgres/Dockerfile). We're [already installing some libraries in the docker image](https://github.com/WordPress/openverse-api/blob/431b2f9f71423f96cc823a6d9141017e065d3329/api/Dockerfile#L80) so I think adding the postgresql-client library would be alright 🙂 We could always create dev and prod images from the Docker build where the prod one does not have any dev libraries.

We don't necessarily have to install pgcli on the Postgres image like we do for the catalog but it would need to be installed somewhere if we wanted to use it. That to say, I'm +1 for setting it up to work with the API one way or another! It seems like we could set up the `dbshell` Django command to use `pgcli`, but it'd require an additional library: https://www.pgcli.com/django-pgcli.html

api/Dockerfile
krysal: That is unfortunate 😞 Will put it as a draft for now... and get back to it later.

justfile
dhruvkb: `django-pgcli` hasn't been updated in 2 years (not that it needs to, it's just 19 lines) so there is a possibility it might not work with the newest Django!

I sometimes use the `psql` client installed on the host to connect to the Docker `db` service and it works fine.

api/Dockerfile
dhruvkb: @AetherUnbound the use of `COPY` means that the files are being copied into the container and those copies are being `chown`-ed. I don't think it'll affect the permissions on the host but I'm not that well-versed to say for sure without trying it out first.

AetherUnbound: My warnings mostly affect any bound volumes we have. It looks like for the API, we just have this one: https://github.com/WordPress/openverse-api/blob/3479b8c1c377a0be02480a40239c97876eafa407/docker-compose.yml#L75

I _think_ that might be fine, but it's also possible that the `opener` user won't have read access to those files. If it does and the stack works fine then I think my warnings may have been a bit too aggressive, apologies 😄 I know for the catalog we'll have to take a closer look because we do have bind mounts in production (whereas I don't think we do for the API).

krysal: Tests passed using the `root` user for output-generating commands, woohoo!!

AetherUnbound: We discussed this approach a bit [on the Make WP Slack](https://wordpress.slack.com/archives/C02012JB00N/p1670944461453309), glad it worked out! 😄

justfile
krysal: Added both commands! 🙂

krysal:
I'm glad this ended up working out! Thanks a lot, @AetherUnbound, [for your suggestion][gh-link] [(in slack)][slack-link]! It was crucial for this. And also, @dhruvkb for suggesting adding `dbshell` by Django that doesn't require writing the password 😃

[gh-link]: https://github.com/WordPress/openverse-catalog/pull/874/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1R126

[slack-link]: https://wordpress.slack.com/archives/C02012JB00N/p1670944856510849?thread_ts=1670944461.453309&cid=C02012JB00N

justfile
```diff
@@ -237,6 +235,9 @@ ipython:
     # The resulting output will be at `api/static` and is git ignored for convenience.
     @STATIC_ROOT="./static" just dj collectstatic --noinput
 
+@collectstatic-as-admin: _api-up
+    just exec -u root web python manage.py collectstatic --noinput
+
```
sarayourfriend, line 240: What is the use-case for this recipe? Is it left-over from debugging, maybe?

For good measure I tried out the nginx build flow locally on this branch and it worked fine, as far as I can tell. The files get copied into the nginx image and are accessible. Thank you for making sure that still works :rocket:

krysal: Indeed it's a leftover. Deleted, thanks!

justfile
```diff
@@ -69,7 +67,7 @@ EXEC_DEFAULTS := if IS_CI == "" { "" } else { "-T" }
 
 # Execute statement in service containers using Docker Compose
 exec +args:
```
AetherUnbound, line 69: I think we'll want to add `up` as a dependency for this, otherwise it'll fail if the stack isn't running:
```suggestion
exec +args: up
```

justfile
```diff
@@ -248,6 +246,14 @@ nginx upstream_url='api.openverse.engineering': collectstatic
       -e DJANGO_NGINX_GIT_REVISION="$(git rev-parse HEAD)" \
       openverse-api-nginx:latest
 
+# Launch a psql shell in the web container
+@dbshell:
+    just exec web python manage.py dbshell
+
+# Launch a pgcli shell in the web container (require typing credentials)
+@pgcli db_host="db":
+    just exec web pgcli -h {{ db_host }} openledger deploy
```
AetherUnbound, line 255: ~~We currently have `set dotenv-load := false` at the top of the file. If we set that to `true` instead, we could pass the `POSTGRES_PASSWORD` in here directly so it doesn't have to be typed manually. Do you think that'd be useful?~~

It looks like there's [no way to override the default `.env` file](https://github.com/casey/just#dotenv-integration) and these are present in `postgres/env.docker` anyway, ah well!

AetherUnbound (approved):
Looks and works great! 😄 I have one comment but otherwise this is good to go.

justfile
krysal: I was actually avoiding adding that dependency 😬 That's the reason why I don't use the `just dj` command here, so I don't have to see extra 10 lines in the terminal + the few seconds checking the container is already up every time I run something 😅

AetherUnbound: Ah that's fair 🙂 Definitely a bit of an annoyance waiting for the containers to start every time.

dhruvkb (approved):
LGTM!

PR #1015, opened 2022-11-29
Allow `/version` to be defined by environment variables

AetherUnbound:
## Fixes

This is necessary for continued work on https://github.com/WordPress/openverse-infrastructure/issues/175

## Description

This PR makes a few changes to the nginx target image for the API in order to prevent the need for mounting any files into the docker container. Previously, the `/version` was configured to return the contents of a `version.json` file. This file was created during deployment and mounted into the nginx image:

https://github.com/WordPress/openverse-infrastructure/blob/369d7858636797de276dbfc299b90b822a5ac527/modules/services/openverse-api/init.tpl#L79-L85

The issue with this is that we are trying to avoid the need for mounting anything into our API-related containers. This was the last necessary item that was being mounted. You can see that this endpoint is not served by the API by running `just up` and visiting http://localhost:50280/version. 

I've altered the image to construct the response from `/version` using environment variables rather than reading from a file on-disk. This involved a few changes to the environment substitution setup, which I was unfamiliar with (there are seemingly no references to `NGINX_ENVSUBST_FILTER` on GitHub where it's being defined by anyone other than us!). I added a few more comments to try and provide reference to what was being used, and where.

I also addressed an issue wherein we were erroneously referencing variables inside the nginx template with `\$` rather than just `$`. This was an artifact of constructing the config inside of the `init.tpl` file during deployment (which required that we escape the `$` character since it was used for its own interpolation). This was producing logs that looked like the following (note the extra `\` characters before field values):

> ` {"time_local":"\29/Nov/2022:00:33:08 +0000","remote_addr":"\172.17.0.1","remote_user":"\","request":"\GET / HTTP/1.1","status": "\400","host_header": "\localhost","body_bytes_sent":\155,"request_time":"\0.020","http_referrer":"\","http_user_agent":"\Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:106.0) Gecko/20100101 Firefox/106.0","upstream_response_time":\0.020,"http_x_forwarded_for":"\"}`

Lastly, I added a `just` recipe for running the nginx image locally with some suggestions on how to test that it works.

## Testing Instructions

1. Before checking out this branch, verify that the API returns a 404 on http://localhost:50280/version after running `just up`
1. Run `just nginx`
1. Visit http://localhost:9090/static/admin/css/base.css and verify that results are returned (this should be from api.openverse.engineering)
1. Visit http://localhost:9090/version and verify that you receive something similar to the following:

```json
{

    "release": "47521cd5b288c35afaf0f996015e8122bbc1ccf8",
    "environment": "local"

}
```

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -48,7 +48,9 @@ WORKDIR /app
 COPY nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
 COPY /static /app/static
 
-ENV NGINX_ENVSUBST_FILTER="DJANGO_UPSTREAM_URL"
+# Only environment variables with this prefix will be available in the template
+ENV NGINX_ENVSUBST_FILTER="DJANGO_NGINX_"
+ENV DJANGO_NGINX_ENVIRONMENT="local"
 
 #######
 # API #
--- a/api/nginx.conf.template
+++ b/api/nginx.conf.template
@@ -2,18 +2,18 @@ error_log /var/log/nginx/error.log;
 
 log_format json_combined escape=json
     '{'
-    '"time_local":"\$time_local",'
-    '"remote_addr":"\$remote_addr",'
-    '"remote_user":"\$remote_user",'
-    '"request":"\$request",'
-    '"status": "\$status",'
-    '"host_header": "\$host",'
-    '"body_bytes_sent":\$body_bytes_sent,'
-    '"request_time":"\$request_time",'
-    '"http_referrer":"\$http_referer",'
-    '"http_user_agent":"\$http_user_agent",'
-    '"upstream_response_time":\$upstream_response_time,'
-    '"http_x_forwarded_for":"\$http_x_forwarded_for"'
+    '"time_local":"$time_local",'
+    '"remote_addr":"$remote_addr",'
+    '"remote_user":"$remote_user",'
+    '"request":"$request",'
+    '"status": "$status",'
+    '"host_header": "$host",'
+    '"body_bytes_sent":$body_bytes_sent,'
+    '"request_time":"$request_time",'
+    '"http_referrer":"$http_referer",'
+    '"http_user_agent":"$http_user_agent",'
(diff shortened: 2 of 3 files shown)


AetherUnbound:
@sarayourfriend I've added you as a reviewer since you were the original author of #990 and are most familiar with the code!

api/Dockerfile
```diff
@@ -48,7 +48,9 @@ WORKDIR /app
 COPY nginx.conf.template /etc/nginx/templates/openverse-api.conf.template
 COPY /static /app/static
 
-ENV NGINX_ENVSUBST_FILTER="DJANGO_UPSTREAM_URL"
+# Only environment variables with this prefix will be available in the template
+ENV NGINX_ENVSUBST_FILTER="DJANGO_NGINX_"
+ENV DJANGO_NGINX_ENVIRONMENT="local"
```
sarayourfriend, line 53: We'll need to be sure to pass this environment variable in the ECS configuration as well, once we get there. That at least is constant for the environment, so not an issue.

sarayourfriend (approved):
Ah, thank you for catching this! I totally forgot about this requirement when working on the original issue. Thanks for fixing the logging as well, I had not noticed that.

sarayourfriend:
Regarding `NGINX_ENVSUBST_FILTER`, btw... I think I ended up having to read the various entrypoint scripts that the Nginx image comes with to discover this! It is not documented in the README for the image on docker hub :cry:

api/Dockerfile
AetherUnbound: Yes, same with the `DJANGO_NGINX_GIT_REVISION`!

stacimc (approved):
Thank you for the testing instructions! LGTM 😄

PR #1027, opened 2022-12-02
Replace grequests with asyncio solution

### LINKED ISSUE #486: Move off of grequests
## Problem

The API currently uses [the `grequests` library](https://github.com/spyoungtech/grequests) for handling some of the `head` requests it completes when returning results (specifically in [`validate_images.py`](https://github.com/WordPress/openverse-api/blob/main/api/catalog/api/utils/validate_images.py#L37)).

## Description

The README for the project says:

> Note: You should probably use requests-threads or requests-futures instead.

The project has also not had a release since 2020. We might want to consider using one of those proposed alternatives.

## Alternatives

Leave as is and hope for the best!

## Additional context

Some more context in Sentry: https://sentry.io/share/issue/061ba99fc3df4c23bdb7643d337bbda0/

## Implementation

sarayourfriend:
## Fixes

Fixes #486 by @AetherUnbound 

## Description

This PR removes `grequests` and replaces it with `aiohttp` and Django's `async_to_sync`.

It also introduces the lovely `pook` library for mocking HTTP requests. I found a bug there as well while working on this PR that I'll try to contribute an upstream fix for: https://github.com/h2non/pook/issues/82

## Testing Instructions

Follow the instructions in the README to run the application locally (be sure to `just build web` for the new dependencies!) and make an image or audio search and verify that it works the same as on `main`.

Verify that the changes to the unit tests make sense and that the same things are still being tested.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/utils/validate_images.py
```diff
@@ -57,24 +81,13 @@ def validate_images(query_hash, start_slice, results, image_urls):
...
     logger.debug(f"len(to_verify)={len(to_verify)}")
 
-    reqs = (
-        grequests.head(
-            url, headers=HEADERS, allow_redirects=False, timeout=2, verify=False
-        )
-        for url in to_verify.keys()
-    )
-    verified = grequests.map(reqs, exception_handler=_validation_failure)
+    verified = _make_head_requests(to_verify.keys())
```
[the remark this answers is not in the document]
sarayourfriend: Nope, excluding them was an oversight on my part. I've added them back in the latest commit. Thanks for catching that.

api/catalog/api/utils/validate_images.py
```diff
@@ -57,24 +81,13 @@ def validate_images(query_hash, start_slice, results, image_urls):
...
+
     # Cache newly verified image statuses.
     to_cache = {}
-    # relies on the consistenct of the order returned by `dict::keys()` which
-    # is safe as of Python 3 dot something
-    for idx, url in enumerate(to_verify.keys()):
+    for result in verified:
+        url, status = result
```
AetherUnbound, line 89: Would it be possible to just unpack on the iteration?
```suggestion
    for url, status in verified:
```

api/catalog/api/utils/validate_images.py
```diff
@@ -99,7 +112,7 @@ def validate_images(query_hash, start_slice, results, image_urls):
     for idx, url in enumerate(to_verify):
         cache_idx = to_verify[url]
         if verified[idx] is not None:
```
AetherUnbound, line 114: I might be wrong, won't this always be true? I believe the type of `verified[idx]` is `tuple[str, int]` which won't be `None` even in the default case.

I think this will still work because the default is now `-1`, but that also means the `else` code path won't be touched.

api/test/unit/utils/validate_images_test.py
```diff
@@ -22,48 +21,26 @@ def get_redis_connection(*args, **kwargs):
...
+        pook.head(pook.regex(r"https://example.org/\d"))
+        .times(len(results))
+        .reply(200)
+        .mock
+    )
+
     validate_images(query_hash, start_slice, results, image_urls)
 
-    requested_urls = [r.url for r in grequests.requests]
+    assert head_mock.calls > 0
```
AetherUnbound, line 41: Would we want this to be an exact check? Is there a reason why some requests might fail?
```suggestion
    assert head_mock.calls == len(result)
```

AetherUnbound (commented):
Wow, I'm so happy this was such a simple change! I tried running this locally and (after I remembered to force dead link filtering since we disabled that by default in #1043 so I could see the logging) verified everything worked 🚀

api/test/unit/utils/validate_images_test.py
sarayourfriend: Nope, you're right, it should check exactly the number of calls :+1:

api/catalog/api/utils/validate_images.py
```diff
@@ -28,6 +30,28 @@ def _get_expiry(status, default):
...
+        async with session.head(url, timeout=2, allow_redirects=False) as response:
+            return url, response.status
+    except aiohttp.ClientError as exception:
+        _validation_failure(exception)
+        return url, -1
+
+
+# https://stackoverflow.com/q/55259755
+@async_to_sync
+async def _make_head_requests(urls: list[str]):
```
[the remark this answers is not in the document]
sarayourfriend: It is a coroutine, but does the return type need to be explicitly set as a coroutine? This makes it seem like you do not need to wrap the return type annotation in `Awaitable` as it's implied by it being on an async function: https://docs.python.org/3/library/typing.html#callable

I think it can just be as you suggested, the type of the result of the coroutine.

api/catalog/api/utils/validate_images.py
sarayourfriend: I think you're right. I'll make the clean up here, thanks for catching it!

api/catalog/api/utils/validate_images.py
```diff
@@ -28,6 +30,28 @@ def _get_expiry(status, default):
     return config(f"LINK_VALIDATION_CACHE_EXPIRY__{status}", default=default, cast=int)
 
 
+async def _head(url: str, session: aiohttp.ClientSession):
```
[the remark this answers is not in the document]
sarayourfriend: Done in https://github.com/WordPress/openverse-api/pull/1027/commits/582d563ada7def1b8ded7b0c80d3b994fe22deda

api/catalog/api/utils/validate_images.py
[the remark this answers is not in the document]
sarayourfriend: Done in https://github.com/WordPress/openverse-api/pull/1027/commits/582d563ada7def1b8ded7b0c80d3b994fe22deda

api/catalog/api/utils/validate_images.py
sarayourfriend: Yup, changed the whole statement to a comprehension in https://github.com/WordPress/openverse-api/pull/1027/commits/582d563ada7def1b8ded7b0c80d3b994fe22deda

sarayourfriend:
@AetherUnbound @obulat please re-review at your convenience. Let me know if there are specific things you'd like to see changed.

api/catalog/api/utils/validate_images.py
```diff
@@ -145,6 +151,6 @@ def validate_images(query_hash, start_slice, results, image_urls):
     )
 
 
-def _validation_failure(request, exception):
+def _validation_failure(exception):
```
obulat, line 154: Nit: can we name this function with a verb, something like `_log_validation_failure`?

api/catalog/api/utils/validate_images.py
```diff
@@ -28,6 +30,28 @@ def _get_expiry(status, default):
...
+        _validation_failure(exception)
+        return url, -1
+
+
+# https://stackoverflow.com/q/55259755
+@async_to_sync
+async def _make_head_requests(urls: list[str]) -> list[tuple[str, int]]:
+    tasks = []
+    async with aiohttp.ClientSession(headers=HEADERS) as session:
+        for url in urls:
```
obulat, line 47: Can this also be converted to a list comprehension, or do you think it makes it more difficult to understand?
`tasks = [asyncio.ensure_future(_head(url, session)) for url in urls`]

api/catalog/api/utils/validate_images.py
```diff
@@ -28,6 +30,28 @@ def _get_expiry(status, default):
...
+            task = asyncio.ensure_future(_head(url, session))
+            tasks.append(task)
+        responses = asyncio.gather(*tasks)
+        await responses
+    return responses.result()
+
+
```
obulat, line 54: Nit: maybe we could also add type annotations to `validate_images`?

api/catalog/api/utils/validate_images.py
```diff
@@ -84,11 +84,7 @@ def validate_images(query_hash, start_slice, results, image_urls):
     verified = _make_head_requests(to_verify.keys())
 
     # Cache newly verified image statuses.
-    to_cache = {}
-    for result in verified:
-        url, status = result
-        cache_key = CACHE_PREFIX + url
-        to_cache[cache_key] = status
+    to_cache = {CACHE_PREFIX + url: status for url, status in verified}
```
AetherUnbound, line 87: Much simpler! 🤩

AetherUnbound (approved):
Looks great! 👍🏼 from me but +1 to Olga's suggestions as well

sarayourfriend:
@obulat I've addressed your comments. Please re-review when you have a moment, thanks.

obulat (approved):
The code looks great, works well locally, and the tests are much simpler. Hopefully, this is a start of a quick transition to async (and the DRF is ready soon).

I wonder if we should add several 'dead' links to the local sample data so that we could see the dead link filtering locally, too.

sarayourfriend:
> I wonder if we should add several 'dead' links to the local sample data so that we could see the dead link filtering locally, too.

According to the dead link filtering test documentation there are currently some number of dead links in the test data: https://github.com/WordPress/openverse-api/blob/ea2162f42dc1d73baa36846515030f78b3dfc593/api/test/dead_link_filter_test.py#L188-L195

The test doesn't really check whether different results are returned if filtering is enabled or not, so the comment could be out of date.

PR #1032, opened 2022-12-06
Make page_count self-truncating when we detect the last page of a query

### LINKED ISSUE #196: Dead link filtering reduces number of queryable pages
## Description

The dead link filtering algorithm significantly limits the pages that can be retrieved.

## Reproduction

0. Go to audio search endpoint: `http://localhost:8000/v1/audio/`. Note the number of pages.
1. Go to a page number that's quite high but less than 250, like 200: `http://localhost:8000/v1/audio/?page=200`
2. See the deep pagination error. This should not be happening.

## Expectation

Page count calculation should be correct and should allow pagination upto the stated limit.

## Resolution

sarayourfriend:
## Fixes

Fixes #196  by @dhruvkb 

## Description

This relatively simple change "truncates" the page count for a query to the current page number if we receive fewer results on the page than are able to fill out the page based on the page_size. This follows the suggestion from the issue by @AetherUnbound.

## Testing Instructions

Checkout the branch and run the project according to the README.

Test the result views and pay attention to the page count. Test queries that have a single page, multiple pages, 20 pages, etc. Tip: use increasingly specific terms to get smaller returns from the test data.

Review the unit test cases added and confirm they are sensible.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -510,11 +510,10 @@ def _get_result_and_page_count(
         return 0, 1
 
     result_count = response_obj.hits.total.value
-    page_count = int(result_count / page_size)
-    if result_count > page_size and page_count % page_size != 0:
-        page_count += 1
+    page_count = ceil(result_count / page_size)
+
     if len(results) < page_size:
-        if page_count == 0:
+        if page_count == 1:
             result_count = len(results)
 
         # If we have fewer results than the requested page size and are
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -511,7 +511,7 @@ def _get_result_and_page_count(
 
     result_count = response_obj.hits.total.value
     page_count = int(result_count / page_size)
-    if page_count % page_size != 0:
+    if result_count > page_size and page_count % page_size != 0:
         page_count += 1
     if len(results) < page_size:
         if page_count == 0:
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -410,7 +410,7 @@ def search(
     )
 
     result_count, page_count = _get_result_and_page_count(
-        search_response, results, page_size
+        search_response, results, page_size, page
     )
     return results or [], page_count, result_count
 
@@ -447,7 +447,7 @@ def related_media(uuid, index, request, filter_dead):
         s, start, end, page_size, response, request, filter_dead
     )
 
-    result_count, _ = _get_result_and_page_count(response, results, page_size)
+    result_count, _ = _get_result_and_page_count(response, results, page_size, page)
 
     return results or [], result_count
 
@@ -496,7 +496,7 @@ def get_sources(index):
 
 
 def _get_result_and_page_count(
-    response_obj: Response, results: list[Hit] | None, page_size: int
+    response_obj: Response, results: list[Hit] | None, page_size: int, page: int
 ) -> tuple[int, int]:
     """
     Elasticsearch does not allow deep pagination of ranked queries.
@@ -513,7 +513,16 @@ def _get_result_and_page_count(
     page_count = int(result_count / page_size)
     if page_count % page_size != 0:
         page_count += 1
-    if len(results) < page_size and page_count == 0:
-        result_count = len(results)
+    if len(results) < page_size:
+        if page_count == 0:
+            result_count = len(results)
+
+        # If we have fewer results than the requested page size and are
+        # not on the first page that means that we've reached the end of
+        # the query and can set the page_count to the currently requested
+        # page. This means that the `page_count` can change during
+        # pagination for the same query, but it's the only way to
+        # communicate under the current v1 API that a query has been exhausted.
+        page_count = page
 
     return result_count, page_count
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -506,8 +506,14 @@ def _get_result_and_page_count(
     :param results: The list of filtered result Hits.
     :return: Result and page count.
     """
-    if results is None:
-        return 0, 1
+    if not results:
+        return 0, 0
+
+    if page == 0:
+        # Realistically ``page`` shouldn't be submitted as 0, but if it is
+        # we treat it synonymously as the first page. We can set that here
+        # explicitly so that the behaviour if clear.
+        page = 1
 
     result_count = response_obj.hits.total.value
     page_count = ceil(result_count / page_size)
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -509,12 +509,6 @@ def _get_result_and_page_count(
     if not results:
         return 0, 0
 
-    if page == 0:
-        # Realistically ``page`` shouldn't be submitted as 0, but if it is
-        # we treat it synonymously as the first page. We can set that here
-        # explicitly so that the behaviour if clear.
-        page = 1
-
     result_count = response_obj.hits.total.value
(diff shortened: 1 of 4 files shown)


sarayourfriend:
I am debugging the tests that are failing. It is easy to fix them, I just change the page_count from 0 to 1 to match the new expected result of the page counting operation... however, now I am second guessing whether `page_count` should be 1 or 0 when there is only a single page of results. Do any @WordPress/openverse-api folks know what the intention is? My intuition is that it should be 1 as there is a single page of results. I've made changes to reflect that, but again, I'm second guessing myself that I actually understand the original intention of the code, in particular because the "dangling page of results" code was obscure to me (why did we modulo page count over page size? this made no sense to me).

Thanks for any insight here.

zackkrida:
This was discussed a bit in the community meeting in the [Make WP Slack](https://make.wordpress.org/chat/):

https://wordpress.slack.com/archives/C02012JB00N/p1670945242644929

@obulat shared that the frontend expects the page count to be `1`, and I mentioned that an api response with these values:

```json
{
  "page": 1,
  "page_count": 1
}
```

feels internally consistent. 

Additionally, the API currently requires users to pass a `page` value of 1 or higher:

https://api.openverse.engineering/v1/audio/?q=puppy&page=0

```json
{
    "detail": {
        "page": [
            "Ensure this value is greater than or equal to 1."
        ]
    }
}
```

sarayourfriend:
That is all in line with what my expectations would be as well. Thanks for uncovering those details, Zack. I'll undraft this now :slightly_smiling_face:

sarayourfriend:
I will be AFK starting tomorrow so Dhruv and Olga, if y'all review and approve this feel free to merge it at your convenience if you're working.

api/catalog/api/controllers/search_controller.py
```diff
@@ -510,10 +510,18 @@ def _get_result_and_page_count(
...
-        page_count += 1
-    if len(results) < page_size and page_count == 0:
-        result_count = len(results)
+    page_count = ceil(result_count / page_size)
+
+    if len(results) < page_size:
+        if page_count == 1:
+            result_count = len(results)
+
+        # If we have fewer results than the requested page size and are
```
obulat, line 519: I was afraid that this change will mean that we have to update the frontend code to account for the `page_count` changing between request, but we actually do update it on every request, so it's fine :)

api/test/unit/controllers/test_search_controller.py
```diff
@@ -13,36 +13,51 @@
...
-    "total_hits, real_result_count, page_size, expected",
+    "total_hits, real_result_count, page_size, page, expected",
     [
         # No results
-        (0, 0, 10, (0, 0)),
+        (0, 0, 10, 0, (0, 0)),
         # Setting page size to 0 raises an exception
         pytest.param(
-            0, 0, 0, (0, 0), marks=pytest.mark.raises(exception=ZeroDivisionError)
+            0, 0, 0, 0, (0, 0), marks=pytest.mark.raises(exception=ZeroDivisionError)
```
obulat, line 22: In practice, this function can never be called with 0 as the `page_size` value because we have a validator on `page_size` with a minimum value of 1.

api/test/unit/controllers/test_search_controller.py
```diff
@@ -13,36 +13,51 @@
...
         ),
         # Fewer results than page size leads to max of result total
-        (5, 5, 10, (5, 0)),
+        (5, 5, 10, 0, (5, 0)),
```
obulat, line 25: I don't understand why the page_count returned here is 0. Isn't it supposed to be 1 even if we have 1 result?

api/test/unit/controllers/test_search_controller.py
sarayourfriend: Correct, but unit tests generally are not meant to embed meta-knowledge of the application like that. The list here tests the unit's possible (relevant and irrelevant to the application) inputs which makes it easier to verify the intended behaviour (at least in my experience).

api/test/unit/controllers/test_search_controller.py
sarayourfriend: It should be, yes. I suppose it should only ever return 0 if there are 0 results. I'll look into this, thanks for catching it.

sarayourfriend: This turned out to mostly be a test quirk, as the page sent to the method was `0` even though that's not a real page that ever really gets requested because we do not 0 index the pages. However, it's easy to correct for it by treating `page = 0` the same as `page = 1`.

sarayourfriend:
@obulat Can you re-review this? @dhruvkb when you have a chance this needs a second review. Thanks@

api/test/unit/controllers/test_search_controller.py
obulat: Thank you for looking into this, @sarayourfriend! Do you think a comment about it would be useful in the tests?

api/catalog/api/controllers/search_controller.py
```diff
@@ -506,14 +506,28 @@ def _get_result_and_page_count(
     :param results: The list of filtered result Hits.
     :return: Result and page count.
     """
-    if results is None:
-        return 0, 1
+    if not results:
+        return 0, 0
+
+    if page == 0:
```
obulat, line 512: Honestly, this even seems redundant since we have the validators that prevent the users from ever setting the `page` to 0, and if you send the request with `page=0`, you get a 400 response: https://api.openverse.engineering/v1/images/?q=nature&page=0.

api/test/unit/controllers/test_search_controller.py
```diff
@@ -13,36 +13,51 @@
...
+        (25, 5, 5, 0, (25, 5)),
         # Unevenly divisible number of pages
-        (21, 5, 5, (21, 5)),
+        (21, 5, 5, 0, (21, 5)),
         # My assumption would be that this yields (20, 4), but the code is such that
         # when the "natural" page count can't be cleanly divisible by the page size,
         # We increment it plus one. Why would that be the case? 20 results, with 5
         # results per-page, would seem to result in 4 pages total not 5 🤷‍♀️
-        (20, 5, 5, (20, 5)),
+        (20, 5, 5, 0, (20, 4)),
```
obulat, line 40: I guess we can remove the comment above now?

obulat (approved):
Fantastic test cases! 
I think one comment is redundant now that these changes have been introduced, note in the inline comments.

api/catalog/api/controllers/search_controller.py
sarayourfriend: That's true. I had mixed feelings about it. Do you think I should instead update the tests to only pass inputs we know would be valid according to the API? That doesn't seem unreasonable, really...

obulat: If we leave it as is, then it'll be very easy to allow for `page=0` in the API request in the future if we ever decide to do that. However, it will also mean that an unnecessary clause will run for every request. Even though it's very (extremely) short. I'm also not sure, but since we're not likely to change the allowed pages,, it seems better to save a tiny amount of resources on each request :)

sarayourfriend: I changed it to remove the impossible edge case, thanks for the feedback :pray:

api/test/unit/controllers/test_search_controller.py
sarayourfriend: Done in https://github.com/WordPress/openverse-api/pull/1032/commits/bfa0802d00cdc4282e2063b8f91fac8e66f1ff9a

dhruvkb (approved):
The change makes sense and the tests are very comprehensive. If the frontend automatically works with it, that's icing on the cake!

PR #1034, opened 2022-12-07
added unit test cases for pagination

### LINKED ISSUE #401: Pagination unit tests
## Current Situation

A bug that was recently fixed in #398 involved the `pagination.py` utilities, which currently don't appear to have tests.

## Suggested Improvement

It would be great to add unit tests for this file!

## Implementation

prikshit-1103:
## Fixes
Fixes #401.

## Description
written unit test cases for pagination class and for pagination in case of the API call

### DIFF
--- a/api/test/unit/utils/test_Pagination.py
+++ b/api/test/unit/utils/test_Pagination.py
@@ -1,0 +1,60 @@
+import pytest
+from catalog.api.utils.pagination import StandardPagination
+from django.core.paginator import Paginator as DjangoPaginator
+from django.db import models
+from django.test import TestCase
+ 
+from rest_framework import (
+   exceptions, filters, generics, pagination, serializers, status
+)
+from rest_framework.pagination import PAGE_BREAK, PageLink
+from rest_framework.request import Request
+from rest_framework.test import APIRequestFactory
+ 
+factory = APIRequestFactory()
+ 
+class TestPagination:
+ 
+ def setup_method(self):
+       class PassThroughSerializer(serializers.BaseSerializer):
+           def to_representation(self, item):
+               return item
+ 
+       class EvenItemsOnly(filters.BaseFilterBackend):
+           def filter_queryset(self, request, queryset, view):
+               return [item for item in queryset if item % 2 == 0]
+ 
+       class BasicPagination(pagination.PageNumberPagination):
+           page_size = 5
+           page_size_query_param = 'page_size'
+           max_page_size = 20
+ 
+       self.view = generics.ListAPIView.as_view(
+           serializer_class=PassThroughSerializer,
+           queryset=range(1, 101),
+           filter_backends=[EvenItemsOnly],
+           pagination_class=BasicPagination
+       )
+ def test_setting_page_size(self):
+ 
+       request = factory.get('/', {'page_size': 10})
+       response = self.view(request)
+       assert response.status_code == status.HTTP_200_OK
+       assert response.data == {
+           'results': [2, 4, 6, 8, 10, 12, 14, 16, 18, 20],
+           'previous': None,
+           'next': 'http://testserver/?page=2&page_size=10',
+           'count': 50
+       }
+ 
+ def test_setting_page_size_to_zero(self):
+  
+       request = factory.get('/', {'page_size': 0})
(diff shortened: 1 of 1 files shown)


api/test/unit/utils/test_Pagination.py
```diff
@@ -0,0 +1,60 @@
...
+ def setup_method(self):
+       class PassThroughSerializer(serializers.BaseSerializer):
+           def to_representation(self, item):
+               return item
+ 
+       class EvenItemsOnly(filters.BaseFilterBackend):
+           def filter_queryset(self, request, queryset, view):
+               return [item for item in queryset if item % 2 == 0]
+ 
+       class BasicPagination(pagination.PageNumberPagination):
```
dhruvkb, line 27: The goal of the issue #401 was to test our `StandardPagination` class.

https://github.com/WordPress/openverse-api/blob/e3a5cfb9e7a88f8afb8167b6e6072fa5115e1e53/api/catalog/api/utils/pagination.py#L6

You should change the pagination class in your view to use `StandardPagination`.

api/test/unit/utils/test_Pagination.py
```diff
@@ -0,0 +1,60 @@
...
+ def test_setting_page_size_to_zero(self):
+  
+       request = factory.get('/', {'page_size': 0})
+       response = self.view(request)
+       assert response.status_code == status.HTTP_200_OK
+       assert response.data == {
+           'results': [2, 4, 6, 8, 10],
+           'previous': None,
+           'next': 'http://testserver/?page=2&page_size=0',
+           'count': 502
```
dhruvkb, line 59: The test is failing because the count is supposed to be 50.

api/test/unit/utils/test_Pagination.py
```diff
@@ -0,0 +1,60 @@
...
+from rest_framework import (
+   exceptions, filters, generics, pagination, serializers, status
+)
+from rest_framework.pagination import PAGE_BREAK, PageLink
+from rest_framework.request import Request
+from rest_framework.test import APIRequestFactory
+ 
+factory = APIRequestFactory()
+ 
+class TestPagination:
```
dhruvkb, line 16: Please use a consistent indentation of 4 spaces per indent level.

api/test/unit/utils/test_Pagination.py
```diff
@@ -0,0 +1,60 @@
...
+       response = self.view(request)
+       assert response.status_code == status.HTTP_200_OK
+       assert response.data == {
+           'results': [2, 4, 6, 8, 10, 12, 14, 16, 18, 20],
+           'previous': None,
+           'next': 'http://testserver/?page=2&page_size=10',
+           'count': 50
+       }
+ 
+ def test_setting_page_size_to_zero(self):
```
dhruvkb, line 50: Test names should be a statement of expectation.

```suggestion
 def test_setting_page_size_to_zero_uses_default(self):
```

api/test/unit/utils/test_Pagination.py
```diff
@@ -0,0 +1,60 @@
...
+           page_size_query_param = 'page_size'
+           max_page_size = 20
+ 
+       self.view = generics.ListAPIView.as_view(
+           serializer_class=PassThroughSerializer,
+           queryset=range(1, 101),
+           filter_backends=[EvenItemsOnly],
+           pagination_class=BasicPagination
+       )
+ def test_setting_page_size(self):
```
dhruvkb, line 38:
```suggestion
 def test_page_size_can_be_set_via_query_params(self):
```

dhruvkb (changes requested):
1. Please follow the PR message template. Not using it makes it harder to review PRs and breaks the link between the issue and the resolution.
2. Please lint the changes. The lint workflow has failed. To run lint, follow these steps.
	1. `just install`
	2. `just precommit`
	3. `just lint`
3. Please also add a test where page size is set to a non-integer value like a string or a float. You can [parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html) the second test to accept all manners of invalid values.

krysal:
@prikshit-1103 I'll close this, but if you want to resume it at any time, please feel free to send another PR.

PR #1040, opened 2022-12-12
Reduce DB queries needed in search results

### LINKED ISSUE #1036: `attribution` is null in search results view
## Description

Example where it is not null: https://api.openverse.engineering/v1/images/0aff3595-8168-440b-83ff-7a80b65dea42/
Example where it is null: https://api.openverse.engineering/v1/images/?q=dog

See any of the results and attribution will be null for them. If you go to the detail view for the result, it will be populated.

See this comment for an explanation of why and a potential fix: https://github.com/WordPress/openverse-api/issues/1036#issuecomment-1345667141

## Reproduction

1.  Go to a search results list view and see attribution is null.

## Additional context

Discovered by @zackkrida and @krysal in Make Slack here: https://wordpress.slack.com/archives/C02012JB00N/p1670442907182319 (Requires a free account from https://chat.wordpress.org to access).

## Resolution

dhruvkb:
## Fixes

Fixes #1036 by @sarayourfriend 

## Description

Currently a search result needs several DB queries to populate the page.
1. Get `Hit` instances from ES
2. Get DB rows corresponding to the hits.
3. For each audio row, search for `AudioSet` and `MatureAudio` instances. For each image search for the `MatureImage` instances.

This PR
- uses ORM-fu to bring the total queries down to 1!
- improves the structure of reporting models to use FK 
- improves the admin experience using autocomplete field
- uses `OpenledgerModel` in more places for consistent timestamping fields

This is an example query for audio, it includes the related `matureaudio` and `audioset` data too!

```sql
SELECT 
  "audio"."id", 
  "audio"."created_on", 
  "audio"."updated_on", 
  "audio"."identifier", 
  "audio"."foreign_identifier", 
  "audio"."title", 
  "audio"."foreign_landing_url", 
  "audio"."creator", 
  "audio"."creator_url", 
  "audio"."thumbnail", 
  "audio"."provider", 
  "audio"."url", 
  "audio"."filesize", 
  "audio"."filetype", 
  "audio"."watermarked", 
  "audio"."license", 
  "audio"."license_version", 
  "audio"."source", 
  "audio"."last_synced_with_source", 
  "audio"."removed_from_source", 
  "audio"."view_count", 
  "audio"."tags", 
  "audio"."tags_list", 
  "audio"."category", 
  "audio"."meta_data", 
  "audio"."bit_rate", 
  "audio"."sample_rate", 
  "audio"."audio_set_foreign_identifier", 
  "audio"."audio_set_position", 
  "audio"."genres", 
  "audio"."duration", 
  "audio"."alt_files", 
  "audioset"."id", 
  "audioset"."created_on", 
  "audioset"."updated_on", 
  "audioset"."foreign_identifier", 
  "audioset"."title", 
  "audioset"."foreign_landing_url", 
  "audioset"."creator", 
  "audioset"."creator_url", 
  "audioset"."thumbnail", 
  "audioset"."provider", 
  "audioset"."url", 
  "audioset"."filesize", 
  "audioset"."filetype", 
  "api_matureaudio"."updated_on", 
  "api_matureaudio"."created_on", 
  "api_matureaudio"."media_obj_id" 
FROM 
  "audio" 
  LEFT OUTER JOIN "audioset" ON (
    "audio"."audio_set_foreign_identifier" = "audioset"."foreign_identifier" 
    AND "audio"."provider" = "audioset"."provider"
  ) 
  LEFT OUTER JOIN "api_matureaudio" ON (
    "audio"."identifier" = "api_matureaudio"."media_obj_id"
  ) 
WHERE 
  "audio"."identifier" = 'f97384f6-a5ba-4668-881f-f616904651f0' :: uuid 
LIMIT 
  21;
```

## Testing Instructions

Screenshot 2022-12-12 at 12 47 41 PM

0. Enable logging for Django DB queries as explained [here](https://stackoverflow.com/a/20161527). 
1. You can also define a formatter with some text like "[db]" for filtering the output using `grep`.
2. Start logging the output: `just dc logs -f web | grep -E '\[db]'`.
3. Visit the search results and single result pages of audio and image media types.
4. You should see a single query for each page view in the logs.

## Checklist

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/catalog/api/models/audio.py
```diff
@@ -141,6 +142,14 @@ class Audio(AudioFileMixin, AbstractMedia):
     category: eg. music, sound_effect, podcast, news & audiobook
     """
 
+    audioset = models.ForeignObject(
+        to="AudioSet",
+        on_delete=models.DO_NOTHING,
```
sarayourfriend, line 147: What is the intention behind using "do nothing" for this?

api/test/dead_link_filter_test.py
```diff
@@ -27,6 +27,11 @@ def get_redis_connection(*args, **kwargs):
     fake_redis.client().close()
 
 
+@pytest.fixture(autouse=True)
+def turn_off_db_read(monkeypatch):
```
sarayourfriend, line 31: Why is this needed for these tests? Can you add a comment explaining it?

api/test/media_integration.py
```diff
@@ -102,12 +102,13 @@ def report(media_type, fixture):
...
     response = requests.post(
         f"{API_URL}/v1/{media_type}/{test_id}/report/",
-        {
+        json={
             "reason": "mature",
             "description": "This item contains sensitive content",
         },
         verify=False,
     )
+    print(response.json())
```
sarayourfriend, line 111: Is this leftover debugging code?

```suggestion
```

sarayourfriend (changes requested):
This is exciting! One request I would have is to use Django's query count checks to ensure that search hits for each image and audio only ever require at most a single query: https://pytest-django.readthedocs.io/en/latest/helpers.html#django-assert-num-queries

We also need to address the renamed column. If it is absolutely necessary for some reason (we need to justify this very carefully considering the implications of downtime) then we need to carefully plan it. If it can at all be avoided, either via a property to unify the API or, if the column definitely needs to be renamed, then via the multi-step zero-downtime approach I described in the other comment (though again, carefully justified, considering how much work it is to do it, including needing to write a command to incrementally move the data between the two columns).

Exciting PR nonetheless :slightly_smiling_face:

api/catalog/api/examples/audio_responses.py
```diff
@@ -64,10 +64,7 @@
         "results": [
             base_audio
             | {
-                "fields_matched": ["title"],
-                "audio_set": None,
-                "attribution": None,
-                "filesize": None,
```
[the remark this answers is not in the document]
dhruvkb: These aren't removed, they're being filled with real info, so I had to stop overriding their values with `None`.

api/catalog/api/examples/audio_responses.py
```diff
@@ -64,10 +64,7 @@
...
+                "fields_matched": [],
```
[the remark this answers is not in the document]
dhruvkb: Yes, I just realised we lose this, so I'm working on bringing it back. Thanks for noting this.

api/test/media_integration.py
dhruvkb: Yes 🤦‍♂️. Thanks for catching this!

api/catalog/api/models/audio.py
dhruvkb: We don't want the audio hard-linked to audio sets. Both tables are destroyed and recreated during ingestion so
- we don't want the deletes to cascade
- we cannot maintain DB integrity during ingestion

Also, this behaviour is identical to the one we had before, so I went for minimal disruption.

api/catalog/api/models/audio.py
```diff
@@ -227,7 +230,28 @@ class Meta(AbstractMedia.Meta):
         db_table = "audio"
 
 
-class DeletedAudio(AbstractDeletedMedia):
+class AbstractAudioRelation(AbstractAbstractMediaRelation):
+    """
+    This class should be inherited by models aiming to have a one-to-one mapping with
+    `Audio` instances. Note that the mapping is not enforced at the DB level so that
```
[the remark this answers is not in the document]
dhruvkb: Same!

api/catalog/api/examples/audio_responses.py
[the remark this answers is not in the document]
dhruvkb: Restored in 640211e.

api/test/dead_link_filter_test.py
dhruvkb: Added comment in c3681fe.

api/test/media_integration.py
dhruvkb: Removed in 516925b.

api/catalog/api/migrations/0053_audio_audioset.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.1.2 on 2022-12-12 08:30
```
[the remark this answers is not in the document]
dhruvkb: Done 👍 in a6e3759.

api/catalog/api/models/audio.py
sarayourfriend: It to me a second to understand how `on_delete` might interact with `db_constraint=False`. I was trying to square it with the Postgres behaviour, where the constraint enforcement is delayed until the end of the transaction in the "do nothing" setting, but I suppose that Django does not enforce it "manually" for us, does it?

api/test/unit/views/media_views_test.py
```diff
@@ -67,6 +70,27 @@ def send(session, req, **kwargs):
...
+)
+def test_retrieve_query_count(api_client, media_type, media_factory, viewset, requests):
+    media = media_factory.create()
+    factory = APIRequestFactory()
+    request = factory.get(f"/v1/{media_type}/{media.identifier}/")
+
+    # This number goes up without `select_related` in the viewset queryset.
+    with pytest_django.asserts.assertNumQueries(1):
+        view = viewset.as_view({"get": "retrieve"})
+        res = view(request, identifier=media.identifier)
```
sarayourfriend, line 89: Totally optional, but is it possible to simplify this using the `api_client` fixture? It actually exercises the entire route including middleware, not just the view itself, so is slightly more comprehensive. It also allows the user to avoid having to instantiate the view and call it themselves like we've done here.

api/catalog/api/models/media.py
```diff
@@ -135,7 +135,7 @@ def __str__(self):
         return f"{self.__class__.__name__}: {self.identifier}"
 
 
-class AbstractMediaReport(OpenLedgerModel):
+class AbstractMediaReport(models.Model):
```
sarayourfriend, line 138: If we wanted to share the same base class, it is possible to tell Django to map a column to a different column name than the attribute name. We'd just need to overwrite the attribute in the sub-classing model. Totally up to you though, this approach also works fine.

sarayourfriend (commented):
@dhruvkb Are you planning to add query counting tests for the search results view in addition to the new ones that you added for the single result page? I think the search results view is actually the one that would most benefit from that "gut check" given that it could be easy to accidentally make n + 1 queries.

The new changes are excellent though, thanks for being so comprehensive with them.

api/catalog/api/models/audio.py
dhruvkb: I'm sorry I didn't understand your question. But I'll take a stab to answer anyway.

1. `on_delete="DO_NOTHING"` prevents the model itself from being deleted or the field value from being changed when the referenced object goes away.
2. `db_contraint=False` allows the field to accept values that are not present in the referenced table.

These are both for situations that could happen during ingestion as both `audio` and `audioset` tables are recreated and filled separately.

api/catalog/api/models/media.py
dhruvkb: It definitely is, but it would complicate the PR even more, mix a bunch of different concerns in a migration and leave the PR non-atomic. We can circle back to it later.

api/test/unit/views/media_views_test.py
dhruvkb: IIRC I tried that first but that makes a request to the endpoint from outside so `assertNumQueries` always sees 0 queries.

api/catalog/api/models/audio.py
sarayourfriend: Right, that is my understanding of the current implementation as well, but critically it's the usage db_constraint=False in conjunction with DO_NOTHING that results in that. In regular db-enforced FK constraints in Postgres, the "NO ACTION" delete effect still enforces the constraint on a DB level, it just delays the check until the end of the transaction instead of immediately: https://www.postgresql.org/docs/15/ddl-constraints.html#DDL-CONSTRAINTS-EXCLUSION

> Restricting and cascading deletes are the two most common options. RESTRICT prevents deletion of a referenced row. NO ACTION means that if any referencing rows still exist when the constraint is checked, an error is raised; this is the default behavior if you do not specify anything. (**The essential difference between these two choices is that NO ACTION allows the check to be deferred until later in the transaction, whereas RESTRICT does not.**) CASCADE specifies that when a referenced row is deleted, row(s) referencing it should be automatically deleted as well. There are two other options: SET NULL and SET DEFAULT. These cause the referencing column(s) in the referencing row(s) to be set to nulls or their default values, respectively, when the referenced row is deleted. Note that these do not excuse you from observing any constraints. For example, if an action specifies SET DEFAULT but the default value would not satisfy the foreign key constraint, the operation will fail.

I've added bold emphasis to the relevant section of the quoted paragraph.

Critically: "no action" (or "do nothing" as Django calls it) _still enforces the constraint_, just not on the DB level. I'm _assuming_ that Django doesn't enforce it programmatically because [the documentation for `db_constraint`](https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.ForeignKey.db_constraint) doesn't mention it doing anything of that nature and even that it's "very bad for data integrity" which seems to imply that using this setting makes it easy to get into a bad state (which we already know anyway :slightly_smiling_face:)

In any case, I was mostly just thinking out loud to understand how "do nothing" and db_constraint=False were working together to make it so that we can freely delete media items without causing damage to media reports, waveforms, and other related data.

api/catalog/api/models/media.py
sarayourfriend: > mix a bunch of different concerns in a migration and leave the PR non-atomic

In what sense would it make the PR non-atomic? My understanding is that it would allow us to re-use the existing column with a unified API on the Python side without needing to worry about the column names being slightly different (which the code _would_ need to care about when we're referencing the columns, but not if we did the aliasing I suggested).

I agree it can be done later, but I don't understand how this would affect the atomic nature of the PR. At the very least it might cut down on extraneous, repeated documentation, especially if the column name change was abstracted out into a separate mixin or something to be used by the cases that need it.

I do think it should be done eventually though, if only to unify the attribute names on the code side so we don't have to try to remember (or be frustrated/confused when we forget) whether one model or another uses `created_on` or `created_at`.

dhruvkb:
@sarayourfriend I tried to write a similar test for the search results view.

```python
@pytest.mark.django_db
@pytest.mark.parametrize(
    "media_type, media_factory",
    [
        ("images", ImageFactory),
        ("audio", AudioFactory),
    ],
)
def test_list_query_count(api_client, media_type, media_factory, requests):
    media = media_factory.create()

    # This number goes up without `select_related` in the viewset queryset.
    with pytest_django.asserts.assertNumQueries(1):
        res = api_client.get(f"/v1/{media_type}/")

    assert res.status_code == 200
```


I kept getting this error:

```
FAILED test/unit/views/media_views_test.py::test_list_query_count[images-ImageFactory] - elasticsearch.exceptions.ConnectionError: ConnectionError('utf-8' codec can't decode byte 0xff in position 0: invalid start byte) caused by: UnicodeDecodeError('utf-8' codec can't decode byte 0xff in ...
FAILED test/unit/views/media_views_test.py::test_list_query_count[audio-AudioFactory] - elasticsearch.exceptions.ConnectionError: ConnectionError('utf-8' codec can't decode byte 0xff in position 0: invalid start byte) caused by: UnicodeDecodeError('utf-8' codec can't decode byte 0xff in ...
```

So I updated the ES package code to write the response bytes to a `.txt` file. 

```python
with open(f"/api/res.{time.time()}.txt", 'wb') as file:
    file.write(raw_data)
```

Examining with `file` said it was a JPEG image so I changed the extension and it was this.

res 1670995695 0843022


I can't understand why this is happening and how to fix it.

api/catalog/api/models/media.py
dhruvkb: I meant that the PR was working on reducing DB queries so fixing timestamp column names was not directly related to the cause of the PR but rather a drive-by refactor.

sarayourfriend: Ohhhh, sorry! Sure thing. The PR felt like it was touching lots of things so it was hard to tell what was in scope, but understood :+1: I misunderstood "atomic" to mean what it means with databases :facepalm: Apologies.

api/test/unit/views/media_views_test.py
```diff
@@ -68,6 +68,32 @@ def send(session, req, **kwargs):
...
+    "media_type, media_factory",
+    [
+        ("images", ImageFactory),
+        ("audio", AudioFactory),
+    ],
+)
+def test_list_query_count(api_client, media_type, media_factory):
+    media = media_factory.create()
+
+    results = ([MagicMock(identifier=str(media.identifier))], 1, 1)
```
sarayourfriend, line 82: The mock here makes sense and the new test is good, but having a single result makes the 1 query count suspicious :slightly_smiling_face: A full page of results would make the test more convincing long term as changes are made to the serializer and rest of the view.

dhruvkb: Updated the test to return 20 results in 48c463f.

krysal:
As much as I'd like to, I don't think I'll get to review this in the week, so tagging other folks in case they can/want to.

api/catalog/api/migrations/0052_relational_fields.py
```diff
@@ -0,0 +1,157 @@
...
+    "nsfw_reports_audio",
+]
+forward_query = "".join([
+    f"UPDATE {table} SET media_obj_id = identifier;"
+    for table in REPORTING_TABLES
+])
+reverse_query = "".join([
+    f"UPDATE {table} SET identifier = media_obj_id;"
+    for table in REPORTING_TABLES
+])
```
sarayourfriend, line 22: @dhruvkb Is there a way to avoid these data migration queries? Because they're on tables that are pretty small I don't think it will be a big issue, but in practice we'd like to avoid these whenever possible.

I've read through the migration a few times and I do not really understand what is going on in it. I'm confused by the `identifier` field getting altered and then removed from the report tables. The changes to `media_obj` are confusing as well...

When I run `makemigrations` locally, after deleting 0052, I get prompted to enter a default value for `media_obj` because it cannot be null (I guess because it is configured to be a PK?)...

I like the changes generally being made to the tables, but I'm a little worried about not understanding this migration and the differences when I try to generate a migration locally :confused:

sarayourfriend: Update: when I give a blank default (which I don't know if it will work due to the field type), it generates the following migration:

```py
# Generated by Django 4.1.3 on 2022-12-19 22:42

from django.db import migrations, models
import django.db.models.deletion


class Migration(migrations.Migration):

    dependencies = [
        ('api', '0051_delete_sourcelogo'),
    ]

    operations = [
        migrations.RemoveField(
            model_name='audioreport',
            name='identifier',
        ),
        migrations.RemoveField(
            model_name='deletedaudio',
            name='identifier',
        ),
        migrations.RemoveField(
            model_name='deletedimage',
            name='identifier',
        ),
        migrations.RemoveField(
            model_name='imagereport',
            name='identifier',
        ),
        migrations.RemoveField(
            model_name='matureaudio',
            name='identifier',
        ),
        migrations.RemoveField(
            model_name='matureimage',
            name='identifier',
        ),
        migrations.AddField(
            model_name='audio',
            name='audioset',
            field=models.ForeignObject(from_fields=['audio_set_foreign_identifier', 'provider'], null=True, on_delete=django.db.models.deletion.DO_NOTHING, to='api.audioset', to_fields=['foreign_identifier', 'provider']),
        ),
        migrations.AddField(
            model_name='audioreport',
            name='media_obj',
            field=models.ForeignKey(db_constraint=False, default='', help_text='The reference to the audio being reported.', on_delete=django.db.models.deletion.DO_NOTHING, related_name='audio_report', to='api.audio', to_field='identifier'),
            preserve_default=False,
        ),
        migrations.AddField(
            model_name='deletedaudio',
            name='media_obj',
            field=models.OneToOneField(db_constraint=False, default='', help_text='The reference to the deleted audio.', on_delete=django.db.models.deletion.DO_NOTHING, primary_key=True, related_name='deleted_audio', serialize=False, to='api.audio', to_field='identifier'),
            preserve_default=False,
        ),
        migrations.AddField(
            model_name='deletedimage',
            name='media_obj',
            field=models.OneToOneField(db_constraint=False, default='', help_text='The reference to the deleted image.', on_delete=django.db.models.deletion.DO_NOTHING, primary_key=True, related_name='deleted_image', serialize=False, to='api.image', to_field='identifier'),
            preserve_default=False,
        ),
        migrations.AddField(
            model_name='imagereport',
            name='media_obj',
            field=models.ForeignKey(db_constraint=False, default='', help_text='The reference to the image being reported.', on_delete=django.db.models.deletion.DO_NOTHING, related_name='image_report', to='api.image', to_field='identifier'),
            preserve_default=False,
        ),
        migrations.AddField(
            model_name='matureaudio',
            name='media_obj',
            field=models.OneToOneField(db_constraint=False, default='', help_text='The reference to the mature audio.', on_delete=django.db.models.deletion.DO_NOTHING, primary_key=True, related_name='mature_audio', serialize=False, to='api.audio', to_field='identifier'),
            preserve_default=False,
        ),
        migrations.AddField(
            model_name='matureimage',
            name='media_obj',
            field=models.OneToOneField(db_constraint=False, default='', help_text='The reference to the mature image.', on_delete=django.db.models.deletion.DO_NOTHING, primary_key=True, related_name='mature_image', serialize=False, to='api.image', to_field='identifier'),
            preserve_default=False,
        ),
    ]
```

I _think_ the difference between this version and the one in the PR currently is that the one you've generated moves the data over from identifier over to media_obj? Is that correct? The _way_ the migration does that is still confusing to me (I don't know what the AlterFields on `identifier` are for) though.

And I kind of wonder if we can avoid this complex migration to the DB queries specific work and do improvements to the admin as a separate issue and PR altogether? Do you think it is possible to split them out or are they both needed at the same time? It would make reviewing this easier if the admin improvements (which are very welcome!) were not included, especially if they're the source of the migrations :thinking:

dhruvkb: @sarayourfriend the changes were not for the admin but for making FK relations from the report, mature and deleted models to the media models. If we can work with calling the reference field as `identifier`, the migration gets a lot simpler as in 523338d.

dhruvkb: I also verified with the `sqlmigrate` command that the migration 0052 is largely a no-op! These are the only SQL changes in the migration:

```sql
BEGIN;
--
-- Add field audioset to audio
-- (no-op)
--
-- Alter field identifier on audioreport
CREATE INDEX "nsfw_reports_audio_identifier_ebe3a079" ON "nsfw_reports_audio" ("identifier");
--
-- Alter field identifier on deletedaudio
-- (no-op)
-- Alter field identifier on deletedimage
-- (no-op)
--
-- Alter field identifier on imagereport
CREATE INDEX "nsfw_reports_identifier_f0374e03" ON "nsfw_reports" ("identifier");
--
-- Alter field identifier on matureaudio
-- (no-op)
-- Alter field identifier on matureimage
-- (no-op)
COMMIT;
```

api/catalog/api/migrations/0052_update_reporting_models.py
```diff
@@ -0,0 +1,182 @@
...
+        migrations.RenameField(
+            model_name='imagereport',
+            old_name='created_at',
+            new_name='created_on',
+        ),
+        migrations.RenameField(
+            model_name='audioreport',
+            old_name='created_at',
+            new_name='created_on',
+        ),
```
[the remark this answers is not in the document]
dhruvkb: Sorry for digging up a past conversation (https://github.com/WordPress/openverse-api/pull/1040#discussion_r1046433040), but would `RenameField` operations be okay if we preserve the underlying column name and only change the model field?

dhruvkb:
The migration 0052 is largely a no-op with only 2 actual SQL statements! This can be verified with the `sqlmigrate` command.

```bash
$ just dc exec web python manage.py sqlmigrate api 0052
```

```sql
BEGIN;
--
-- Add field audioset to audio
-- (no-op)
--
-- Alter field identifier on audioreport
CREATE INDEX "nsfw_reports_audio_identifier_ebe3a079" ON "nsfw_reports_audio" ("identifier");
--
-- Alter field identifier on deletedaudio
-- (no-op)
--
-- Alter field identifier on deletedimage
-- (no-op)
--
-- Alter field identifier on imagereport
CREATE INDEX "nsfw_reports_identifier_f0374e03" ON "nsfw_reports" ("identifier");
--
-- Alter field identifier on matureaudio
-- (no-op)
--
-- Alter field identifier on matureimage
-- (no-op)
--
-- Rename field identifier on audioreport to media_obj
-- (no-op)
--
-- Rename field identifier on deletedaudio to media_obj
-- (no-op)
--
-- Rename field identifier on deletedimage to media_obj
-- (no-op)
--
-- Rename field identifier on imagereport to media_obj
-- (no-op)
--
-- Rename field identifier on matureaudio to media_obj
-- (no-op)
--
-- Rename field identifier on matureimage to media_obj
-- (no-op)
COMMIT;
```

AetherUnbound (approved):
Okay, I _think_ I understand everything happening here! This is definitely some ORM magic 🤯 I ran this locally as described and confirmed that only 1 query is now made for search results. I also migrated my local database, then hopped back over to main to make sure we were backwards compatible after the migration. I wasn't able to raise any errors after several searches and poking around in the admin UI. Based on that _and_ your lovely SQL plan output (thanks for posting that @dhruvkb) I think this is good to go 🙂

sarayourfriend (approved):
Brilliant, thanks @dhruvkb. Excellent work sorting through the complexities here :rocket:

PR #1049, opened 2022-12-15
Add Django database migrations to Dockerfile entrypoint

### LINKED ISSUE #1048: Add Django database migration to Dockerfile entrypoint
## Description

As part of the ECS-ification of the Django service, we are moving towards an automated database migration handling approach. Under this new approach, migrations will be automatically applied when a new version of the Django application is deployed. The easiest way to accomplish this would be to add it as an additional step in the Dockerfile entrypoint.

We may want to explore also adding an environment variable which could be used as a flag to determine if the migrations should be run. This would default to true for production, but could be turned off in order to speed up local development (and prevent issuing a migration before it's fully baked during local testing).

## Additional context

See the motivation for this change and the rest of the discussion here: https://github.com/WordPress/openverse-api/issues/1030

AetherUnbound:
## Fixes

Fixes #1048 by @AetherUnbound

## Description

This PR adds the Django database migration to the Dockerfile entrypoint so it occurs automatically on startup. This can be disabled by setting `DJANGO_MIGRATE_DB_ON_STARTUP=False`.

> **Note**
> We'll want to add the disable flag to the current EC2-based deployment setup, since all of the instances in the auto-scaling group start at the same time which we want to avoid (see https://github.com/WordPress/openverse-infrastructure/issues/176).

## Testing Instructions

1. Run `just build && just up && just logs web` and check that the "Running database migrations..." line does not show up in the logs.
1. Comment out the `DJANGO_MIGRATE_DB_ON_STARTUP` line in `api/env.docker` then run `just up && just logs web` and check that the "Running database migrations..." line _does_ show up in the logs.
1. (Optional) Run `just dj migrate api 0050` to roll back a migration. Then run `just down && just up && just logs web` and check that the migration is run automatically.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (approved):
LGTM, though, what is the reason to _disable_ this currently? I believe it would make applying the migrations in the current deployment easier as well (no need to manually connect the app locally, etc). In what cases do we absolutely need to disable the migrations running on startup?

For what it's worth, they run _very_ fast when it's a no-op, so it wouldn't affect local development in any noticeable way.

AetherUnbound:
My only concern about enabling this now is that we could see some undesired behavior if all 5 instances within the autoscaling group run the migration at the same time (this was a piece I [mentioned in the original discussion too](https://github.com/WordPress/openverse-infrastructure/issues/176#issuecomment-1336267874)). I'm not aware of a way to have the ASG start up one instance at a time and verify that it's healthy, but I haven't looked in depth at how that could work. So in order to prevent concurrent migrations, I thought it best to disable this for the time being.

sarayourfriend:
Oh, I see. I misremembered what you'd described the ASG deployment process was and thought that the ASG did deploy one at a time, but of course it does not :+1: This makes sense to me then.

stacimc (approved):
Tests well for me 👍

PR #1056, opened 2022-12-20
Switch to Photon for thumbnail generation

sarayourfriend:
## Fixes

Fixes #979 by @AetherUnbound

## Description

Now that the WordPress.com Photon instance supports an authentication header to allow us to forward query params, we're able to use it for thumbnail generation.

Note: This PR is still missing:

1. Add/move thumbnail unit tests to match new location `catalog.api.utils.photon` module

## Testing Instructions

Checkout the branch and `just recreate`. Test thumbnails and confirm they work.

If you have access to the Openverse secrets, you can add the `PHOTON_AUTH_KEY` variable to your `api/.env` file to enable query forwarding.

I am using the following additional sample data to be able to test the query forwarding:

```diff
diff --git a/sample_data/sample_images.csv b/sample_data/sample_images.csv
index 79a691a9..ecbb61c6 100644
--- a/sample_data/sample_images.csv
+++ b/sample_data/sample_images.csv
@@ -4999,3 +4999,4 @@ d4307b61-bd39-44bb-a3cd-aa760029d051,2021-12-15 20:48:57.914727+00,2021-12-15 20
[...]

PR #1060, opened 2022-12-27
Use thumbnail as image url image_view if available

### LINKED ISSUE #896: Use `thumbnail_url` for thumbnail generation when present
## Problem

Some providers (like SMK) may link to quite large images in their `image_url`. This can cause unacceptably slow response times or even timeouts when generating thumbnails via our thumbnail service.

## Description

We already have a `thumbnail_url` available on the Image model which we can use.

When the `thumbnail_url` is available, we should send this to the thumbnail service instead of the `image_url`. This should be a small change [here](https://href.li/?https://github.com/WordPress/openverse-api/blob/f0c44d571249a2631eef5a13ea0d56702210daf3/api/catalog/api/views/image_views.py#L109):

```
image_url = image.thumbnail or image.url
```
_(using the serialized field names)_

## Additional context

https://github.com/WordPress/openverse-catalog/issues/698 tracks updating the SMK provider script in the Catalog to populate `thumbnail_url` with a link to a smaller image size.

## Implementation

sahil-R:
## Fixes

Fixes #896 by @stacimc

## Description

Some providers can link large images in their image_url, This can cause unacceptably slow response times or even timeouts when generating thumbnails via our thumbnail service.

This pr tries to tackle this by checking if there is a thumbnail URL available or not, if yes, then it uses the thumbnail URL; if not, only then it uses the image URL.

## Checklist

      `Update index.md`).
      a parent feature branch.
- [ ] I tried running the project locally and verified that there are no visible
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/api/catalog/api/utils/photon.py
+++ b/api/catalog/api/utils/photon.py
@@ -52,6 +52,12 @@ def get(
         # request URL.
         params["q"] = parsed_image_url.query
 
+    if parsed_image_url.scheme == "https":
+        # Photon defaults to HTTP without this parameter
+        # which will cause some providers to fail (if they
+        # do not serve over HTTP and do not have a redirect)
+        params["ssl"] = "true"
+
     # Photon excludes the protocol so we need to reconstruct the url + port + path
     # to send as the "path" of the Photon request
     domain = parsed_image_url.netloc
--- a/api/test/unit/utils/photon_test.py
+++ b/api/test/unit/utils/photon_test.py
@@ -11,7 +11,7 @@ from catalog.api.utils.photon import get as photon_get
 
 
 PHOTON_URL_FOR_TEST_IMAGE = f"{settings.PHOTON_ENDPOINT}subdomain.example.com/path_part1/part2/image_dot_jpg.jpg"
-TEST_IMAGE_URL = PHOTON_URL_FOR_TEST_IMAGE.replace(settings.PHOTON_ENDPOINT, "https://")
+TEST_IMAGE_URL = PHOTON_URL_FOR_TEST_IMAGE.replace(settings.PHOTON_ENDPOINT, "http://")
 
 UA_HEADER = HEADERS["User-Agent"]
 
@@ -261,3 +261,29 @@ def test_get_generic_exception(capture_exception, setup_requests_get_exception):
         photon_get(TEST_IMAGE_URL)
 
     capture_exception.assert_called_once_with(exc)
+
+
+@pook.on
+def test_get_successful_https_image_url_sends_ssl_parameter(mock_image_data):
+    https_url = TEST_IMAGE_URL.replace("http://", "https://")
+    mock_get: pook.Mock = (
+        pook.get(PHOTON_URL_FOR_TEST_IMAGE)
+        .params(
+            {
+                "w": settings.THUMBNAIL_WIDTH_PX,
+                "quality": settings.THUMBNAIL_QUALITY,
+                "ssl": "true",
+            }
+        )
+        .header("User-Agent", UA_HEADER)
+        .header("Accept", "image/*")
+        .reply(200)
+        .body(MOCK_BODY)
+        .mock
+    )
+
+    res = photon_get(https_url)
+
+    assert res.content == MOCK_BODY.encode()
+    assert res.status_code == 200
+    assert mock_get.matched


api/catalog/api/views/image_views.py
```diff
@@ -132,7 +132,18 @@ def watermark(self, request, *_, **__):
         params.is_valid(raise_exception=True)
 
         image = self.get_object()
-        image_url = image.url
+
+        # initialising the image url to None by default
+        image_url = None
```
dhruvkb, line 137: Since `image_url` is assigned in both the `if` and `else` branches below, there is no need to initialise it with `None`.

api/catalog/api/views/image_views.py
```diff
@@ -132,7 +132,18 @@ def watermark(self, request, *_, **__):
...
+
+        # checking if thumbnail url is available or not
+        # if yes , setting image_url to thumbnailurl
+        # if no , set image_url to its default url
```
dhruvkb, line 141: This comment isn't very helpful as it only repeats what the code below is saying. See rule#1 of the [best practices for writing code comments](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/).

A better comment would be to explain what `image.thumbnail` is and why it might be messing it some cases.

api/catalog/api/views/image_views.py
```diff
@@ -132,7 +132,18 @@ def watermark(self, request, *_, **__):
...
+        if image.thumbnail is not None:
+            image_url = image.thumbnail
+        else:
+            image_url = image.url
```
dhruvkb, line 145: This can be simplified into 1 line. We could even drop the `is not None` part because an empty string in the thumbnail field is no good for us either.

```suggestion
        image_url = image.thumbnail if image.thumbnail else image.url
```

dhruvkb (changes requested):
Thanks @sahil-R! The logic is sound but the code quality can be improved in some ways.

api/catalog/api/views/image_views.py
sahil-R: Hey @dhruvkb, can you please review the changes suggested? Thank you very much.

api/catalog/api/views/image_views.py
```diff
@@ -132,7 +132,15 @@ def watermark(self, request, *_, **__):
...
         image = self.get_object()
-        image_url = image.url
+
+        # image.thumbnail is the thumbnail url
+        # which is present in the image model
+        # media providers may link large images in as image url
+        # causing the services to slow down or time out
+        # will use a thumbnail url if available for thumbnail generation
+
+        image_url = image.thumbnail if image.thumbnail else image.url
```
dhruvkb, line 142:
```suggestion
        # Media providers may provide large images in ``image.url``,
        # causing the services to slow down or time out. So, we will use
        # ``image.thumbnail`` (if available) for thumbnail generation.
        image_url = image.thumbnail if image.thumbnail else image.url
```

dhruvkb (approved):
The comment can use some punctuation and grammar improvements but other than that, LGTM!

api/catalog/api/views/image_views.py
```diff
@@ -132,7 +132,12 @@ def watermark(self, request, *_, **__):
         params.is_valid(raise_exception=True)
 
         image = self.get_object()
-        image_url = image.url
+
+        # Media providers may provide large images in ``image.url``,
+        # causing the services to slow down or time out. So, we will use
+        # ``image.thumbnail`` (if available) for thumbnail generation.
+        image_url = image.thumbnail if image.thumbnail else image.url
```
sarayourfriend, line 139: This change applies to the `watermark` method, but the original issue was related to thumbnail generation. Can you apply the same change to line 113 in the `thumbnail` method and remove the change from this line in the `watermark` method? For watermark generation I believe we would want the original image (and in any case, is out of scope of the original issue).

sarayourfriend (changes requested):

api/catalog/api/views/image_views.py
sahil-R: Hey @sarayourfriend, Thank you so much for the review, I thought since the line was mentioned in the issue i had to change it here, I'll make the required changes.

sarayourfriend: The issue linked to the line in the `thumbnail` method, but the line is identical between the two methods, so it is an easy mix-up to make. Thanks for correcting it so quickly.

sarayourfriend (approved):
LGTM, thank you for the contribution and quick changes @sahil-R.

sarayourfriend:
I approved, but prematurely. If you can please add unit tests for this change, then it will be able to be approved and merged without issue. The new tests would need to be added in `media_views_test.py`, however it will conflict with changes in #1056 which will make adding those new tests significantly easier as it cleans up the tedious fixtures currently present in that test module. If you'd like to wait for the Photon PR to be merged to make it simpler to add the tests, feel free. Otherwise, please note that there may be conflicts if the other PR is merged first.

krysal:
Ups, I made the test based on the old ones, before noticing the conflict. I'll get back to this later if @sahil-R doesn't get to it firsts.

PR #1083, opened 2023-01-13
Add database connectivity to healthcheck endpoint

### LINKED ISSUE #1079: Add database connectivity to healthcheck endpoint
## Problem

The healtcheck endpoint should check that the database is accessible. If the db is inaccessible, the service is definitively not healthy.

## Description

Add another check (in addition to the ES check) for the database connectivity. Calling `django.db.connection.ensure_connection()` should be sufficient. It raises an error when the database connection is unavailable.

krysal:
## Fixes

Fixes #1079 by @sarayourfriend

## Description

As the issue suggest, this PR adds the check using `django.db.connection.ensure_connection()`. It's actually hard to find official documentation about it (or is just me that didn't find it) but I think it's used in the expected way.

Not sure how to test a bad connection here though. I tried two approaches but the test do not fail in those cases. As it's basically just a call to a django function I assume it should work but it'd be nice to have this case covered just in case.

<details>
<summary>Attemp 1</summary>

```python
# api/test/unit/views/health_views_test.py
from django.test import override_settings

@override_settings(DATABASES=None)
@pytest.mark.django_db
def test_health_check_catch_bad_db_connection(api_client):
    res = api_client.get("/healthcheck/")
    assert res.status_code != 200    # fails
```

</details>

<details>
<summary>Attemp 2</summary>

```python
# api/test/unit/views/health_views_test.py

@pytest.mark.django_db
def test_health_check_catch_bad_db_connection(api_client):
    connection.creation.destroy_test_db("openledger")
    res = api_client.get("/healthcheck/")
    assert res.status_code != 200    # fails
```

</details>

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


sarayourfriend (commented):
LGMT. As for the tests, mocking (or "spying"/"wrapping") the method and confirming it is called when we expect it to be called when the view is hit would be sufficient, imo. We can be confident that Django's implemented it in a sensible way.

The only documentation I found for the method was in the code itself: https://github.com/django/django/blob/5da5f3773ec281a027f99c9b56add47d076650b5/django/db/backends/base/base.py#L285

api/catalog/api/views/health_views.py
```diff
@@ -22,7 +23,16 @@ class HealthCheck(APIView):
 
     swagger_schema = None
 
-    def _check_es(self) -> Response | None:
+    @staticmethod
+    def _check_db() -> None:
+        """
+        Check that the database is available.
+        Returns nothing is everything is OK, throws error otherwise.
```
sarayourfriend, line 30:
```suggestion
        Returns nothing if everything is OK, throws error otherwise.
```

api/catalog/api/views/health_views.py
```diff
@@ -37,5 +47,6 @@ def _check_es(self) -> Response | None:
     def get(self, request: Request):
         if "check_es" in request.query_params:
             self._check_es()
+        self._check_db()
```
dhruvkb, line 50: We should create an exception class (like `ElasticsearchHealthcheckException`) for the database as well, so that we can return a 503 response when the DB connection fails.

Also shouldn't this also be behind a flag? I guess maybe not, because Django wouldn't be up unless the DB connection succeeds.

dhruvkb (approved):
I have some non-blocking comments, but this is good to go without them.

api/catalog/api/views/health_views.py
krysal: The native function already [handles DB](https://github.com/django/django/blob/5da5f3773ec281a027f99c9b56add47d076650b5/django/db/backends/base/base.py#L285) [specific errors](https://github.com/django/django/blob/5da5f3773ec281a027f99c9b56add47d076650b5/django/db/backends/base/base.py#L677) for us.

sarayourfriend (approved):
LGTM!

PR #1088, opened 2023-01-14
Tally provider occurrences in results

### LINKED ISSUE #1072: Set up a `tallies` cache alias for tracking longer-term values
## Problem

We are presently using the default `django.core.cache` alias for tracking view caches & authentication information _in addition to_ general tallies and aggregates we wish to track. This means that we have both cache and (ostensibly) longer-term data living in the same Redis database, and (unless explicitly adjusted) existing under the same TTLs.

## Description

We would like to set up a separate cache alias, `tallies`, which could be used to store longer-living values. The two examples that come to mind would be the dead link tallies that are accumulated in `validate_images`, and the thumbnail domain timeouts that were added in #982.

This alias could have a longer (or nonexistent) TTL, with the option to define individual timeouts when writing to the cache.

We should also check through other uses of the `cache` and swap any uses that are not view caches & authentication info to use `tallies`.

### LINKED ISSUE #1084: Count provider occurrences in executed query results
## Problem

We do not currently have any visibility into the number of times a provider shows up in search queries or how many queries they show up in period.

When considering the concerns @AetherUnbound brought up in [this Make post about iNaturalist](https://make.wordpress.org/openverse/2023/01/14/preparing-for-inaturalist/), it became clear that we need to have visibility into these metrics so that we know how dramatic changes to existing providers and the addition of new providers is actually reflected in real user searches.

## Description

In the `search_controller.search` method, before returning the search results, create a tally of the providers for each result present in the page being returned:

```py
provider_occurrences = defaultdict(int)
for result in results:
    provider_occurences[result["provider"]] += 1

for provider, occurrences in provider_occurrences:
    redis.incr(f"provider_occurrences:{week_timestamp}:{provider}", occurrences)
    redis.incr(f"provider_appeared_in_searches:{week_timestamp}:{provider}", 1)
```

This will allow us to see (a) how many searches each provider actually shows up in over the course of a week and (b) how many results from each provider were shown in search. If we can collect this data before iNaturalist data is fully available in the API, then we'll be able to directly see the effect of adding iNaturalist, both in the count of how many iNaturalist results start to show up _and_ in and potential decreases in result counts for other providers.

## Alternatives

Store the entire list of result ids returned for all queries in a time series. This would allow us to do more different kinds of analysis long term but would take way longer for us to get usable results from for the purposes of measuring the impact of iNaturalist on the actual user query results.

sarayourfriend:
## Fixes

Fixes #1084 by @sarayourfriend 
Also fixes #1072 by @AetherUnbound

## Description

Tallies provider occurrences in Redis. Disambiguates by week (via a Monday date-stamp shared for all days of a given week), index, and provider.

Only tallies the first four pages. 

## Testing Instructions

Check out the new unit tests.

For manual testing: run the application with the instructions from the README (you will need to rebuild web if you already have it for the new test dependency, btw).

Then, make a search. Open up a redis connection (`redis-cli` or use a Redis connection from `django_redis` in `just ipython`) and execute `SCAN 0 'provider*'`. You should see the new tallies appear. You can use the keys that you get back from SCAN to view the individual results.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>


api/test/unit/utils/tallies_test.py
```diff
@@ -0,0 +1,113 @@
...
+        == b"1"
+    )
+    assert (
+        redis.get(
+            f"provider_appeared_in_searches:{FAKE_MEDIA_TYPE}:{expected_timestamp}:stocksnap"
+        )
+        == b"1"
+    )
+
+
```
AetherUnbound, line 51: Did we want to add a test (or would it be easily possible to test) that no results are incremented for page queries > 4?

AetherUnbound (approved):
This looks great! I was able to test this locally and see that keys were added for different queries, but _weren't_ added when I requested `page=5`.

api/test/unit/utils/tallies_test.py
sarayourfriend: This is tested by the test that verifies that the function isn't called by `search_controller.search` if the page is 5. Is there a different approach you would like to take?

stacimc:
@sarayourfriend Can you expand on your instructions for manual testing?

stacimc (approved):
This LGTM. I was able to figure out how to test and verified that the `occurrences` and `appeared_in_searches` tallies were increased as expected, and not for pages > 4.

For what it's worth, I wasn't clear how to test in Redis. I looked for other PRs involving similar testing and first tried exec'ing into the container with `docker-compose exec cache redis-cli`, but then the given SCAN command threw syntax errors. I'm not sure if I was missing steps or if you meant a different command.

To get it working in `just ipython` I did:

```
> import django_redis
> tallies = django_redis.get_redis_connection("tallies")

# Get the keys
> tallies.keys("provider*")

# Probably an easier way to do this, but individually check the value of relevant keys after making searches at http://localhost:50280/v1/images
> tallies.get('provider_appeared_in_searches:image:2023-01-23:flickr')
(and so on)
```

sarayourfriend:
Sorry, I left out the `MATCH` keyword for the SCAN argument (I just wrote it from memory). The Redis documentation is pretty good about these things though, if you run into similar issues in the future: https://redis.io/commands/scan/

Testing using ipython is easier and as you did, can follow the code as an example.

stacimc:
No worries. I looked at that exact documentation when debugging the syntax error 😅 I needed to add `match` and remove the quotation marks around the pattern. However even once I did that, I wasn't able to see the tallies. The only key I could see at all was `filtered_providers`. I was not sure if maybe I was missing a step to connect to the right cache?

Before that FWIW I also initially tried running `redis-cli` outside of Docker with the API running and it did not work for me -- I got connection errors when trying to connect. I did not spend much time trying to debug that part in particular, though 🤷‍♀️

sarayourfriend:
> I was not sure if maybe I was missing a step to connect to the right cache?

Ah, yes, that would be the issue.

Later I will open an issue to add more documentation for how to read from redis in development :+1:

Issue #588, opened 2022-03-24
Consider hiding Flickr behind the 'mature' flag


zackkrida:
## Problem

Currently, Flickr is the only source that has a risk of containing nudity or sexually explicit content that isn't properly identified. There is also rare gore (in the context of wilderness) from Wikimedia, that could be considered here.

## Description

We could _automatically_ hide these entire sources as mature until we have some kind of assurance that images are _indeed_, not mature. We might want to add an additional content report to mark an image as _incorrectly labeled mature_, to supplement this change. 

## Consequences

- We would have many, many fewer results on all searches; Flickr currently makes up something like 2/3 of our data set.

sarayourfriend:
Is there an alternative approach like more aggressively sorting flickr results by some account metric? Like popularity + number of uploads + activity, etc? It'd be a shame to remove the largest repository of CC images from our search by default :disappointed:

sarayourfriend:
I think treating flickr as a single "block" of results is also potentially restrictive. There are GLAM quality institutions that post on Flickr for example. If we could treat each Flickr account individually as a "source" (with a provider as Flickr) would that allow us more flexibility in being able to still promote very high quality Flickr results without punishing them for being on the same platform as lower quality or difficult-to-recommend content?

zackkrida:
> If we could treat each Flickr account individually as a "source" (with a provider as Flickr) would that allow us more flexibility in being able to still promote very high quality Flickr results without punishing them for being on the same platform as lower quality or difficult-to-recommend content?

We _do_ some of this already! And more would be an excellent way to surface the best content Flickr has to offer.

Here's where we define Flickr 'sub providers':

https://github.com/WordPress/openverse-catalog/blob/3034e31f6f204c641a89f7acbcbb85f026b9c9c3/openverse_catalog/dags/common/loader/provider_details.py#L43-L56

zackkrida:
Side note: It'd be awesome to have `FLICKR_SUB_PROVIDERS` in Airflow or the Django Admin so we could update this list dynamically and extract new subproviders in new DAG runs without deployment.

AetherUnbound:
> Side note: It'd be awesome to have `FLICKR_SUB_PROVIDERS` in Airflow or the Django Admin so we could update this list dynamically and extract new subproviders in new DAG runs without deployment.

[...]

PR #1108, opened 2023-01-25
Create pre-filtered secondary indexes and add ability to automatically filter sensitive terms at query time

sarayourfriend:
## Fixes

Related to https://github.com/WordPress/openverse-api/issues/588 by @zackkrida
Fixes https://github.com/WordPress/openverse-api/issues/1109 by @obulat (at least potentially, based on loose decisions we made last week during offline chats)

## Description

Adds a new settings variable, `SENSITIVE_TERMS`. It should be a comma separated list of terms to exclude, parsed on application startup. This variable exists for both the ingestion server and the Django API.

In the ingestion server, the terms are used to create a filtered index via the reindex API. I've updated `load_sample_data.sh` to created this filtered index and allow for easy local testing. Note that the filtered index uses the terms "dog" and "water".

In the Django API, the terms are excluded from search via an inverted `MultiMatch` query. The approach is naive and may perform poorly in production. It may be necessary to connect a local box to the production Elasticsearch to try a query with "dog" excluded, for example, to see how it performs (or some other way to test, maybe using staging Elasticsearch cluster, cc @AetherUnbound @obulat @krysal who may all have better ideas for how to safely test). I also don't know if this will scale if we have, say, 100 or so terms. I'd wager it should be fine, but it's a completely naive guess founded on essentially nothing other than trusting Elasticsearch's ability to aggregate such queries.

I don't know if this is the right approach. The significant alternative I can imagine is to store the list of terms in Postgres and caching their retrieval by 30 minutes to an hour (potentially busting the cache immediately when the model saves). This would allow us to change the terms without needing to redeploy. It makes sharing the list slightly harder because we'd have to give Django Admin access or export it from Django Admin. Leaving it as an environment variable allows the sharer to copy/paste the list out of the private infrastructure repository instead. 
A production redeployment currently takes ~10 minutes and are tedious now but will be easier (and faster when staying on the same version) in the future once the ECS migration is completed.

## Testing Instructions

By default, local environment is set up with two excluded terms for the filtered index: "dog" and "water". For the Django API, the excluded terms are "spoiled" and "perched". I'd advise searching these terms on `main` first, before running this branch, and noting the result counts and such for related queries. I tested using images so the following instructions only include specifics for images, but the same principles will apply for audio.

Make a query for "dog" and "water" and pick out a separate word that would hit that document. For example, "running" will include 2 "water" documents (amongst others of people running).

Afterwards, run this branch and make the same queries with `mature=True` and `mature=False` (the latter being the default). You should see different behaviour. When mature results are excluded, for "running", you should get 2 less documents, equivalent to searching "running -water". For "dogs" you should get no documents. When mature results are included via the query parameter, you'll receive the same results as on `main`.

Again, those are for the images index, but the same principles will apply for audio.

To test the API terms, follow the same pattern of testing, but use the terms listed above for the query-time feature instead of the filtered index feature.

## Checklist

      `Update index.md`).
      a parent feature branch.
      errors.

[best_practices]:

<details>

```
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/ingestion_server/ingestion_server/indexer.py
+++ b/ingestion_server/ingestion_server/indexer.py
@@ -57,6 +57,10 @@ REP_TABLES = config(
     "COPY_TABLES", default="image", cast=lambda var: [s.strip() for s in var.split(",")]
 )
 
+SENSITIVE_TERMS = config(
+    "SENSITIVE_TERMS", default="", cast=lambda x: (t.strip() for t in x.split(","))
+)
+
 
 def database_connect(autocommit=False):
     """
@@ -309,6 +313,38 @@ class TableIndexer:
         schedule_distributed_index(
             database_connect(), model_name, table_name, destination_index, self.task_id
         )
+
+    def create_filtered_index(self, model_name: str, index_suffix: str, **_):
+        """
+        Create a secondary index that excludes documents filtered by a list of terms.
+        """
+
+        source_index = f"{model_name}-{index_suffix}"
+        destination_index = f"{source_index}-filtered"
+
+        log.info(f"Creating filtered index '{destination_index}' from '{source_index}'")
+        self.es.reindex(
+            body={
+                "source": {
+                    "index": source_index,
+                    "query": {
+                        "bool": {
+                            "must_not": [
+                                {
+                                    "multi_match": {
+                                        "query": term,
+                                        "fields": ["tags.name", "title", "description"],
+                                    }
+                                }
+                                for term in SENSITIVE_TERMS
+                            ]
+                        }
+                    },
+                },
+                "dest": {"index": destination_index},
+            }
+        )
+        self.ping_callback()
 
     def update_index(self, model_name: str, index_suffix: str, since_date: str, **_):
         """
--- a/ingestion_server/ingestion_server/indexer.py
+++ b/ingestion_server/ingestion_server/indexer.py
@@ -316,7 +316,10 @@ class TableIndexer:
 
     def create_filtered_index(self, model_name: str, index_suffix: str, **_):
         """
-        Create a secondary index that excludes documents filtered by a list of terms.
+        Create a secondary index that excludes documents by a list of sensitive terms.
+
+        :param model_name: the name of the media type
+        :param index_suffix: the unique suffix to add to the index name
         """
 
         source_index = f"{model_name}-{index_suffix}"
--- a/ingestion_server/ingestion_server/indexer.py
+++ b/ingestion_server/ingestion_server/indexer.py
@@ -344,6 +344,7 @@ class TableIndexer:
                 "dest": {"index": destination_index},
             }
         )
+        self.refresh(index_name=destination_index, change_settings=True)
         self.ping_callback()
 
     def update_index(self, model_name: str, index_suffix: str, since_date: str, **_):
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -13,7 +13,7 @@ from rest_framework.request import Request
 
 from elasticsearch.exceptions import NotFoundError, RequestError
 from elasticsearch_dsl import Q, Search
-from elasticsearch_dsl.query import EMPTY_QUERY, MoreLikeThis, Query
+from elasticsearch_dsl.query import EMPTY_QUERY, MoreLikeThis, MultiMatch, Query
 from elasticsearch_dsl.response import Hit, Response
 
 import catalog.api.models as models
@@ -345,6 +345,10 @@ def search(
             "fields": search_fields,
             "default_operator": "AND",
         }
+
+        if not search_params.data["mature"]:
+            for t in settings.SENSITIVE_TERMS:
+                s = s.query(~MultiMatch(query=t, fields=search_fields))
 
         if '"' in query:
             base_query_kwargs["quote_field_suffix"] = ".exact"
--- a/api/catalog/api/controllers/search_controller.py
+++ b/api/catalog/api/controllers/search_controller.py
@@ -303,6 +303,9 @@ def search(
     :return: Tuple with a List of Hits from elasticsearch, the total count of
     pages, and number of results.
     """
+    if not search_params.data["mature"]:
+        index = f"{index}-filtered"
+
     search_client = Search(index=index)
 
     s = search_client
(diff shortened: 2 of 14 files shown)


sarayourfriend:
This PR will need to include an update to the "Search Algorithm" documentation describing this new feature.

zackkrida:
This is awesome! I had a random thought about how to test it (and potentially other changes like this) against production data:

What if we only applied this to searches with a "secret password" in the query? So, it'd work something like this in production:

1. A user searches "dog". No sensitive keyword filtering is applied.
2. An Openverse contributor searches "dog SECRET_PASSWORD". Sensitive keyword filtering is applied and the `SECRET_PASSWORD` is stripped from the query before searching ES

This would allow us to test in production for some arbitrary period so we could make performance comparisons.

sarayourfriend:
@dhruvkb I was wondering if you could help me take a look at the ingestion server tests? I made some changes to the ordering so that it was easier to put new tests in the middle without having to update every other test's order number. That appears to be working fine, but no matter where I put the tests for creating the filtered index and pointing the alias in the order, it seems to cause subsequent tests that use the promote action to fail. I'm not entirely sure why. I'd appreciate any insight you might have into this issue.

dhruvkb:
The tests are all passing locally for me, so that's even more confusing! I'll keep looking.

ingestion_server/ingestion_server/indexer.py
```diff
@@ -312,6 +316,38 @@ def reindex(
...
+                                }
+                                for term in SENSITIVE_TERMS
+                            ]
+                        }
+                    },
+                },
+                "dest": {"index": destination_index},
+            }
+        )
+        self.ping_callback()
```
dhruvkb, line 349: This will send the callback immediately, whereas ES will continue to process the `reindex` call on its end and create the index. This is most likely what is breaking the tests because the index is not ready by the time the test starts asking for it.

stacimc:
This is so cool! I'm really excited to test this with more data 😮 The 'password' idea from @zackkrida sounds really interesting.

Is this PR meant as an exploration or do you intend to actively keep pushing this one? I have some questions but obviously not urgent if this is on the backburner.

I understand the filtered index in the ingestion server. If I search "dog photo" with the mature filter enabled, I will end up with zero results because everything matching my query is also in the index. I will not receive partial match results (meaning, things that match "photo" but not "dog"). The filtering also happens on every query including ones that don't intentionally query on a sensitive term: so if I search just "photo", this time I'll get lots of results, but I still won't see any dog photos (or water photos for that matter).

I am confused about the API filtering, though. It looks like it detects sensitive terms in the query params, and then excludes results that match only those terms? So, if one of the configured terms is "perched":
* If I search `?q="bird"`, I will see results that include "bird perched", because "perched" wasn't in the query params
* If I search `?q="bird perched", I will see results that match "bird -perched"? Is that the intention?

Is that interpretation of what it's supposed to do correct? Records matching sensitive terms defined in the API's list are only filtered when those terms actually appear in query params?

I think I'm wrong about that, but I'm not sure what I'm missing. When I query `?q="bird perched"` locally with the mature filter enabled, I get 25 results including many with "perched" in the title or tags 🤔  When I add `mature=False` I get one additional result (I did not have time to look into what was different with the one excluded record).

sarayourfriend:
> Is this PR meant as an exploration or do you intend to actively keep pushing this one? I have some questions but obviously not urgent if this is on the backburner.

I thought it could be merged, but I still have not received clarification from others on the team of whether they want that to happen (I asked during our retrospective). I haven't heard anyone say to "stop" working on this or that it shouldn't move forward, so assuming there aren't big problems with it, we could try it. Then again, if it's not something we think we would enable any time soon, then I should close the PR as an unmerged proof of concept to be referred back to later. Either way does not matter to me.

> I am confused about the API filtering, though. It looks like it detects sensitive terms in the query params, and then excludes results that match only those terms? So, if one of the configured terms is "perched":

Your summary is incorrect but the behaviour you're seeing is reproducible. Just to clarify the behaviour first though: the code applies the sensitive word filter _always_ to all queries, for all sensitive words, regardless of whether they appear in the query. In fact, it applies the filter in precisely the same way as the filtered index is produced, so the behaviour is (essentially) the same.

This is the part of the code that applies the filter: https://github.com/WordPress/openverse-api/pull/1108/files#diff-1f1af6f89cdc3071047abe1d692e5803c38df879f2da130bf178dcb444cb8e28R345-R347

It doesn't check any query params aside from whether the mature filter is disabled. None of it's operation or implementation depends on any other query parameters or their values.

There was, however, a bug in the environment variable reading implementation. It wasn't applying _any_ sensitive term filters at runtime because the cast was creating a generator, not a tuple. I've fixed this now and the behaviour you were seeing is no longer reproducible. If you search "bird" you won't see and results for "perched". If you search "bird perched", you will also not see results for "perched". (Unless you disable the mature filter, to be clear). When you search, if you look at the logs, you'll be able to see the multimatch queries being sent at all times, regardless of what the other terms of the query are.

sarayourfriend:
Closing this PR again as I don't have any idea whether anyone else wants this to move forward, and I do not feel confident about getting it reviewed and merged before the 17th when @dhruvkb will be doing the monorepo migration.

stacimc:
> Closing this PR again as I don't have any idea whether anyone else wants this to move forward and I do not feel confident about getting it reviewed and merged before the 17th

Noted. Leaving the comment I was working on for posterity when this is revisited. For the record I think this is really exciting.

> Your summary is incorrect but the behaviour you're seeing is reproducible. Just to clarify the behaviour first though: the code applies the sensitive word filter always to all queries, for all sensitive words, regardless of whether they appear in the query. In fact, it applies the filter in precisely the same way as the filtered index is produced, so the behaviour is (essentially) the same.

That makes way more sense 😅 I was trying to make sense of my test behavior and assumed the filters must do different things and work together somehow. The context I was missing/had forgotten was in the [comments](https://github.com/WordPress/openverse-api/issues/1109#issuecomment-1408104242) of one of the linked issues: they do the same thing, but the reason for also having filtering in the API is to allow for adding additional sensitive terms in a hypothetical emergency without a redeploy.

Thank you for the explanation! I'm inclined to agree with your [comment](https://github.com/WordPress/openverse-api/issues/1109#issuecomment-1408104242) about the API filtering possibly not being needed, especially to your point on this issue about the deploys being fairly quick and getting even better 😄 That said, you've already done the work and it works great!

